PR #14, opened 2016-12-06
use service title in image name

hugoesthere:
Use `author/service-title` naming convention for docker images

@kevgo @trushton @martinjaime


exo-deploy/src/docker/docker-hub.ls
```diff
@@ -40,7 +40,8 @@ class DockerHub
         service-config = require path.join(process.cwd!, config.location, 'service.yml')
         names.push do
           author: service-config.author
-          name: path.basename config.location
+          name: service-config.title |> (.replace /\s/g, '-')
```
kevgo: Use https://www.npmjs.com/package/kebab-case to translate the string properly into kebab-case.

exo-deploy/src/docker/docker-hub.ls
```diff
@@ -40,7 +40,8 @@ class DockerHub
...
+          #TODO: get image name if location is docker on dockerhub
```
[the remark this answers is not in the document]
kevgo: Tododrop is right, this should be a ticket on Github!

exo-deploy/src/terraform/aws-terraform-file-builder.ls
```diff
@@ -71,7 +71,8 @@ class AwsTerraformFileBuilder
   _generate-services: (type) ->
     for service-name, service-data of @app-config.services["#{type}"]
       service-config = require path.join('/var/app', service-data.location, 'service.yml')
-      @_build-service-container-definition service-name, service-config
+      image-name = @_get-image-name service-data
```
kevgo: I think you can inline this variable

kevgo (approved):

exo-deploy/src/docker/docker-hub.ls
hugoesthere: This is converting whitespace separated titles to kebab-case. That package will convert camel-case to kebab-case only. Should we standardize what format service titles can be in?

kevgo: Okay, then use https://www.npmjs.com/package/dashify. I'd rather leave the format of service titles open.

PR #18, opened 2016-12-07
Ignore documentation sub-project in morula.yml

hugoesthere:
references https://github.com/Originate/exosphere-sdk/issues/202

@kevgo @trushton @martinjaime


trushton:
@hugobho As I'm sure you saw, Travis has the tests failing because the subproject `.github` also has no `bin/setup` script. I know morula only takes one value for the `never` field so maybe we want to just create an empty `bin/setup` for that folder for now?

morula.yml
```diff
@@ -1,5 +1,5 @@
 before-all: ""
-after-all: "documentation"
+after-all: ""
 always: ""
-never: ""
+never: ".github"
```
kevgo: Directories starting with a dot are (or at least should be) always ignored: https://github.com/Originate/morula/blob/master/features/ignore.feature#L11-L20

Are you using Morula 0.1?

kevgo (approved):
Thanks for doing this! Please update https://github.com/Originate/exosphere/blob/master/.travis.yml#L21 to download Morula 0.1, then you can remove `.github` from the `never` block. The placeholder scripts for `documentation` can remain, we'll need them in a few days anyways.

Issue #19, opened 2016-12-07
Install/Test ExoSphere via Webpack


trushton:
As an Exosphere developer
I want to run/test all different code bases that make up the SDK together
So that I can develop and release changes in multiple code bases at the same time.

Rules

- running the tests uses the local code bases
- publishing a new version uses the local code bases

Notes

- here is the build script from YARN: https://github.com/yarnpkg/yarn/blob/master/scripts/build-webpack.js

PR #20, opened 2016-12-07
Update exo-clone for Webpack

trushton:
partially resolves #19

Updates the `exo-clone` command to use the local exosphere-shared directory, as well as updates the command for the changes that have been made to `application.yml` recently involving the public/private fields. Finally, wraps the `cli.ls` code in a function so that it works nicely with Webpack. 

@kevgo @hugobho @martinjaime


exo-clone/package.json
```diff
@@ -3,30 +3,29 @@
   "version": "0.0.6",
   "author": "Kevin Goslar",
   "dependencies": {
-    "async": "2.0.1",
+    "async": "^2.1.4",
```
kevgo: Remove the `^` everywhere here, and run `npm config set save-exact true` in the terminal to make NPM do the right thing here once and for all.

exo-clone/src/app-cloner.ls
```diff
@@ -23,14 +23,16 @@ class AppCloner extends EventEmitter
...
-      cloners = for service-name in Object.keys @app-config.services
-        service-dir = path.join @repository.path, @app-config.services[service-name].local
-        service-origin = @app-config.services[service-name].origin
-        new ServiceCloner service-name, root: @repository.path, path: service-dir, origin: service-origin
-          ..on 'service-clone-success', (name) ~> @emit 'service-clone-success', name
-          ..on 'service-clone-fail', (name) ~> @emit 'service-clone-fail', name
-          ..on 'service-invalid', (name) ~> @emit 'service-invalid', name
-          ..on 'output', (data) ~> @emit 'output', data
+      cloners = []
+      for type of @app-config.services
```
kevgo: Why do you break up the loop comprehension here?

exo-clone/src/service-cloner.ls
```diff
@@ -23,11 +23,11 @@ class ServiceCloner extends EventEmitter
 
 
   _is-valid-service: ->
-      try
-        fs.access-sync path.join(@config.path, 'service.yml')
-        true
-      catch
-        false
+    try
```
kevgo, line 26: This could have been its own PR

exo-clone/features/clone.feature
```diff
@@ -20,12 +20,14 @@ Feature: cloning an Exosphere application and its services
       version: '1.0'
 
       services:
-        web:
-          local: ./web
-          origin: ../origins/web
-        users:
-          local: ./users
-          origin: ../origins/users
+        public:
```
kevgo, line 23: This change should have been its own PR

kevgo (approved):
This PR contains a number of changes that should have been their own PRs. Besides that well done!

PR #21, opened 2016-12-07
Update exo-lint for Webpack

trushton:
**partially** resolves #19 

Updates the `exo-lint` command to use the local exosphere-shared directory and wraps the `cli.ls` code in a function so that it works nicely with Webpack. 

@kevgo @hugobho @martinjaime


exo-lint/src/cli.ls
```diff
@@ -2,15 +2,21 @@ require! {
...
 
-app-config = yaml.safe-load fs.read-file-sync('application.yml', 'utf8')
-console.log "Running linter for #{cyan app-config.name}\n"
-logger = new Logger Object.keys(app-config.services)
-app-linter = new AppLinter app-config
-  ..on 'lint-success', -> logger.log name: 'exo-lint', text: 'Lint passed'
-  ..on 'output', (data) -> logger.log data
-  ..on 'reset-colors', (service-names) -> logger.set-colors service-names
-  ..start!
+lint = ->
```
kevgo: You can also just say `module.exports = ->` here.

exo-lint/bin/exo-lint
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env node
 
-require('../dist/cli');
+require('../dist/cli')('lint');
```
kevgo: Why do you call this with the parameter `lint` here? It isn't used.

kevgo (approved):

PR #22, opened 2016-12-07
Update exo-run for Webpack

trushton:
**partially** resolves #19

Updates the `exo-run` command to use the local exosphere-shared directory and wraps the `cli.ls` code in a function so that it works nicely with Webpack. 

@kevgo @hugobho @martinjaime


exo-run/bin/exo-run
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env node
 
-require('../dist/cli');
+require('../dist/cli')('run');
```
kevgo: Remove the unused parameter.

kevgo (approved):

PR #24, opened 2016-12-07
Update exo-sync for Webpack

trushton:
partially resolves #19

Updates the `exo-sync` command to use the local exosphere-shared directory and wraps the `cli.ls` code in a function so that it works nicely with Webpack. 

@kevgo @hugobho @martinjaime


exo-sync/src/app-syncer.ls
```diff
@@ -14,9 +14,13 @@ class AppSyncer extends EventEmitter
 
 
   start-syncing: ->
-    syncers = for service-name, service of @app-config.services ? {}
-      new ServiceSyncer service-name, root: service.location
-        ..on 'output', ~> @emit 'output', &0
+    syncers = []
+    for type of @app-config.services
```
kevgo: Why do you break up the loop comprehension?

kevgo (approved):

PR #27, opened 2016-12-07
Update exo-create for Webpack

trushton:
**partially** resolves #19

Updates the `exo-create` command to use the local exosphere-shared directory, as well as updates the command for the changes that have been made to `application.yml` recently involving the public/private fields. Finally, wraps the `cli.ls` code in a function so that it works nicely with Webpack. 

@kevgo @hugobho @martinjaime


exo-create/src/cli.ls
```diff
@@ -1,19 +1,25 @@
...
 
-entity-name = process.argv[2]
-return missing-entity! unless entity-name
-command-handler-path = "#{__dirname}/entities/#{abbrev(entity-names!)[entity-name]}.js"
-fs.access command-handler-path, (err) ->
-  | err  =>  return unknown-command entity-name
-  require command-handler-path
+create = ->
+
+
```
kevgo, line 17: Remove these two empty lines

kevgo (approved):

PR #31, opened 2016-12-08
update morula

hugoesthere:
Update morula take 2


martinjaime:
Why would the error be that the documentation directory has no `bin/setup`?

hugoesthere:
@martinjaime that's what I'm trying to figure out. I've put up two different PRs both with `bin/setup` that I can execute locally. Not sure why Travis can't find that file..

martinjaime:
@hugobho, to repeat what I messaged on slack, it also seems to be running commands in hidden directories even though morula is supposed to ignore them.

hugoesthere:
Another branch making the same changes has been merged into master.

PR #34, opened 2016-12-09
Bundle exosphere as a single package using Webpack

trushton:
resolves #19 

This PR adds the `webpack-config.js` to bundle all of the repos' content into `bundle.js`. It requires all of the exo commands from their local directories as opposed to the previous method of having them as their own npm dependencies. Finally, this PR updates the feature tests to pass with the aforementioned updates. 

@kevgo @hugobho @martinjaime @alexdavid


cli/webpack.config.js
```diff
@@ -0,0 +1,42 @@
+var path = require('path');
+var fs = require('fs');
+
+var nodeModules = {};
+fs.readdirSync('node_modules')
+  .filter(function(x) {
+    return ['.bin'].indexOf(x) === -1;
```
kevgo: Move this to the end of the previous line

cli/webpack.config.js
```diff
@@ -0,0 +1,42 @@
...
+  })
+  .forEach(function(mod) {
+    nodeModules[mod] = 'commonjs ' + mod;
```
kevgo: move to the end of the previous line

cli/webpack.config.js
kevgo: Also, can you simplify this to

```js
return x !== '.bin'

**Review by kevgo — APPROVED** [review]
This looks great!

File: cli/webpack.config.js [hunk 3]
```diff
@@ -0,0 +1,38 @@
+var path = require('path');
+var fs = require('fs');
+
+var nodeModules = {};
+fs.readdirSync('node_modules')
+  .filter(function(x) { return x !== '.bin'; })
+  .forEach(function(mod) { nodeModules[mod] = 'commonjs ' + mod; });
```
alexdavid, line 7: Can we read these keys from `package.json` instead?

cli/webpack.config.js
```diff
@@ -0,0 +1,38 @@
+var path = require('path');
```
alexdavid, line 1: webpack supports non-js config files if you have the proper compiler installed in node_modules. This could be a `.ls` file

kevgo: @trushton please address this comment

cli/webpack.config.js
kevgo: @trushton please address this comment

PR #36, opened 2016-12-09
Fix exo-add tests

hugoesthere:
Some feature files in `exo-add` were out of date. More importantly, exosphere-shared wasn't being built in its setup script, hence the Travis error complaining that it couldn't find the exosphere-shared module.

@kevgo


kevgo (approved):
Excellent! Thanks for fixing that, Hugo!

PR #37, opened 2016-12-09
Fix failing exo-create test

hugoesthere:
@kevgo @trushton @martinjaime


trushton (approved):

trushton:
Good catch, this was passing before the changes to the app-templates were made, but they really shouldn't have been.

kevgo (approved):

PR #38, opened 2016-12-09
Fix exo-lint tests

hugoesthere:
@kevgo @trushton @martinjaime


exo-lint/src/app-linter.ls
```diff
@@ -38,18 +38,23 @@ class AppLinter extends EventEmitter
...
+          (received-messages[message] or= []).push service-name
 
     {sent-messages, received-messages}
 
 
-  get-config: (service-name) ->
-    service-root = path.join process.cwd!, @app-config.services[service-name].location
+  get-config: (service-data) ->
+    service-root = path.join process.cwd!, service-data.location
     yaml.safe-load fs.read-file-sync(path.join(service-root, 'service.yml'), 'utf8')
```
trushton: This could be simplified to `yaml.safe-load fs.read-file-sync(path.join process.cwd!, service-data.location. 'service.yml'), 'utf8')`. No need for multiple `path.join` calls when the `service-root` variable is only used here.

exo-lint/src/app-linter.ls
```diff
@@ -38,18 +38,23 @@ class AppLinter extends EventEmitter
   aggregate-messages: ->
     sent-messages = {}
     received-messages = {}
-    for service-name of @app-config.services
-      service-config = @get-config service-name
-      for message in service-config.messages.sends or []
-        (sent-messages[message] or= []).push service-name
-      for message in service-config.messages.receives or []
-        (received-messages[message] or= []).push service-name
+    console.log @app-config
```
trushton: Remove `console.log`s

exo-lint/src/app-linter.ls
```diff
@@ -38,18 +38,23 @@ class AppLinter extends EventEmitter
...
+    for service-type of @app-config.services
+      console.log service-type
```
trushton: here

exo-lint/src/app-linter.ls
```diff
@@ -38,18 +38,23 @@ class AppLinter extends EventEmitter
...
+      for service-name, service-data of @app-config.services[service-type]
+        console.log service-name
+        console.log service-data
```
trushton: here

trushton (approved):

exo-lint/src/app-linter.ls
hugoesthere: Thanks for catching these

kevgo (approved):

PR #39, opened 2016-12-09
Fix exo-setup tests

hugoesthere:
@kevgo @trushton @martinjaime


trushton (approved):
Another good catch, don't want to be using those old node_modules for exosphere-shared

kevgo (approved):
Nice cleanup!

PR #40, opened 2016-12-09
Remove exocom container if exo-run was executed

trushton:
Previously, when running `bin/spec` in the `cli` subrepo, there would be an error in the tutorial feature because port 8000 was being used by an exocom docker container that was started in the `abbreviations.feature`. Since the process started by `abbreviations.feature` was killed using `kill-child-processes` of exosphere-shared, it does not give time for exo-run to shut down correctly and remove its docker containers. This change accounts for that. 
 

@kevgo @hugobho @martinjaime

### DIFF
--- a/cli/features/steps/steps.ls
+++ b/cli/features/steps/steps.ls
@@ -140,6 +140,7 @@ module.exports = ->
       | 'exo create service'     => 'We are about to create a new Exosphere service'
       | 'exo add'                => 'We are about to add a new Exosphere service to the application'
     @process.wait expected-text, ~>
+      # Need to remove exocom dontainer to clear ports for future tests
       DockerHelper.remove-container \exocom if expected-text is \exo-run
       done!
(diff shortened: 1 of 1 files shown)


cli/features/steps/steps.ls
```diff
@@ -139,7 +139,9 @@ module.exports = ->
       | 'exo create application' => 'We are about to create a new Exosphere application'
       | 'exo create service'     => 'We are about to create a new Exosphere service'
       | 'exo add'                => 'We are about to add a new Exosphere service to the application'
-    @process.wait expected-text, done
+    @process.wait expected-text, ~>
+      DockerHelper.remove-container \exocom if expected-text is \exo-run
```
kevgo, line 144: Please add a comment describing why you need to wait here when `exo-run` is running

kevgo (approved):

PR #41, opened 2016-12-09
Always run exosphere-shared on Travis

hugoesthere:
When changes to sub-projects (such as `exo-run`, `exo-setup`) are made, Morula will run tests for only that sub-project. This means `exosphere-shared` isn't built and cannot be used. This PR always runs exosphere-shared on Travis.

@kevgo @trushton @martinjaime

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,8 @@ before_install:
   - cd $TRAVIS_REPO_SLUG
   - git checkout $TRAVIS_BRANCH
   - curl --output morula --location --fail https://github.com/Originate/morula/releases/download/0.1/morula-linux-amd64 && chmod +x morula
-
+  - pwd
+  - ls
 
 install:
   - './exosphere-shared/bin/setup'
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ before_install:
 
 
 install:
+  - './exosphere-shared/bin/setup'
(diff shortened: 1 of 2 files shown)


kevgo:
I don't understand this change. If `exosphere-shared` isn't changed on a feature branch, it shouldn't be tested. If it is changed, it will be tested by Morula.

There is no point in always testing `exosphere-shared` in my opinion.

hugoesthere:
When we make changes to let's say `exo-run`, Morula will only run tests for that sub-project in the `travis-ci/push` test. However, Travis will encounter an error because `module exosphere-shared cannot be found`. This is because the `setup` script of exosphere-shared was never run so `exo-run` cannot find the compiled source code of exosphere-shared (`dist/index.js`). 

We can add `exosphere-shared` to the always field in morula.yml, or we can have each of the exo commands always compile `exosphere-shared` when it runs a test. I think since this is only an error we're encountering on Travis, it makes sense to make the change in morula.yml

kevgo:
In [.travis.yml](https://github.com/Originate/exosphere/blob/master/.travis.yml#L26) it runs `morula changed bin/setup`. If the `exosphere-shared` repo contains changes, that means it will (should) setup and compile that repo as well. Is it not doing that?

hugoesthere:
@kevgo what do you think?

kevgo (commented):
I like it!

PR #42, opened 2016-12-09
Add setup step when starting applications in exo-run tests

trushton:
Adds a step after checking out the application to run `exo-setup`. This is needed in case a user does not have any of the docker images before running the spec test for `exo-run`. We had not encountered this yet as we had tested exo-setup beforehand which took care of this.

@kevgo @hugobho @martinjaime

### DIFF
--- a/exosphere-shared/example-apps/crashing-service/runner/server
+++ b/exosphere-shared/example-apps/crashing-service/runner/server
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
 
 echo 'running ...'
+sleep 100000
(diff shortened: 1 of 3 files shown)


trushton:
@kevgo, was there a reason for the `read test` line in the crashing-service example app needed? It was causing the crashing-service example app to fail its test because the `read test` line was telling exo-run that the wrong service had crashed.

exosphere-shared/example-apps/crashing-service/runner/server
```diff
@@ -1,4 +1,3 @@
 #!/usr/bin/env bash
 
 echo 'running ...'
-read test
```
kevgo, line 4: This example app contains two services: crasher and runner. Crasher crashes, while runner keeps on running. This file is the implementation of the runner service. Without this line, the service will also stop. It should keep running

trushton:
@kevgo Should be good to go now.

kevgo (approved):

PR #43, opened 2016-12-10
Implement exosphere-shared method in Docker

hugoesthere:
We cannot require `exopshere-shared` as a module now that we are using Webpack. Instead of mounting `exosphere-shared` into Docker to use just one method, we are copying that code over instead.

@kevgo @trushton @martinjaime

### DIFF
--- a/exo-deploy/src/terraform/aws-terraform-file-builder.ls
+++ b/exo-deploy/src/terraform/aws-terraform-file-builder.ls
@@ -1,6 +1,6 @@
 require! {
   'async'
-  '../../../exosphere-shared' : {compile-service-messages}
+  '/usr/src/exosphere-shared' : {compile-service-messages}
   'fs-extra' : fs
   'handlebars'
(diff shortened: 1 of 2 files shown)


exo-deploy/src/terraform/aws-terraform-file-builder.ls
```diff
@@ -154,11 +153,26 @@ class AwsTerraformFileBuilder
...
     fs.write-file-sync target-path, JSON.stringify(container-definition, null, 2)
 
 
+  _compile-service-messages: ->
+    service-messages = []
+    for type of @app-config.services
+      for service-name, service-data of @app-config.services["#{type}"]
+        service-config = require path.join('/var/app', service-data.location, 'service.yml')
+        service-messages.push do
+          {
```
kevgo: Remove the curly braces here

kevgo (approved):

PR #46, opened 2016-12-13
Build exo-setup in exo-run tests

### LINKED ISSUE #45: exo-run tests failing
One of the `exo-run` tests timed out on Travis. The scenario is testing auto-restart of services and it times out when the process is still rebuilding the service. Either this is an anomaly or we were just lucky not to run into this error before. See: https://travis-ci.org/Originate/exosphere/jobs/183388747

EDIT: upon closer inspection it looks like the error is related to building `exo-setup`.
```
Error: Cannot find module '../dist/cli'
...
  at Object.<anonymous> (/home/travis/build/Originate/exosphere/Originate/exosphere/exo-setup/bin/exo-setup:3:1)
```
This must be from the `@setup-app` method in `world.ls` which uses `exo-setup`.

hugoesthere:
resolves #45 

Since `exo-run` uses `exo-setup` in its tests, we must run setup in that directory before running the tests.

@kevgo @trushton @martinjaime

### DIFF
--- a/exo-run/bin/features
+++ b/exo-run/bin/features
@@ -2,6 +2,8 @@
 set -e
 
 node_modules/o-tools-livescript/bin/build
+cd ../exo-setup && bin/setup && node_modules/.bin/build
+cd ../exo-run
 if [ "$#" == "0" ]; then
   cucumber-js --strict
 else
--- a/exo-run/bin/features
+++ b/exo-run/bin/features
@@ -2,8 +2,7 @@
 set -e
 
 node_modules/o-tools-livescript/bin/build
-cd ../exo-setup && bin/setup && node_modules/.bin/build
-cd ../exo-run
+(cd ../exo-setup && bin/setup && node_modules/.bin/build)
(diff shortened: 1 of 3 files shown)


exo-run/bin/features
```diff
@@ -2,6 +2,8 @@
 set -e
 
 node_modules/o-tools-livescript/bin/build
+cd ../exo-setup && bin/setup && node_modules/.bin/build
```
kevgo: Run this line inside a subshell, to avoid having to go back. Changing directory in the subshell doesn't affect the current directory in the main shell.

```bash
(cd ../exo-setup && bin/setup && node_modules/.bin/build)
if [ "$#" == "0" ]; then

hugoesthere:
@kevgo 

Ready for another look. Although I'm not sure how all 3 tests are passing, when I look at the log for the `/pr` tests I still see `Error: Cannot find module '../dist/cli'`

kevgo (approved):

PR #50, opened 2016-12-14
Convert webpack.config to livescript

trushton:
This PR addresses the comment made by @alexdavid on #34 to rewrite the webpack config in livescript. In response to the other comment of reading `node_modules` from `package.json`, I don't believe we can do that, as the modules listed in the `node_modules` directory include all of the dependencies of the dependencies listed in `package.json`, which would not be compiled for commonjs if we only read from `package.json`.

@kevgo @alexdavid


kevgo (approved):

alexdavid:
Looks like we could use this for externals: https://www.npmjs.com/package/webpack-node-externals

However, is it that bad if we compile the dependencies in the bundle as well? We could then distribute exosphere sdk with 0 dependencies and it won't break if a dependency gets unpublished or a dependency of a dependency updates

kevgo:
If we distribute Exosphere via NPM I think it's fine to use NPM to install dependencies. If we distribute it via a binary, then yes we should bundle up all external dependencies with it.

PR #51, opened 2016-12-14
Replace require-yaml with yaml.safe-load

trushton:
Webpack does not seem to like require statements as expressions (like when using require-yaml to import yaml files), this change resolves issues related to that in the exo-deploy sub-repo

@kevgo @martinjaime @hugobho


kevgo (approved):
I think webpack requires that `require` statements have hard-coded values. You can't require a module defined by a variable with webpack.

PR #53, opened 2016-12-14
Use service.yml title as docker image name

### LINKED ISSUE #13: Naming convention of Docker images (exo-setup)
Docker images created by `exo-setup` have until now lacked a proper naming convention. We have decided on using the `title` field of `service.yml` as the name of Docker images that will be created. The full image identifier will include the images author as well, so the format is `service.yml-author/service.yml-title`.

trushton:
resolves #13

Using dashify to make the title field of `service.yml` compatible as a docker image name. This name reflects what will be run in exo-run and deployed to dockerhub with exo-deploy. 

@kevgo @hugobho @martinjaime

### DIFF
--- a/exo-setup/src/docker-setup.ls
+++ b/exo-setup/src/docker-setup.ls
@@ -25,8 +25,8 @@ class DockerSetup extends EventEmitter
 
 
   _build-docker-image: (done) ~>
-    new ObservableProcess(call-args(DockerHelper.get-build-command author: @service-config.author, name: dashify @service-config.title),
-                          cwd: @config.root,
+    new ObservableProcess(call-args(DockerHelper.get-build-command author: @service-config.author, name: dashify(@service-config.title)),
+                          cwd: @config.root
                           stdout: {@write}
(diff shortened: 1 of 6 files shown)


exo-setup/src/docker-setup.ls
```diff
@@ -26,8 +25,7 @@ class DockerSetup extends EventEmitter
 
 
   _build-docker-image: (done) ~>
-    service-name = @config.root.split path.sep |> last
-    new ObservableProcess(call-args(DockerHelper.get-build-command author: @service-config.author, name: service-name),
+    new ObservableProcess(call-args(DockerHelper.get-build-command author: @service-config.author, name: dashify @service-config.title),
```
kevgo: I would put parentheses for the dashify call, so that it is more clear that we have a method call here.

```ls
new ObservableProcess call-args(DockerHelper.get-build-command author: @service-config.author, name: dashify(@service-config.title),

kevgo (approved):
Great idea!

PR #55, opened 2016-12-14
Check message name in template test def (exosphere-shared)

trushton:
resolves [Originate/exosphere-sdk#195](https://github.com/Originate/exosphere-sdk/issues/195)

The modified step was not checking that the correct message name was being sent to and received by exocom. Once this check was added, it was noticed that the `list.feature` itself had an additional 's' in the message name that it was expecting, causing the message-name check to fail.

@kevgo @hugobho @martinjaime


kevgo (approved):
Great, thanks for fixing this!

PR #56, opened 2016-12-14
Read online-text from service.yml (exosphere-shared)

trushton:
resolves [Originate/exosphere-sdk#155](https://github.com/Originate/exosphere-sdk/issues/155)

When the service comes online, the `'online-text` of `service.yml` is printed. Whereas previously the text had to be copied over, it is now read from `service.yml` itself.

@kevgo @hugobho @martinjaime


kevgo (approved):
Nice!

PR #59, opened 2016-12-14
Append stack trace to error notification in exo-run

trushton:
resolves [Originate/exosphere-sdk#33](https://github.com/Originate/exosphere-sdk/issues/33)

A stack trace was previously not being shown with the error notification when shutting down the application. 

@kevgo @hugobho @martinjaime

### DIFF
--- a/exo-run/src/docker-runner.ls
+++ b/exo-run/src/docker-runner.ls
@@ -53,7 +53,7 @@ class DockerRunner extends EventEmitter
 
 
   _on-container-error: ~>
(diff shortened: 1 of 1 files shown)


kevgo (commented):
What this ticket requires is that when a service crashes (and shows a stacktrace for it), that `exo-run` immediately stops, so that the developer sees the stack trace. 

If this already happens, you can simply close the issue and this PR.

Let's not show a stacktrace within the Exosphere framework itself.

trushton:
Ah okay, yes `exo-run` already immediately stops when an error is thrown, this can be closed.

PR #62, opened 2016-12-15
Ignore dotfiles in in watcher (exo-run)

trushton:
Services were being restarted, and their docker images rebuilt, if a `.swp` file was created when opening a file to work on. This change makes it so that those swapfiles are ignored by the watcher.

@kevgo @martinjaime

### DIFF
--- a/exo-run/src/service-runner.ls
+++ b/exo-run/src/service-runner.ls
@@ -38,8 +38,9 @@ class ServiceRunner extends EventEmitter
         ..start-service!
         ..on 'online', -> done?!
         ..on 'error', (message) ~> @emit 'error', error-message: message
-
-    @watcher = watch @config.root, ignore-initial: yes, ignored: [/.*\/node_modules\/.*/, /(^|[\/\\])\../] # Ignores any sub path including dotfiles e.g. service/.cli/.swp
+        /* Ignores any sub-path including dotfiles.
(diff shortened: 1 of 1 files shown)


exo-run/src/service-runner.ls
```diff
@@ -39,7 +39,7 @@ class ServiceRunner extends EventEmitter
         ..on 'online', -> done?!
         ..on 'error', (message) ~> @emit 'error', error-message: message
 
-    @watcher = watch @config.root, ignore-initial: yes, ignored: /.*\/node_modules\/.*/
+    @watcher = watch @config.root, ignore-initial: yes, ignored: [/.*\/node_modules\/.*/, /(^|[\/\\])\../]
```
kevgo: I don't understand this regex. Can you please shed some light on it, ideally with a comment that explains it?

trushton:
@kevgo The regex matches any sub-path from the given directory that includes a hidden file or folder. The `[\/\\]` section accounts for it being either a unix or windows system, the `\.` matches a single `.` after the slash, and the final `.` catches any character.

kevgo:
Great! Please add this as a comment, so that future generations of Exonauts have access to this secret knowledge as well!  :)

trushton:
@kevgo Already done! Or do you want the specific details in the in-line comment?

kevgo:
Yes, please add the details into the comment. I think this will really help people understand this regex in case they need to fix a bug in it.

kevgo (approved):

PR #63, opened 2016-12-15
Add docker_image field to application.yml in example apps

trushton:
@kevgo

### DIFF
--- a/exosphere-shared/example-apps/crashing-service/application.yml
+++ b/exosphere-shared/example-apps/crashing-service/application.yml
@@ -6,5 +6,7 @@ services:
   public:
     crasher:
       location: ./crasher
+      docker_image: exospheredev/test-service-that-crashes-on-startup
     runner:
       location: ./runner
+      docker_image: exospheredev/test-service-that-keeps-running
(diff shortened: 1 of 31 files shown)


kevgo (approved):

exosphere-shared/example-apps/crashing-service/application.yml
```diff
@@ -6,5 +6,7 @@ services:
   public:
     crasher:
       location: ./crasher
+      docker_image: exospheredev/test-service-that-crashes-on-startup
```
kevgo, line 9: Just out of curiosity, aren't we also specifying the docker image in `service.yml`? Given that, why is it necessary to specify them here as well? Can't we do either `location` that points to a directory with `service.yml` that contains the Docker image, or if one doesn't have the source code checked out, then they need to provide a `docker_image` key here. But not both.

trushton: That was my initial thought when we were having the discussion on this. The author and title are listed in `service.yml`, because if we are developing a service, we need to identify it. The `docker_image` field of `application.yml` is to tell exosphere what services are being used, whether you have them locally or not, they should be available as docker_images.

Location is needed when using services that are in development, for use in commands like `exo-setup`, `exo-lint`, etc. But the `docker_image` field is needed for actually running. I think this actually exposes an issue for `exo-setup`. I'll put in the ticket now.

kevgo: Okay, makes sense as a first and most simple solution. Thanks for the explanation!

PR #69, opened 2016-12-16
Add method to ensure container is running

trushton:
Initial working version of a method to make sure that dependencies of an application's services have app-specific docker containers running.  

@kevgo @hugobho @martinjaime

### DIFF
--- a/exosphere-shared/src/docker-helper.ls
+++ b/exosphere-shared/src/docker-helper.ls
@@ -9,10 +9,10 @@ class DockerHelper
     child_process.exec-sync('docker ps -a --format {{.Names}}', 'utf8') |> (.includes container)
 
 
-  @ensure-container-is-running = (container, image) ->
-    return if child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}' , 'utf8') |> (.includes "#{container}/Up")
-    @remove-container that if @container-exists container
-    @run-image container, image
+  @ensure-container-is-running = (container-name, image) ->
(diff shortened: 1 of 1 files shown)


exosphere-shared/src/docker-helper.ls
```diff
@@ -9,6 +9,12 @@ class DockerHelper
     child_process.exec-sync('docker ps -a --format {{.Names}}', 'utf8') |> (.includes container)
 
 
+  @ensure-container-is-running = (container, image) ->
```
kevgo: This is the name of the container, and not the container itself, right? Let's call this variable `container-name`.

exosphere-shared/src/docker-helper.ls
```diff
@@ -9,6 +9,12 @@ class DockerHelper
...
+    return if child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}' , 'utf8') |> (.includes "#{container}/Up")
+    @remove-container that if @container-exists container
```
kevgo: What is `that` here? Where does it come from?

kevgo (approved):

exosphere-shared/src/docker-helper.ls
trushton: From the Livescript documentation, "`that` refers implicitly to the value of the condition. It will unwrap existence checks."... I was using it incorrectly here. 

I changed this line to restart the container in this case anyway. The first line returns if the container is running, the second restarts the container if it exists but has been stopped, and the third runs the container if it does not exist at all.

kevgo: This is an amazing piece of syntax sugar that I wasn't aware of! It can really help reduce redundancy in a lot of places. Let's apply it more often!

PR #70, opened 2016-12-16
Create dependency docker containers in exo-run

trushton:
resolves [Originate/exosphere-sdk#183](https://github.com/Originate/exosphere-sdk/issues/183)

Reads the link field of each service's `service.yml` file and makes sure that docker containers are running for each of the specified dependencies. The ip address of the container is then passed to the service itself so that it can connect to the running docker container.

The naming convention for these dependency containers is a "dashified" `app-name`, appended with the name of the dependency, e.g. `space-tweet-mongo`.

@kevgo @hugobho @martinjaime

### DIFF
--- a/exo-run/src/service-runner.ls
+++ b/exo-run/src/service-runner.ls
@@ -25,6 +25,7 @@ class ServiceRunner extends EventEmitter
     @docker-config =
       author: @service-config.author
       image: dashify @service-config.title
+      app-name: dashify @config.app-name
       start-command: @service-config.startup.command
       start-text: @service-config.startup['online-text']
       cwd: @config.root
--- a/exo-run/src/service-runner.ls
+++ b/exo-run/src/service-runner.ls
@@ -32,8 +32,8 @@ class ServiceRunner extends EventEmitter
       env:
         EXOCOM_PORT: @config.EXOCOM_PORT
         SERVICE_NAME: @name
-      publish: @service-config.docker.publish if @service-config.docker
-      dependencies: @service-config.docker.dependencies if @service-config.docker
+      publish: @service-config.docker?.publish
+      dependencies: @service-config.docker?.dependencies
 
     @docker-runner = new DockerRunner {@name, @docker-config, @logger}
         ..start-service!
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -5,6 +5,7 @@ dist: trusty
 
 services:
   - docker
+  - mongodb
```
kevgo, line 8: Why do we need MongoDB running on Travis? I thought we run Mongo in a Docker image? Let's do that here as well, to make sure that running Mongo in Docker works.

.travis.yml
```diff
@@ -25,6 +26,7 @@ install:
   - 'cd ./exosphere-shared'
   - './bin/setup'
   - 'cd ..'
+  -
```
kevgo, line 29: remove empty point here

exo-run/src/service-runner.ls
```diff
@@ -25,14 +25,15 @@ class ServiceRunner extends EventEmitter
...
+      app-name: dashify @config.app-name
       start-command: @service-config.startup.command
       start-text: @service-config.startup['online-text']
       cwd: @config.root
       env:
         EXOCOM_PORT: @config.EXOCOM_PORT
         SERVICE_NAME: @name
       publish: @service-config.docker.publish if @service-config.docker
-      link: @service-config.docker.link if @service-config.docker
+      dependencies: @service-config.docker.dependencies if @service-config.docker
```
kevgo: You can simplify this to

```ls
dependencies: @service-config.docker?.dependencies

File: exosphere-shared/src/docker-helper.ls [hunk 4]
```diff
@@ -9,12 +9,18 @@ class DockerHelper
     child_process.exec-sync('docker ps -a --format {{.Names}}', 'utf8') |> (.includes container)
 
 
+  @ensure-container-is-running = (container-name, image) ->
+    return if child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}' , 'utf8') |> (.includes "#{container-name}/Up")
+    return child_process.exec-sync("docker start #{container-name}") if @container-exists container-name
+    @run-image container-name, image
```
kevgo, line 15: Does it make sense to wait here until the image is actually running, or should that happen somewhere else?

kevgo (approved):

PR #71, opened 2016-12-16
Build cli and run end-to-end tests

trushton:
resolves [Originate/exosphere-sdk#203](https://github.com/Originate/exosphere-sdk/issues/203)

Adds a `bin/setup` script to get all of the commands setup and built, and then finally builds the cli using Webpack. Additionally adds a `bin/build` script to just recompile the cli `bundle.js`.

Finally, a few very small miscellaneous changes were made:
1. Line 60 of cli/features/tutorial.feature is needed now that the docker_image field has been added to application.yml.

2. Line 135 of cli/features/steps/steps.ls was changed from expecting `exo-run` to `Running`, because `Running` is the first thing printed by `exo-run` that is indicative of the command having been called.

3. `exo-deploy` was still using an npm version of `exosphere-shared` in one file, so it was changed to the local version.

@kevgo @hugobho @martinjaime


kevgo:
Pro tip: add the comments around certain changes as comments inside the code diff, i.e. similar to if you were a reviewer of the code.

.travis.yml
```diff
@@ -5,6 +5,7 @@ dist: trusty
 
 services:
   - docker
+  - mongodb
```
kevgo: Why mongo here? Travis should use the Mongo Docker image

.travis.yml
```diff
@@ -25,6 +26,7 @@ install:
   - 'cd ./exosphere-shared'
   - './bin/setup'
   - 'cd ..'
+  -
```
kevgo: remove empty line

cli/bin/setup
```diff
@@ -1,3 +1,14 @@
 #!/usr/bin/env bash
-echo "disabled until ZMQ is removed"
-# npm install --loglevel error --depth 0
+
+npm install --loglevel error --depth 0
+(cd ../exo-add && bin/setup && node_modules/.bin/build && echo 'exo-add √')
```
kevgo, line 4: Use [Morula](https://github.com/originate/morula) here so that when we add/rename/remove projects, this file still works?

cli/bin/setup
```diff
@@ -1,3 +1,14 @@
...
+(cd ../exo-clone && bin/setup && node_modules/.bin/build && echo 'exo-clone √')
+(cd ../exo-create && bin/setup && node_modules/.bin/build && echo 'exo-create √')
```
kevgo, line 6: Cool idea to print the checkmark at the end!

cli/features/steps/steps.ls
```diff
@@ -130,9 +130,9 @@ module.exports = ->
       jsdiff-console actual-content.to-string!trim!, expected-content.trim!, done
 
 
-  @Then /^the full command "([^"]*)" is executed$/ timeout: 5_000, (command, done) ->
+  @Then /^the full command "([^"]*)" is executed$/ timeout: 20_000, (command, done) ->
```
kevgo: I've been running into repeated timeouts with this setting on my machine. Downloading the Docker images can take a while. Let's go with 60 seconds here.

exosphere-shared/src/docker-helper.ls
```diff
@@ -9,12 +9,18 @@ class DockerHelper
     child_process.exec-sync('docker ps -a --format {{.Names}}', 'utf8') |> (.includes container)
 
 
+  @ensure-container-is-running = (container-name, image) ->
+    return if child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}' , 'utf8') |> (.includes "#{container-name}/Up")
```
kevgo: This method still implements more than one thing: checking whether a docker image is running, starting a docker image, and doing its actual business logic (what its name says). 

Please extract a `container-is-running` method and a `start-container` method. Then use pattern matching to format this method more readable:

```ls
@ensure-container-is-running = (container-name, image) ->
  | @container-is-running container-name  =>  return
  | @container-exists container-name      =>  @start-container container-name
  | otherwise                             =>  @run-image container-name, image
```

This format reads almost like plain English, and tells exactly what this method is doing!

exosphere-shared/src/docker-helper.ls
```diff
@@ -9,12 +9,18 @@ class DockerHelper
...
+    @run-image container-name, image
+
+
   @get-build-command = (image, build-flags) ->
     return "docker build -t #{image.author}/#{image.name} #{if build-flags then build-flags else ""} ."
 
 
   @get-config = (image) ->
-    return child_process.exec-sync("docker run #{image} cat service.yml", 'utf8') |> (.to-string!)
+    return child_process.exec-sync("docker run --rm=true #{image} cat service.yml", 'utf8') |> (.to-string!)
```
kevgo: Omit return here

cli/bin/setup
trushton: The only problem with using Morula here is that this is inside of `cli`s `bin/setup` and if we tell Morula to execute `bin/setup` for each sub-repo, we will get caught in a loop. I could change the name of this script to something else to resolve this?

kevgo: The `setup` script is only to set up that particular code base, not the entire app. I like the idea of having a script that sets up the entire app, though. Let's call it `setup-all` and put it in the root directory of the monorepo! Then it can use Morula.

.travis.yml
trushton: Taken care of in #82.

trushton:
@kevgo

cli/bin/setup
```diff
@@ -1,2 +1,14 @@
 #!/usr/bin/env bash
+
 npm install --loglevel error --depth 0
+(cd ../exo-add && bin/setup && node_modules/.bin/build && echo 'exo-add √')
```
kevgo, line 4: Since this is setting up all subprojects, let's move this script to the root folder.

cli/bin/setup
```diff
@@ -1,2 +1,14 @@
...
+(cd ../exo-clone && bin/setup && node_modules/.bin/build && echo 'exo-clone √')
```
kevgo, line 5: Use Morula

```
cd .. && morula all bin/setup && morula all node_modules/.bin/build
```

This runs these commands in all subprojects

kevgo: This can be done in a separate PR btw

exo-run/src/docker-runner.ls
```diff
@@ -67,5 +66,12 @@ class DockerRunner extends EventEmitter
         @emit 'online'
 
 
+  _check-dependency-containers: ~>
+    for dependency in @docker-config.dependencies
+      app-dependency = "#{@docker-config.app-name}-#{dependency}"
+      DockerHelper.ensure-container-is-running app-dependency, dependency
+      @docker-config.env["#{dependency.to-upper-case!}"] = DockerHelper.get-docker-ip app-dependency
```
kevgo: No point in wrapping this into a string.

```ls
@docker-config.env[dependency.to-upper-case!] =

kevgo (approved):
This PR seems to have a ton of changes from other PRs. git sync?

PR #74, opened 2016-12-19
Replace npm with yarn in exo-sync

martinjaime:
Addresses issue Originate/exosphere-sdk#170


trushton (approved):

exo-sync/appveyor.yml
```diff
@@ -3,13 +3,13 @@ environment:
...
 install:
   - ps: Install-Product node $env:nodejs_version
-  - npm install --loglevel error --depth 0
+  - yarn install
   - git config --global user.name "Appveyor"
   - git config --global user.email ci@appveyor.com
 
 test_script:
   - node --version
-  - npm --version
```
martinjaime, line 12: Actually, shouldn't we keep `npm` here? since we use it for publishing?

exo-sync/bin/setup
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env bash
-npm install --loglevel error --depth 0
+yarn install
```
kevgo, line 12: Check whether yarn is installed here, and if not install it by running `npm i -g yarn`?

martinjaime: @kevgo, i think we should leave installing `yarn` up to the user since its installation is [different for each platform](https://yarnpkg.com/en/docs/install#mac-tab).

exo-sync/appveyor.yml
kevgo: Yes, if we still use it, let's keep it here. Meaning have both here.

exo-sync/bin/setup
kevgo: `npm i -g yarn` works on all platforms. The different solutions mentioned in your link are for when you don't have NPM.

martinjaime: I've found that yarn developers strongly recommend against installing through npm. should we install through npm anyway? Or how about install it locally instead?

kevgo: As agreed on the check-in, let's not install yarn here, but show an error message and abort.

PR #75, opened 2016-12-19
Fix bug in DockerHelper

trushton:
resolves [Originate/exosphere-sdk#203](https://github.com/Originate/exosphere-sdk/issues/203)

There was a bug in the `container-exists` method of DockerHelper that was incorrectly returning true sometimes. 

The issue arose when a name of a container was a subset of another. E.g. Docker container `users-service` exists and we are checking for `users`. We would receive true, and then potentially attempt to remove the `users` container, resulting in an error as it doesn't exist. This has been fixed now.

@kevgo @martinjaime

### DIFF
--- a/exosphere-shared/src/docker-helper.ls
+++ b/exosphere-shared/src/docker-helper.ls
@@ -40,6 +40,9 @@ class DockerHelper
 
 
   @remove-container = (container) ->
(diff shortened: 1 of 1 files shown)


trushton:
@kevgo All tests, including end-to-end tests, are green!

kevgo (approved):

PR #76, opened 2016-12-19
Add yarn badge to README

martinjaime:
forgot to add yarn badge to README


kevgo (approved):

exo-test/bin/setup
```diff
@@ -1,2 +1,12 @@
 #!/usr/bin/env bash
-yarn install
+
+# check if yarn is installed
+command -V yarn >/dev/null 2>&1
```
alexdavid, line 4: Are you just testing for the existence of `yarn` here? This isn't posix compliant. It's better to just use `which`.

```shell
if which yarn; then
  yarn install
else
  echo "ERROR: yarn is not installed."
  echo "       visit https://yarnpkg.com/en/docs/install to see installation instructions based on your OS"
  exit 1
fi

kevgo: @alexdavid this is an old PR, Martin is no longer on the team. Can you please implement the changes yourself and 🚢 this?

alexdavid: Actually looks like this is already using `which`.

Going to close this PR

PR #77, opened 2016-12-19
Replace npm with yarn in exo-clone

martinjaime:
Addresses Originate/exosphere-sdk#170


kevgo (commented):
Same here, please install yarn in case it is missing

PR #78, opened 2016-12-19
Replace npm with yarn in exo-lint

martinjaime:
Addresses Originate/exosphere-sdk#170


martinjaime:
@kevgo, seems like tests are failing due to docker issues.

exo-lint/bin/setup
```diff
@@ -1,2 +1,13 @@
 #!/usr/bin/env bash
-npm install --loglevel error --depth 0
+
+# check if yarn is installed
+command -V yarn >/dev/null 2>&1
```
kevgo: Is `command` available on all platforms? What about using `which yarn` here?

exo-lint/bin/setup
```diff
@@ -1,2 +1,13 @@
...
+
+if [ $? -eq 0 ]; then
```
kevgo: Check for the error first and exit, then do the actual thing later.

```bash
if [ $? -gt 0 ]; then
  echo "ERROR...
  exit 1
fi

yarn install

PR #79, opened 2016-12-19
Use mongo Docker image in Travis

trushton:
Previously we were requiring `mongodb` as a service to be running in Travis, for part of the tutorial feature. Now, when running the feature tests for the service created in `tutorial.feature`, a mongo container is started up beforehand with port `27017` exposed as to be accessible through `localhost`. It is then removed afterwards as it is no longer needed from that point. This can be changed to a more elegant solution in the future as we create the fully fledged `tutorial runner`.

@kevgo @martinjaime

### DIFF
--- a/cli/features/tutorial.feature
+++ b/cli/features/tutorial.feature
@@ -134,7 +134,9 @@ Feature: Following the tutorial
           - 'mongo'
       """
     When running "exo setup" in this application's directory
+    And setting up a temporary mongo docker container
     And running "exo test" in this application's directory
+    And removing the temporary mongo docker container
     Then it prints "todo-service works" in the terminal
     And it prints "html-server has no tests, skipping" in the terminal
     And it prints "All tests passed" in the terminal
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ dist: trusty
 
 services:
   - docker
-  - mongodb
 
 node_js:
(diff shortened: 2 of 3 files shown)


cli/features/tutorial.feature
```diff
@@ -134,7 +134,9 @@ Feature: Following the tutorial
           - 'mongo'
       """
     When running "exo setup" in this application's directory
+    And setting up a temporary mongo docker container
```
kevgo, line 137: In my understanding, `exo test` should start and stop the Mongo Docker container. Because it sees that a service needs MongoDB when looking through the `service.yml` files.

trushton: Created #81 to track this.

kevgo: Great, thanks! How much sense does this PR with #81 in place? Can we do #81 instead of this PR?

trushton: We could do #81 instead, the only reason for this PR would be if we wanted to get #71 through more quickly and then remove this change later on when the changes in #81 are implemented.

kevgo: #71 is already a PR, so let's try to get this in. I'd like to avoid the detour with subsequent need to remove this if possible

trushton:
Closing this as #82 addresses the issues commented on.

PR #80, opened 2016-12-19
Change link to dependency

trushton:
Change the field `link` to `dependencies` in `exo-add` and `exo-create` feature tests, as well as in `exo-deploy`. 

@kevgo @hugobho @martinjaime

### DIFF
--- a/exo-add/features/exoservice-es6-mongodb.feature
+++ b/exo-add/features/exoservice-es6-mongodb.feature
@@ -54,7 +54,7 @@ Feature: scaffolding an ExoService written in ES6, backed by MongoDB
           - user.updated
 
       docker:
-        link:
+        dependencies:
           - 'mongo'
       """
     And my application contains the file "user-service/src/server.js"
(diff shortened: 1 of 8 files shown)


exo-add/features/exoservice-es6-mongodb.feature
```diff
@@ -54,7 +54,7 @@ Feature: scaffolding an ExoService written in ES6, backed by MongoDB
           - user.updated
 
       docker:
-        link:
+        dependencies:
```
kevgo, line 57: This should be a top-level key.

```yml
name: ...

messages: ...

dependencies: 
  - mongo

trushton: My thought was that seeing as these are Docker images that the service requires to have containers running for (could be specifying a certain Docker image version), it would make sense to keep these dependencies listed in the docker section, along with anything else Docker related such as ports to be published, etc. 

What do you think?

kevgo: The service just declares that it has a dependency on "MongoDB". In dev mode that's provided as a Docker image. In production it is provided as a link to an external, possibly third-party system. So this really isn't a Docker thing at all, it's just coincidentally Docker in dev mode.

trushton:
Closing this as #83 resolves the issues commented on.

PR #83, opened 2016-12-20
Move 'dependencies' field to root of 'service.yml'

trushton:
Additionally makes any changes across repos to account for that change to the app templates, such as updating what is expected in feature tests, and how the field is accessed in `exo-run`

@kevgo @martinjaime


trushton:
@kevgo

kevgo (approved):

PR #86, opened 2016-12-23
Change access-sync to exists-sync

trushton:
@kevgo @hugobho

### DIFF
--- a/exo-setup/src/docker-setup.ls
+++ b/exo-setup/src/docker-setup.ls
@@ -39,7 +39,7 @@ class DockerSetup extends EventEmitter
 
   _docker-file-exists: ~>
     try
-      fs.access-sync path.join(@config.root, 'Dockerfile')
+      fs.exists-sync path.join(@config.root, 'Dockerfile')
     catch
       no
 


exo-setup/src/docker-setup.ls
```diff
@@ -39,7 +39,7 @@ class DockerSetup extends EventEmitter
 
   _docker-file-exists: ~>
     try
-      fs.access-sync path.join(@config.root, 'Dockerfile')
+      fs.exists-sync path.join(@config.root, 'Dockerfile')
```
kevgo, line 42: Exists is deprecated. Please use `fs.stat-sync` here

trushton: From the docs: 
`Note that fs.exists() is deprecated, but fs.existsSync() is not. (The callback parameter to fs.exists() accepts parameters that are inconsistent with other Node.js callbacks. fs.existsSync() does not use a callback.)`

 Would you prefer we still go with `stat-sync`?

kevgo: Ha, great find! I missed that. Good as is then!

kevgo (approved):

hugoesthere (approved):

PR #87, opened 2016-12-23
Don't check in the webpack output into Git

kevgo:
We agreed to not check in compiled assets into Git. I think that also applies to Webpack output.

@trushton @alexdavid


trushton (approved):
👍

PR #88, opened 2016-12-23
Use the webpack binary

kevgo:
Makes the CLI call the web pack binary now, instead of the manually transpiled code.
In order to do that, also ships the web pack binary in the NPM module now.

@trushton @alexdavid


kevgo:
When running the e2e tests (`spec features/tutorial.feature`), it produces this error: 

```
Cannot find module './findout'
```

@trushton can you please investigate and fix this issue? It is a dependency of tmplconv. Web pack correctly bundles it. It seems like it is incompatible with Webpack, since it [requires files via a variable instead of a constant](https://github.com/okunishinishi/node-findout/blob/master/lib/findout.js#L20). So I'm okay with throwing out tmplconv if that fixes the issue, this library has been problematic in other ways as well.

trushton:
@kevgo Ah, that can be resolved by removing `/findout/` from the `noParse` field in `webpack-config.ls`. It will produce a couple of warnings by doing so but everything will work correctly.

kevgo:
@trushton can you please take over this branch - or create your own branch with similar changes - and make this work?

PR #89, opened 2016-12-27
Wait for dependencies to start

hugoesthere:
`exo-tests` starts a Mongo container if a service lists 'mongo' as a dependency. The Docker containers were being started but Mongo itself was taking longer to boot up within the container, so tests that use Mongo would fail with this error: `MongoError: connection 0 to localhost:27017 closed`

@kevgo @trushton

### DIFF
--- a/exo-test/src/service-tester.ls
+++ b/exo-test/src/service-tester.ls
@@ -20,8 +20,7 @@ class ServiceTester extends EventEmitter
       @emit 'service-tests-skipped', @name
       return done?!
 
-    @_start-dependencies!
-    wait 500, ~>
+    @_start-dependencies ~>
       new ObservableProcess(call-args(@_create-command @service-config.tests)
                             cwd: @config.root,
                             env: @config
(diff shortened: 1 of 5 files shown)


exo-test/src/service-tester.ls
```diff
@@ -19,29 +20,30 @@ class ServiceTester extends EventEmitter
...
   remove-dependencies: ~>
     for dep in @service-config.dependencies or []
       DockerHelper.remove-container "test-#{dep}"
 
 
-  _start-dependencies: ~>
+  _start-dependencies: (done) ~>
     for dep in @service-config.dependencies or []
       DockerHelper.ensure-container-is-running "test-#{dep}", dep
+    wait 500, done
```
kevgo, line 47: Why do you wait half a second here?

kevgo (approved):

exo-test/src/service-tester.ls
```diff
@@ -19,29 +20,30 @@ class ServiceTester extends EventEmitter
...
-                          stdout: {@write}
-                          stderr: {@write})
-      ..on 'ended', (exit-code) ~>
-        if exit-code > 0
-          @emit 'service-tests-failed', @name
-        else
-          @emit 'service-tests-passed', @name
-        @remove-dependencies!
-        done?(null, exit-code)
+    @_start-dependencies ~>
```
kevgo: This should receive an error in case something went wrong.

kevgo (approved):

exo-test/src/service-tester.ls
hugoesthere: I've created an issue to address this in the future: https://github.com/Originate/exosphere/issues/91

PR #90, opened 2016-12-27
Update exo-test to use new service.yml format

hugoesthere:
using this format:
```
dependencies:
  mongo:
    prod: 'ds143588.mlab.com:43588'
```

@kevgo @trushton

### DIFF
--- a/exo-test/src/service-tester.ls
+++ b/exo-test/src/service-tester.ls
@@ -36,12 +36,12 @@ class ServiceTester extends EventEmitter
 
 
   remove-dependencies: ~>
-    for dep in @service-config.dependencies or []
+    for dep of @service-config.dependencies
       DockerHelper.remove-container "test-#{dep}"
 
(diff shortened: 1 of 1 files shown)


kevgo (commented):
Shouldn't this change also require some changes in a `service.yml` template in tests? Or were the tests broken before?

hugoesthere:
These tests were broken from before. We aren't compiling the source code in `bin/setup`.

But there are a number of other changes that have to be made to this monorepo to account for the new `service.yml` format. I was planning on extracting those changes in another PR, I've only made these changes so that I can test that the new users/tweets services (which use this new service.yml format) pass via `exo-test`.

hugoesthere:
I've made a ticket to address the format change to `service.yml`: https://github.com/Originate/exosphere/issues/92
@kevgo

kevgo (approved):

PR #96, opened 2016-12-28
Use dashify

hugoesthere:
Remain consistent with how images names are generated in `exo-setup`

@kevgo @trushton


kevgo (approved):
Very cool! Yes, leaning on proven third-party libraries is much better than reinventing the wheel.

trushton (approved):

PR #97, opened 2016-12-28
Teardown bug

hugoesthere:
Incorrect message being printed

@kevgo @trushton

### DIFF
--- a/exo-deploy/src/deployers/aws-deployer.ls
+++ b/exo-deploy/src/deployers/aws-deployer.ls
@@ -54,7 +54,7 @@ class AwsDeployer
 
   teardown: ({nuke}, done) ->
     @terraform-file-builder.generate-provider-credentials!
-    process.stdout.write "terraform starting nuke from AWS"
+    process.stdout.write "terraform starting teardown from AWS"
     @terraform.destroy (err) ~>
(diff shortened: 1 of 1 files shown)


exo-deploy/src/deployers/aws-deployer.ls
```diff
@@ -54,7 +54,7 @@ class AwsDeployer
 
   teardown: ({nuke}, done) ->
     @terraform-file-builder.generate-provider-credentials!
-    process.stdout.write "terraform starting nuke from AWS"
+    process.stdout.write "terraform starting teardown from AWS"
```
kevgo: Is it a teardown __from__ AWS of a teardown __of__ AWS?

hugoesthere: Well, it's actually both. If `nuke` is set in the parameters, it's teardown of application infrastructure including the hosted zone. If not, it's teardown from AWS. That distinction is already made in earlier print statements, this statement is just supposed to say "we're starting teardown now".

kevgo: Maybe say "terraform starting AWS teardown" then? "teardown from" sounds weird, grammar wise

hugoesthere:
@kevgo what do you think?

kevgo (approved):
Much better now!

PR #100, opened 2016-12-28
Set node env

hugoesthere:
The base image for the node services (`nodesource/node`) sets `NODE_ENV=prod` by default. In `exo-run` we need `NODE_ENV=dev`. 

However, this change isn't language agnostic. We need some way to determine what language a service is written in and set the necessary environment variables. Perhaps we could do something like this in `service.yml`:
```
development:
  docker:
    env:
      - 'NODE_ENV=dev'
    publish:
      - '3000:3000'
```

@kevgo @trushton

### DIFF
--- a/exo-run/src/docker-runner.ls
+++ b/exo-run/src/docker-runner.ls
@@ -41,7 +41,7 @@ class DockerRunner extends EventEmitter
   _create-run-command: ->
     command = "
       docker run 
-        --name=#{@docker-config.env.SERVICE_NAME} "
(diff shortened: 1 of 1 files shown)


kevgo:
I don't think hard-coding Node-specific env variables into exo-run will get us very far. Let's do what @hugobho proposes in the comment and add a section with env variables to service.yml.

kevgo:
image

hugoesthere:
This has been resolved for the time being: https://github.com/Originate/space-tweet/pull/23

PR #101, opened 2016-12-28
Use updated ExoCom in exo-run test

hugoesthere:
Since updating the `originate/exocom` docker image, this step definition is no longer waiting for the correct string. 

@kevgo @trushton


trushton:
@hugobho I think the reason that this was failing is that the `latest` version of exocom (which exo-setup is pulling) was still the `0.15.1` version. I just pushed `0.15.2` as the latest to dockerhub, so re-running the tests should resolve this.

hugoesthere:
Yeah it was related to the versioning. I had actually pushed `0.15.2` a couple days ago which is why in this PR I had to change it from `0.15.1` to `0.15.2`. But regardless I am working on a PR for `observable-process` to search the text stream with regex so this PR will be redone with that.

hugoesthere:
@kevgo ready for another look

kevgo (approved):

PR #105, opened 2016-12-28
Docker refactor

hugoesthere:
The exo-deploy source code was getting confusing because some code was being run inside a Docker container while some just locally. This refactor makes that distinction clear: all code inside the `dockerized` dir is executed in a Docker container, everything else is executed locally.

@kevgo @trushton


kevgo (approved):
Awesome change, thanks for doing this!

PR #106, opened 2016-12-29
Fix exo-test

trushton:
We were addressing the `dependencies` field incorrectly, seeing as it is an array, not an object itself.

@kevgo @hugobho

### DIFF
--- a/exo-test/src/service-tester.ls
+++ b/exo-test/src/service-tester.ls
@@ -37,14 +37,18 @@ class ServiceTester extends EventEmitter
 
 
   remove-dependencies: ~>
-    for dep of @service-config.dependencies
-      DockerHelper.remove-container "test-#{dep}"
+    if @service-config.dependencies
+      for dep in @service-config.dependencies
+        DockerHelper.remove-container "test-#{dep}"
 
 
   _start-dependencies: (done) ~>
-    for dep of @service-config.dependencies
-      DockerHelper.ensure-container-is-running "test-#{dep}", dep
-    wait 500, done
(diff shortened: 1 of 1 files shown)


hugoesthere:
`dependencies` looks like this now, no?
```
dependencies:
  mongo:
    prod: mongo_uri
  some-other-dependency:
    prod: dependency_uri
```
which is a map

trushton:
Ah, I was unaware of that change, needed to do a `git sync`

exo-test/src/service-tester.ls
```diff
@@ -37,14 +37,18 @@ class ServiceTester extends EventEmitter
 
 
   remove-dependencies: ~>
-    for dep of @service-config.dependencies
-      DockerHelper.remove-container "test-#{dep}"
+    if @service-config.dependencies
```
kevgo, line 40: Just say

```ls
for dep in @service-config.dependencies or []
```

to deal with cases where it doesn't exist.

exo-test/src/service-tester.ls
```diff
@@ -37,14 +37,18 @@ class ServiceTester extends EventEmitter
...
+      for dep in @service-config.dependencies
+        DockerHelper.remove-container "test-#{dep}"
 
 
   _start-dependencies: (done) ~>
-    for dep of @service-config.dependencies
-      DockerHelper.ensure-container-is-running "test-#{dep}", dep
-    wait 500, done
+    if @service-config.dependencies
```
kevgo, line 46: same here

PR #107, opened 2016-12-29
Publishing without Webpack  (WIP)

kevgo:
This is a work in progress. Please comment.

- exports the whole monorepo in the NPM module (compiled as `dist`)
- creates a `package.json` file in the root that exports the `exo` command from the `dist` subproject
- all the code can require other files in the monorepo similar to how it is done in development
- we still have the folder structure, so tmplconv should work

@trushton


cli/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "exosphere-sdk",
+  "name": "exosphere-cli",
```
kevgo, line 2: This package.json is now only for the cli subproject.

package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "exosphere-sdk",
```
kevgo, line 2: This is the new package.json for the whole monorepo. It is used only for publishing.

package.json
```diff
@@ -0,0 +1,39 @@
...
+  "bin": {
+    "exo": "cli/bin/exo"
+  },
+  "dependencies": {},
+  "description": "SDK for the Exosphere framework",
+  "devDependencies": {},
+  "engines": {
+    "npm": ">= 3.0.0"
+  },
+  "files": [
```
kevgo, line 14: All these folders get exported into the NPM module

package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "exosphere-sdk",
+  "version": "0.9.0",
+  "author": "Originate",
+  "bin": {
+    "exo": "cli/bin/exo"
```
kevgo, line 6: Here we export the `exo` command

package.json
```diff
@@ -0,0 +1,39 @@
...
+    "exo-add/dist",
+    "exo-clone/dist",
+    "exo-create/dist",
+    "exo-deploy/dist",
+    "exo-lint/dist",
+    "exo-run/dist",
+    "exo-setup/dist",
+    "exo-sync/dist",
+    "exo-test/dist",
+    "exosphere-shared/dist"
```
trushton: Won't we also need the `exosphere-shared/templates` folder available so that `exo-add`, `exo-create`, and `exo-setup` can use it?

kevgo: Good point, added

trushton (approved):

PR #108, opened 2016-12-29
Use Morula for setup

kevgo:
Hard-coding all the project names in several different places creates significant risk around forgetting to update one of those places when we add/remove/change a subproject. Better to determine the sub projects dynamically at runtime.

@trushton @hugobho


.travis.yml
```diff
@@ -27,9 +27,6 @@ before_install:
 
 
 install:
-  - 'cd ./exosphere-shared'
```
kevgo, line 25: Morula runs `exosphere-shared` first already.

cli/bin/setup
```diff
@@ -7,15 +7,4 @@ if [ $? -gt 0 ]; then
...
-(cd ../exo-deploy && bin/setup && echo 'exo-deploy √')
-(cd ../exo-lint && bin/setup && echo 'exo-lint √')
-(cd ../exo-run && bin/setup && echo 'exo-run √')
-(cd ../exo-setup && bin/setup && echo 'exo-setup √')
-(cd ../exo-sync && bin/setup && echo 'exo-sync √')
-(cd ../exo-test && bin/setup && echo 'exo-test √')
-(cd ../exosphere-shared && bin/setup && echo 'exosphere-shared √')
-(node_modules/.bin/build)
-(./bin/build)
+node_modules/.bin/build
```
kevgo, line 10: This repo now only prepares itself, and expects the other repos to be set up as well. The idea here is that if one changes code in this repo, it should only recompile this repo. I'm not 100% sure on this, open for feedback.

.travis.yml
hugoesthere: Where is that defined? I don't think I see it in `morula.yml`

cli/bin/setup
hugoesthere: I think the issue before was that the cli repo uses the rest of the `exo` commands. So when a PR is made with only changes to cli, Morula will only run tests (and `setup`) for cli. That means `setup` will not be run for the rest of the `exo` commands and their `dist` folders will not be available to the cli feature tests.

trushton: Exactly what @hugobho said. On Travis, if we only made changes to the `cli` repo, we still need to setup all of the other repos too (to get their `dist` folders available) as they are required by `cli` and needed to run tests like the `tutorial.feature`.

.travis.yml
kevgo: Excellent catch! Added now.

cli/bin/setup
kevgo: Good catch! It now always sets up all subrepos (in https://github.com/Originate/exosphere/blob/kg-morula/.travis.yml#L30), but only tests the changed ones.

PR #110, opened 2016-12-31
Templates wait for ExoCom to come online

kevgo:
changes from pairing up

@hugobho


hugoesthere (approved):
Will be good once paired up with the PR that actually updates MockExoCom

PR #113, opened 2017-01-03
TR-1: Pull images from dockerhub (exo-setup)

### LINKED ISSUE #66: Determine if service is currently in development (exo-setup)
In `exo-setup`, we currently use the `location` field of `application.yml` as a directory to build a docker image from. To account for cases where a user may be using some pre-made services from the future exosphere "service-store", we should only build the image if the `location` field is there (indicating that the user is actively developing that service), and if not, but there is a `docker_image` field, we should pull the image at that time.

trushton:
resolves #66

If a service in `application.yml` has a `docker-image` field, but no `location`, the image should be pulled from docker hub, but not set up in the way that we do for local services.

@kevgo @hugobho


exo-setup/features/setup.feature
```diff
@@ -25,3 +25,13 @@ Feature: Setup of Exosphere applications
       | test-mongo-db-server   |
       | test-app-web-server    |
       | exocom                 |
+
+  Scenario: set up the "docker-service" application
```
kevgo: Scenario: setting up an application with external Docker images

exo-setup/features/setup.feature
```diff
@@ -25,3 +25,13 @@ Feature: Setup of Exosphere applications
...
+    Given a freshly checked out "docker-service" application
+    When running "exo-setup" in this application's directory
+    Then it has created the folders:
+      | SERVICE       | FOLDER       |
+      | local-service | node_modules |
```
kevgo: Not sure if we need to re-test this here again. I think this scenario is only for verifying that it downloads the Docker images. Let's remove this `Then it has created the folders` step here.

exo-setup/src/docker-setup.ls
```diff
@@ -14,11 +15,21 @@ require! {
 class DockerSetup extends EventEmitter
 
   ({@name, @logger, @config}) ->
-    @service-config = if @config then yaml.safe-load fs.read-file-sync(path.join(@config.root, 'service.yml'), 'utf8')
-
+      @service-config = if @config.root then yaml.safe-load fs.read-file-sync(path.join(@config.root, 'service.yml'), 'utf8')
```
hugoesthere: I think this has an extra tab to it

hugobho:
What happens when both the `location` and `docker-image` fields are listed, does it use only the local version of the service?

exosphere-shared/example-apps/docker-service/application.yml
```diff
@@ -0,0 +1,11 @@
+name: docker-app
+description: An application that relies on dockerhub images
+version: 0.0.1
+
+services:
+  public:
+    local-service:
+      docker_image: exospheredev/local-service
+      location: ./local-service
```
hugoesthere: What happens if both the `docker_image` and `location` field are defined? Does it use only the local version of the service? If so shouldn't we only define one or the other?

trushton: Thats right, it uses the local version. Reason being that if you have the repository locally, it is reasonable to believe that you are developing it and want to use your latest changes.

hugoesthere: What's the point of defining `docker_image` as well then? Isn't the image name generated in `exo-setup`? Having that in two places might be confusing. It seems like it's not being used here anyways?

trushton: Its not being used here, but it would still be beneficial to have the field here for when the time comes to deploy a service to dockerhub, we wouldn't need to generate the name again, it could just be grabbed from here. 

Overall I thought it best to allow for any configuration of having the fields, you could have both and it use the location, if you just have one it uses that method, and if you have neither, an error is signaled.

Does that seem right? or would it be better to have a user remove the `docker_image` if they want to build it themselves?

hugoesthere: I don't think this field can be used to deploy a service to Dockerhub right? Because it's `exo-setup` that generates the Docker images, and that uses a certain naming convention. The user can name the image whatever he/she wants in this `docker_image` field, but that won't be the name of the image that was built in `exo setup` (or the one that will be deployed in `exo deploy`).

I think it makes sense to allow for any configuration, and we can do that by having one of  either `docker_image` or `location`. That would still allow the user to use either a remote docker image or their local service: you list the `location` field, you use a local service; you list the `docker_image` field, you use a remote service. But when both `docker_image` and `location` are included, the `docker_image` field will always be ignored it seems like. Does that make sense? What do you think?

Or maybe it's a bit confusing that we call one of the fields `location` when a service's location can be either local or remote..

exo-setup/src/docker-setup.ls
```diff
@@ -14,11 +15,21 @@ require! {
...
 
   ({@name, @logger, @config}) ->
-    @service-config = if @config then yaml.safe-load fs.read-file-sync(path.join(@config.root, 'service.yml'), 'utf8')
-
+    @service-config = if @config.root then yaml.safe-load fs.read-file-sync(path.join(@config.root, 'service.yml'), 'utf8')
 
   start: (done) ~>
-    | !@_docker-file-exists!  =>  cp path.join(templates-path, 'docker', 'Dockerfile'), path.join(@config.root, 'Dockerfile')
+    | !@service-config  =>
+      if @config.docker-image
```
kevgo: Extract this into a `start-external-service` method to make this more readable.

exosphere-shared/example-apps/docker-service/application.yml
```diff
@@ -0,0 +1,11 @@
+name: docker-app
```
kevgo: Let's call this app `app-with-external-docker-images`. Also change the directory name.

exosphere-shared/example-apps/docker-service/application.yml
```diff
@@ -0,0 +1,11 @@
...
+description: An application that relies on dockerhub images
+version: 0.0.1
+
+services:
+  public:
+    local-service:
+      docker_image: exospheredev/local-service
+      location: ./local-service
+    test:
```
kevgo: There should be an `external-service` entry here, which uses the external Docker image.

exosphere-shared/example-apps/docker-service/application.yml
kevgo: Let's follow the principle of least surprise here, and do the simple and straightforward thing that causes the least amount of confusion. I think that is to have only one or the other setting, but not both. Later we can add extra logic, documentation, and testing to allow for both settings being present at the same time.

exosphere-shared/example-apps/docker-service/local-service/features/create-many.feature
```diff
@@ -0,0 +1,36 @@
+Feature: Creating multiple tests
```
kevgo: Please slim this service down to the simplest possible service. Only one functionality, and only one test for it. It could just be a ping service, which replies "pong" when receiving "ping".

trushton:
@kevgo This should be good now

exo-setup/src/app-setup.ls
```diff
@@ -19,23 +19,25 @@ class AppSetup extends EventEmitter
...
 
-    # Note: Windows does not provide atomic file operations,
-    #       causing file system permission errors when multiple threads read and write to the same cache directory.
-    #       We therefore run only one operation at a time on Windows.
-    operation = if process.platform is 'win32'
-      async.map-series
-    else
-      async.map
-    operation setups, (-> &0.start &1), (err) ~>
+    async.map-series setups, (-> &0.start &1), (err) ~>
```
kevgo: Move this line to line 28, so that you first deal with setups, and then down here with docker-setups.

exo-setup/src/docker-setup.ls
```diff
@@ -14,11 +15,11 @@ require! {
...
 
   ({@name, @logger, @config}) ->
-    @service-config = if @config then yaml.safe-load fs.read-file-sync(path.join(@config.root, 'service.yml'), 'utf8')
-
+    @service-config = if @config.root then yaml.safe-load fs.read-file-sync(path.join(@config.root, 'service.yml'), 'utf8')
 
   start: (done) ~>
-    | !@_docker-file-exists!  =>  cp path.join(templates-path, 'docker', 'Dockerfile'), path.join(@config.root, 'Dockerfile')
+    | !@service-config          =>  return @_setup-external-service done
+    | !@_docker-file-exists!    =>  cp path.join(templates-path, 'docker', 'Dockerfile'), path.join(@config.root, 'Dockerfile')
```
kevgo: Too many spaces before `=>`

exosphere-shared/example-apps/app-with-external-docker-images/local-service/server.ls
```diff
@@ -0,0 +1,9 @@
+module.exports =
+
+  before-all: (done) ->
+    done!
+
+
+  # replies to the "ping" command
```
kevgo: No need for the comment here

exo-setup/features/setup.feature
```diff
@@ -25,3 +25,10 @@ Feature: Setup of Exosphere applications
       | test-mongo-db-server   |
       | test-app-web-server    |
       | exocom                 |
+
+  Scenario: set up an application with external Docker images
+    Given a freshly checked out "app-with-external-docker-images" application
+    When running "exo-setup" in this application's directory
+    Then it has acquired the Docker images:
+      | test-web-server |
+      | local-service   |
```
kevgo: What does "aquired Docker images" mean here? Does it download a Docker image for the `local-service`?

kevgo: I'm not sure we need the local-service here at all. Isn't this scenario supposed to specify how applications with external Docker images work? I think an app with a single service that uses an external Docker image is enough here.

trushton:
@kevgo

kevgo (commented):
Some preliminary comments from last week. This PR is huge now, probably contains a ton of unrelated changes. Can you please sync the branch?

trushton:
I did a `git sync`, I think the issue is that there were sub-branches off of this one that got approved and then merged in, so this one looks as though it has a lot of changes. 
Either way, the comments have been addressed and this should be ready for another look @kevgo

kevgo:
This PR contains elements from #136, which has been merged into `tr-1` instead of master. I think the branch `tr-organize-steps` was a child of this branch. You should have shipped this branch first, then `tr-organize-steps`.

kevgo:
The solution is:
* create a new branch containing the changes from `tr-organize-steps` 
* check out `tr-1` and run `git revert 027304657094aeffe7d5c4407d7c477b34d17cf8`. This should undo the merge of `tr-organize-steps` into this branch here.
* we review this branch and ship it
* you ship the new `tr-organize-steps` branch.

trushton:
@kevgo

exo-setup/src/docker-setup.ls
```diff
@@ -1,5 +1,6 @@
 require! {
   'chalk' : {red}
+  'child_process'
```
kevgo: unused module?

exo-setup/src/docker-setup.ls
```diff
@@ -44,6 +46,19 @@ class DockerSetup extends EventEmitter
       no
 
 
+  _setup-external-service: (done) ~>
+    if @config.docker-image
+      image = @config.docker-image |> (.split '/')
+      new ObservableProcess((DockerHelper.get-pull-command author: image[0], name: image[1]),
+                            stdout: {@write}
+                            stderr: {@write})
+        ..on 'ended', (exit-code) ~> done!
```
kevgo: You should check the exit code here, and display at least an error message when something went wrong. Or bubble the error up and handle it higher up in the call chain via `..on 'ended', done`.

exo-setup/src/docker-setup.ls
```diff
@@ -44,6 +46,19 @@ class DockerSetup extends EventEmitter
...
+      return
```
kevgo: unnecessary return

exo-setup/src/docker-setup.ls
```diff
@@ -44,6 +46,19 @@ class DockerSetup extends EventEmitter
...
+    else
+      @logger.log "No location or docker-image specified"
```
kevgo: Handle the error case first via a guard clause.

exosphere-shared/example-apps/app-with-external-docker-images/application.yml
```diff
@@ -0,0 +1,8 @@
+name: app-with-external-docker-images
+description: An application that relies on dockerhub images
```
kevgo: An application whose services are Dockerhub images

exosphere-shared/example-apps/app-with-external-docker-images/application.yml
```diff
@@ -0,0 +1,8 @@
...
+version: 0.0.1
```
kevgo: 0.0.0

trushton:
@kevgo Latest comments have been addressed

exo-setup/src/service-setup.ls
```diff
@@ -30,7 +30,11 @@ class ServiceSetup extends EventEmitter
 
 
   write: (text) ~>
+<<<<<<< Updated upstream
```
kevgo: Please fix this merge conflict

kevgo (approved):
Good to go once the merge conflict is resolved

PR #114, opened 2017-01-03
Remove dependency containers (exo-run)

### LINKED ISSUE #109: exo-run: remove dependencies on shutdown
Currently, we remove `exocom` and other service containers when the user exits `exo-run` with `ctrl+c`. We need to stop dependency containers as well when the user exits.

trushton:
resolves #109

@kevgo @hugobho


kevgo (approved):
Looks good to me, but @hugobho should also review this.

hugoesthere (approved):

PR #116, opened 2017-01-03
Remove command used to build webpack

trushton:
This should at least fix the issue that is causing all of the tests to fail currently. The `bin/build` script was being used to compile the webpack version of the cli. Seeing as that script no longer exists, this was causing errors.

@kevgo @hugobho


trushton:
@kevgo This change should get the tests working across other branches.

kevgo (approved):

PR #117, opened 2017-01-03
Fix linter errors

trushton:
resolves [Originate/exosphere-sdk/issues/208](https://github.com/Originate/exosphere-sdk/issues/208)

@kevgo @hugobho


kevgo (approved):
Nice cleanup, thanks dependency-lint!

hugoesthere (commented):
Awesome! Should we also update the `bin` files to run the linter?

kevgo:
Only `bin/spec` (which runs all the tests) should run the linter. `bin/tests` (which runs the unit tests) and `bin/features` (which runs the feature specs should not run the linter, since doing so slows them down unnecessarily.

PR #128, opened 2017-01-04
TR: Access service.yml correctly (exo-run)

### LINKED ISSUE #127: Access service.yml correctly in exo-run
Depending on whether services are local, or remote (pullled from Dockerhub), we want to access `service.yml` differently. If it is local, just read from the file, and if it is remote, then `cat` the file out from the docker image.

trushton:
resolves #127 

If the `@config.image` field is valid, it gets `service.yml` from the docker image, otherwise it gets the local version, seeing as these are the only two possibilities currently.

@kevgo @hugobho

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,6 @@ node_js:
   - "6"
   - "7"
 
-cache:
-  yarn: true
-  directories:
-    - node_modules
-
 git:
   depth: 2
 
--- a/exo-run/src/service-runner.ls
+++ b/exo-run/src/service-runner.ls
@@ -18,7 +18,8 @@ require! {
 class ServiceRunner extends EventEmitter
 
   ({@name, @config, @logger}) ->
-    @service-config = yaml.safe-load (DockerHelper.get-config @config.image)
+    | @config.image  =>  @service-config = yaml.safe-load DockerHelper.get-config(@config.image)
+    | otherwise      =>  @service-config = yaml.safe-load fs.read-file-sync(path.join @config.root, 'service.yml')
 
 
   start: (done) ~>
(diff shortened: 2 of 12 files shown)


.travis.yml
```diff
@@ -10,11 +10,6 @@ node_js:
   - "6"
   - "7"
 
-cache:
```
kevgo, line 13: This should be a separate PR

exo-run/src/service-runner.ls
```diff
@@ -18,7 +18,8 @@ require! {
 class ServiceRunner extends EventEmitter
 
   ({@name, @config, @logger}) ->
-    @service-config = yaml.safe-load (DockerHelper.get-config @config.image)
+    | @config.image  =>  @service-config = yaml.safe-load DockerHelper.get-config(@config.image)
```
kevgo: A class modifying a variable provided to it is an anti-pattern. Now `@config` is different for everybody else in the app, just because it was provided to a ServiceRunner instance. This means `@config` does not reflect what is in the config file, but also what ServiceRunner things the `image` attribute should be. 

Either leave this as an attribute of this class, as it was before, or extract all the logic for gathering configuration information into a `ServiceConfiguration` class.

exosphere-shared/example-apps/external-service/app/application.yml
```diff
@@ -6,4 +6,3 @@ services:
   public:
     web:
       location: ../web-service
-      docker_image: exospheredev/simple-app-web-service
```
kevgo, line 9: Aren't these separate changes?

kevgo (changes requested):

exo-run/src/service-runner.ls
kevgo: You can dry this up to:

```ls
@service-config = yaml.safe-load switch
  | @config.image  =>  DockerHelper.get-config(@config.image)
  | otherwise      =>  @fs.read-file-sync(path.join @config.root, 'service.yml')
```

Further refactoring would be to extract a method that returns the configuration for this service:

```ls
@service-config = yaml.safe-load @_service-configuration-content!

# Returns the content of the service's configuration file in textual form
_service-configuration-content: (config) ->
  | @config.image  =>  DockerHelper.get-config(@config.image)
  | otherwise      =>  @fs.read-file-sync path.join(@config.root, 'service.yml')

kevgo (approved):

PR #129, opened 2017-01-04
TR: Inject logger into exo-lint

### LINKED ISSUE #124: Inject logger into subclasses of exo commands

Check off once completed:
- [x] exo-clone
- [x] exo-lint
- [x] exo-sync
- [x] exo-test

trushton:
**Partially** resolves #124 

@kevgo @hugobho

### DIFF
--- a/exo-lint/src/app-linter.ls
+++ b/exo-lint/src/app-linter.ls
@@ -25,13 +25,13 @@ class AppLinter extends EventEmitter
       return @logger.log name: 'exo-lint', text: 'Lint passed'
 
     if not-received.length
-      @logger.log {name: 'exo lint', text: "The following messages are sent but not received:"}
+      @logger.log name: 'exo lint', text: "The following messages are sent but not received:"
       for msg in not-received
-        @logger.log {name: sent[msg], text: msg}
+        @logger.log name: sent[msg], text: msg
     if not-sent.length
-      @logger.log {name: 'exo lint', text: "The following messages are received but not sent:"}
+      @logger.log name: 'exo lint', text: "The following messages are received but not sent:"
       for msg in not-sent
-        @logger.log {name: received[msg], text: msg}
+        @logger.log name: received[msg], text: msg
 
 
   aggregate-messages: ->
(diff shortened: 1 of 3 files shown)


exo-lint/src/cli.ls
```diff
@@ -11,8 +11,5 @@ module.exports = ->
   app-config = yaml.safe-load fs.read-file-sync('application.yml', 'utf8')
   console.log "Running linter for #{cyan app-config.name}\n"
   logger = new Logger Object.keys(app-config.services)
-  app-linter = new AppLinter app-config
-    ..on 'lint-success', -> logger.log name: 'exo-lint', text: 'Lint passed'
-    ..on 'output', (data) -> logger.log data
-    ..on 'reset-colors', (service-names) -> logger.set-colors service-names
```
hugoesthere, line 17: This bit of logic `logger.set-colors service-names` still has to be replaced somewhere else right?

exo-lint/src/app-linter.ls
```diff
@@ -23,16 +22,16 @@ class AppLinter extends EventEmitter
     not-sent = difference Object.keys(received), Object.keys(sent)
 
     if not-received.length is 0 and not-sent.length is 0
-      return @emit 'lint-success'
+      return @logger.log name: 'exo-lint', text: 'Lint passed'
 
     if not-received.length
-      @emit 'output', {name: 'exo lint', text: "The following messages are sent but not received:"}
+      @logger.log {name: 'exo lint', text: "The following messages are sent but not received:"}
```
kevgo: Please remove all these terrible curly braces while you are at it here  :)

exo-lint/src/cli.ls
kevgo: +1, I only see usage of `logger.log` in the changes, not `logger.set-colors`.

trushton: The line that was removed is actually unneeded. When initializing the logger with `Object.keys (app-config.services)`, colors are assigned for each of the services. The event that was emitted to reset the colors was passing the same `Object.keys (app-config.services)`, so there was really no reason for it to be there.

trushton:
@kevgo

kevgo (approved):

PR #131, opened 2017-01-04
TR: Inject logger into exo-test

### LINKED ISSUE #124: Inject logger into subclasses of exo commands

trushton:
**Partially** resolves #124

@kevgo @hugobho

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,6 @@ node_js:
   - "6"
   - "7"
 
-cache:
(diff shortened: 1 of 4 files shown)


hugoesthere (approved):

kevgo (approved):
This is a substantial cleanup!

PR #132, opened 2017-01-05
Run setup synchronously

### LINKED ISSUE #119: Travis tests are hitting a common error
The tests for `exo-run` hit this yarn related error when running `exo-setup` on the `running` example app
``` 
users  error An unexpected error occurred: ENOTEMPTY: directory not empty, rmdir /home/travis/.cache/yarn/npm-lodash._baseindexof-3.1.0-fe52b53a1c6761e42618d654e4a25789ed61822c.
users  info If you think this is a bug, please open a bug report with the information provided in "/home/travis/build/Originate/exosphere/Originate/exosphere/exo-run/tmp/running/mongo-service/yarn-error.log".
users  info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.```

### PR DESCRIPTION
**Pull request by trushton**
resolves #119 

The errors on Travis are all coming from setting up services with yarn. After looking into documentation, running yarn installs in parallel can have problems, and should be avoided for now at least. 

@kevgo

### DIFF
--- a/exo-setup/src/app-setup.ls
+++ b/exo-setup/src/app-setup.ls
@@ -37,7 +37,7 @@ class AppSetup extends EventEmitter
       async.map-series
     else
       async.map
-    operation setups, (-> &0.start &1), (err) ~>
+    async.map-series setups, (-> &0.start &1), (err) ~>
       operation docker-setups, (-> &0.start &1), (err) ~>
         @logger.log name: 'exo-setup', text: 'setup complete'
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: exo-setup/src/app-setup.ls [hunk 1]
```diff
@@ -37,7 +37,7 @@ class AppSetup extends EventEmitter
       async.map-series
     else
       async.map
-    operation setups, (-> &0.start &1), (err) ~>
+    async.map-series setups, (-> &0.start &1), (err) ~>
```
kevgo, line 33: You can now remove lines 36-39 as well.

kevgo (commented):
good catch!

kevgo (approved):

exo-setup/src/app-setup.ls
trushton: The `docker-setups` can still be done in parallel on non-Windows systems, so should we still use `async.map` for those, or just do it all synchronously with `map-series` here?

kevgo: I see. Please calculate the `operation` variable right before you use it. I.e. move line 40 above line 33.

PR #133, opened 2017-01-05
Ignore yarn in dependency-lint

trushton:
@kevgo


exosphere-shared/example-apps/app-with-external-docker-images/local-service/service.yml
```diff
@@ -1,4 +1,4 @@
-title: simple app web server
+title: local-service
```
kevgo, line 1: This line is not part of this PR

kevgo (approved):

PR #134, opened 2017-01-05
Inject logger into exo-clone

### LINKED ISSUE #124: Inject logger into subclasses of exo commands

trushton:
resolves #124  

@kevgo

### DIFF
--- a/exo-clone/src/cli.ls
+++ b/exo-clone/src/cli.ls
@@ -13,16 +13,11 @@ clone = ->
 
   [_, _, repo-origin] = process.argv
   return missing-origin! unless repo-origin
-  repo = repo-info repo-origin
+  repository = repo-info repo-origin
 
   logger = new Logger
 
-  new AppCloner repo
-    ..on 'output', (data) -> logger.log data
-    ..on 'app-config-ready', (app-config) -> logger.set-colors Object.keys(app-config.services)
-    ..on 'app-verification-failed', (err) -> logger.log name: 'exo-clone', text: red "Error: application could not be verified.\n" + red err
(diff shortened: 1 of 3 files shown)


exo-clone/src/cli.ls
```diff
@@ -13,20 +13,11 @@ clone = ->
 
   [_, _, repo-origin] = process.argv
   return missing-origin! unless repo-origin
-  repo = repo-info repo-origin
+  repository = repo-info repo-origin
```
kevgo, line 16: You renamed this variable only in one place in this file. Do you even use it?

exo-clone/src/service-cloner.ls
```diff
@@ -35,7 +36,7 @@ class ServiceCloner extends EventEmitter
 
 
   write: (text) ~>
-    @emit 'output', {@name, text: text.trim!.replace /\.*$/, ''}
+    @logger.log {@name, text: (text.trim!.replace /\.*$/, '')}
```
kevgo: Please use parentheses for normal function calls: 

```ls
text: text.trim!.replace(/\.*$/, '')
```

and not 

```
text: (text.trim!.replace /\.*$/, '')
```

We aren't writing LISP or Scheme here  :)

kevgo (approved):

exo-clone/src/cli.ls
trushton: The `repository` referred to here is the object that is passed into the `AppCloner`

kevgo: Ah, right! I really need to take it easy today, not fit..

kevgo (approved):

PR #141, opened 2017-01-10
TR-Logger-1: Fix logger 'name' field

trushton:
After the find and replace for `@name -> @role`, the logger was printing `undefined` as the `name` for a lot of services, this fixes that.

@kevgo

### DIFF
--- a/exo-sync/features/steps/steps.ls
+++ b/exo-sync/features/steps/steps.ls
@@ -21,7 +21,7 @@ module.exports = ->
     fs.write-file-sync path.join(@app-dir, filename), file-content
 
 
-  @Given /^The origin of "([^"]*)" contains a new commit not yet present in the local clone$/, (service-role) ->
+  @Given /^The origin of "([^"]*)" contains a new commit not yet present in the local clone$/, (service-type) ->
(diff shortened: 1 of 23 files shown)


kevgo (commented):
Can you instead change the logger to use the "role" field instead of the "name' field? I think "role" is more appropriate for it.

exo-sync/features/steps/steps.ls
```diff
@@ -21,7 +21,7 @@ module.exports = ->
     fs.write-file-sync path.join(@app-dir, filename), file-content
 
 
-  @Given /^The origin of "([^"]*)" contains a new commit not yet present in the local clone$/, (service-role) ->
+  @Given /^The origin of "([^"]*)" contains a new commit not yet present in the local clone$/, (service-type) ->
```
trushton, line 24: An error had slipped past from one of the other PRs. This one instance of service-role needed to change to service-type.

trushton (commented):
@kevgo All sub-repos were tested individually and are passing. Exceptions being `exo-run` anmd the `cli` tests as they are affected by the changes made in preparation for the exocom subrepo releases.

kevgo (approved):

PR #143, opened 2017-01-11
Organize steps

trushton:
Organizes feature steps into specific `given`, `when`, and `then` files.

@kevgo


exo-run/features/support/world.ls
```diff
@@ -24,4 +29,5 @@ World = !->
 
 
 module.exports = ->
+  @app-dir = app-dir
```
kevgo: Why do you export this here? Is this variable really necessary?

kevgo (commented):
The end-to-end tests are only in the `cli` repo, not the others. So you shouldn't have to export the `app-dir` variable everywhere.

exo-run/features/support/world.ls
trushton: the `@app-dir` variable has to be used by multiple of the `given/when/then` steps, so it needs to have the same value when being called from the different files.

The comment associated with that variable in `world.ls` does seem misleading in regards to it only being used for end-to-end tests. It was copied over from the original `steps.ls` files, but I think it should be removed?

kevgo: Please make the comments meaningful. Regarding sharing the `@app-dir` variable, making it an instance variable (by prepending it with `@`) makes it available to all steps, no matter which file. Because `this` in a step is the same `world` instance. So you should be good with naming it `@app-dir` instead of `app-dir`, then it is shared everywhere automatically.

trushton:
@kevgo Thank you, it should be cleaned up now.

kevgo (approved):

PR #144, opened 2017-01-12
Fix last issues with logger

trushton:
There were a few last places that the logger needed to be updated. These were noticed now that more tests are being done with the new ExoCom changes.

@kevgo


kevgo (approved):
Please abbreviate to `@logger.log {@role, text, ...`

PR #146, opened 2017-01-17
Add comments to service.yml template

kevgo:
The generated files should contain comments to make them self-describing

@hugobho


hugoesthere (approved):
Should this also be added to the other service templates?

kevgo:
Good catch, added the comments to the other templates. Now that the tests are green, they told me exactly which ones!

PR #149, opened 2017-01-17
Use named arguments on the CLI

kevgo:
It is more descriptive to use normal parameter style (like most unix apps) on the CLI instead of positional arguments (like in Rails generators).

@hugobho


hugoesthere (approved):

trushton (approved):
Very nice! I much prefer the look of this.

PR #150, opened 2017-01-18
Compile docker run command in individual lines

hugoesthere:
Previously, when compiling a docker command string like so: 

```
command = "
  docker run
    --name=#{some-name}"
```
my editor would always remove the white space so I would have problems with `exo run` trying to execute an improperly spaced command (i.e. `docker run--name=some-name`). This PR gets rid of that issue, but it is only a temporary fix.

In the future, we will be using a docker runner class that will avoid this problem of manually compiling docker commands altogether.

@kevgo

### DIFF
--- a/exo-run/src/docker-runner.ls
+++ b/exo-run/src/docker-runner.ls
@@ -38,16 +38,13 @@ class DockerRunner extends EventEmitter
 
 
   _create-run-command: ->
-    command = "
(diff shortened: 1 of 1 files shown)


kevgo (approved):
Good fix. In the future we use https://github.com/apocas/dockerode I think

PR #151, opened 2017-01-18
Define exocom version in application.yml

hugoesthere:
partially resolves https://github.com/Originate/exosphere-sdk/issues/216

@kevgo

### DIFF
--- a/exo-create/src/entities/application.ls
+++ b/exo-create/src/entities/application.ls
@@ -25,7 +25,7 @@ application = ->
 function parse-command-line command-line-args
   data = {}
   questions = []
-  [_, _, _, app-name, app-version, ...app-description] = command-line-args
+  [_, _, _, app-name, app-version, exocom-version, ...app-description] = command-line-args
 
   if app-name
     data['app-name'] = app-name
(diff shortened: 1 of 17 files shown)


exo-create/src/entities/application.ls
```diff
@@ -54,6 +54,15 @@ function parse-command-line command-line-args
...
       default: '0.0.1'
 
+  if exocom-version
+    data['exocom-version'] = exocom-version
+  else
+    questions.push do
+      type: 'input'
+      name: 'exocom-version'
+      message: 'ExoCom version:'
+      default: '0.16.0' # eventually automate how we find latest version
```
kevgo, line 64: good idea for later!

exo-run/src/app-runner.ls
```diff
@@ -26,6 +26,7 @@ class AppRunner extends EventEmitter
         @docker-config =
           author: 'originate'
           image: 'exocom'
```
kevgo: This should be `@app-config.bus.type`

kevgo (approved):
Cool! In the future we are going to fine-tune this (auto-detect the version), but for the first few releases this is good enough!

PR #154, opened 2017-01-20
Update to use latest Exocom/ExoRelay/ExoService

hugoesthere:
End-to-end tests pass

@kevgo

### DIFF
--- a/cli/features/tutorial.feature
+++ b/cli/features/tutorial.feature
@@ -28,7 +28,7 @@ Feature: Following the tutorial
       | Name of the application to create | todo-app           |
       | Description                       | A todo application |
       | Initial version                   |                    |
-      | ExoCom version                    | latest             |
+      | ExoCom version                    | 0.16.1             |
     And waiting until the process ends
     Then my workspace contains the file "todo-app/application.yml" with content:
       """
@@ -38,7 +38,7 @@ Feature: Following the tutorial
 
       bus:
         type: exocom
-        version: latest
+        version: 0.16.1
 
       services:
         public:
@@ -61,7 +61,7 @@ Feature: Following the tutorial
 
       bus:
         type: exocom
-        version: latest
+        version: 0.16.1
 
       services:
         public:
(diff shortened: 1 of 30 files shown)


cli/features/tutorial.feature
```diff
@@ -28,7 +28,7 @@ Feature: Following the tutorial
       | Name of the application to create | todo-app           |
       | Description                       | A todo application |
       | Initial version                   |                    |
-      | ExoCom version                    | 0.16.0             |
+      | ExoCom version                    | latest             |
```
kevgo: Why `latest` here? Can we use a real version?

hugoesthere: We can, but each time we update ExoCom we'd have to change the version number in the feature tests and all of the example application templates.

It seemed hacky to me when I had to go through the whole monorepo to find and replace `0.16.0` with `0.16.1` when I had published a patch of ExoCom. Plus a new version of ExoCom could be released and our tests would be running against an older version. The developer can certainly define which version of ExoCom they want in their application.yml though.

hugoesthere: Nevermind, `latest` could mean the user pulled the `latest` from 3 weeks ago. I'll change this back to a hard version number

hugoesthere:
@kevgo good for another review

cli/features/tutorial.feature
kevgo: I don't expect us to update Exocom that much. If it becomes a problem, we can automate the search and replace. It's pretty straightforward in Vim..  :)

kevgo (approved):

kevgo:
CI is still failing though..

hugoesthere:
Yeah, sometimes the end-to-end test fails on the step executing `exo test` on an example todo app. So technically it's an issue with the tests we wrote for that example app and not the `exo` commands. I've been having trouble reproducing that error on my machine, and sometimes on Travis if I rebuild it goes away.

Regardless, Travis will still be red right now because some of the other sub-projects are failing - which I'm working on fixing.

PR #155, opened 2017-01-20
Fix failing exo-create tests

hugoesthere:
@kevgo

### DIFF
--- a/exo-create/features/steps/steps-when.ls
+++ b/exo-create/features/steps/steps-when.ls
@@ -10,7 +10,7 @@ require! {
 
 module.exports = ->
 
-  @When /^executing "([^"]*)"$/ (command, done) ->
+  @When /^executing "([^"]*)"$/, timeout: 20_000, (command, done) ->
     @process = new ObservableProcess(call-args(path.join process.cwd!, 'bin', command),
                                  cwd: @app-dir,
                                  stdout: dim-console.process.stdout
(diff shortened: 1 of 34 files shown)


exo-create/features/steps/steps-when.ls
```diff
@@ -10,7 +10,7 @@ require! {
 
 module.exports = ->
 
-  @When /^executing "([^"]*)"$/ (command, done) ->
+  @When /^executing "([^"]*)"$/, timeout: 20_000, (command, done) ->
```
hugoesthere, line 13: This step is called in `create-service.feature`

```
 When executing "exo-create service users-service test-author exoservice-es6-mongodb user manage users"
```

For whatever reason, it took about 10 seconds to create the service on my machine, so I just doubled that time. I'm not sure why it's taking so long though, because I'm pretty sure all we're doing is rendering a template.

kevgo: Hmm, strange. Let's keep an eye on that. What happens if you just run this test? Does it hang at some point?

kevgo (approved):

exo-create/features/steps/steps-when.ls
hugoesthere: Yeah, even when I just run `create-service.feature` it hangs on the same step for the same amount of time. I've `cd`ed into the tmp dir and ran the same command manually and the same behavior exists.

kevgo: Can you investigate this using debug statements (adding a bunch of `console.log` statements to see which step takes a long time)?

hugoesthere: Turns out I had `node_modules` installed in each of the service templates. I've made a PR to address this: https://github.com/Originate/exosphere/pull/157

PR #159, opened 2017-01-21
Add docker cleanup to tutorial tests

hugoesthere:
@kevgo

### DIFF
--- a/cli/features/support/env.ls
+++ b/cli/features/support/env.ls
@@ -1,4 +1,5 @@
 require! {
+  'child_process'
   '../../../exosphere-shared' : {kill-child-processes}
   'fs-extra' : fs
 }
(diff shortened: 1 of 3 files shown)


cli/features/support/env.ls
```diff
@@ -13,3 +14,12 @@ module.exports = ->
     if @app-dir
       fs.remove-sync @app-dir
     kill-child-processes done
+
+
+  #stop and remove all running docker containers
+  @After tags: ['~@docker-cleanup'], timeout: 20_000, (scenario, done) ->
+    running-containers = child_process.exec-sync 'docker ps -q' |> (.to-string!)
```
kevgo, line 21: Don't we have a DockerHelper class that encapsulates such things?

kevgo (approved):
Very cool! Does this fix the thing where I had to restart Docker manually between each test run?

cli/features/support/env.ls
hugoesthere: It doesn't have a function to remove all running containers.

hugoesthere:
Yes, you should be able to run `tutorial.feature` multiple times without restarting Docker now.

cli/features/support/env.ls
kevgo: Can you add one? It would be good if all Docker-related functionality is encapsulated in this class, rather than spread all over the codebase.

hugoesthere: That change would go along nicely with the docker-runner refactor we have in the works. I will make a ticket out of this.

PR #170, opened 2017-01-24
Fix exo-run tests

hugoesthere:
At this point all sub-projects are green (at least locally) with the exception of `website`

@kevgo

### DIFF
--- a/exo-run/features/run.feature
+++ b/exo-run/features/run.feature
@@ -24,11 +24,9 @@ Feature: running Exosphere applications
       | web   |
       | users |
     And ExoCom uses this routing:
-      | COMMAND       | SENDERS | RECEIVERS |
-      | users.list    | web     | users     |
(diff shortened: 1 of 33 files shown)


exo-run/features/run.feature
```diff
@@ -24,11 +24,9 @@ Feature: running Exosphere applications
       | web   |
       | users |
     And ExoCom uses this routing:
-      | COMMAND       | SENDERS | RECEIVERS |
-      | users.list    | web     | users     |
-      | users.listed  | users   | web       |
-      | users.create  | web     | users     |
-      | users.created | users   | web       |
+      | SERVICE | SENDS                       | RECEIVES                    |
```
kevgo, line 27: Rename `SERVICE` to `ROLE`?

kevgo (approved):

PR #172, opened 2017-01-24
Move 'remove-all-containers' to DockerHelper

### LINKED ISSUE #165: exosphere-shared docker helper: add remove all running containers function
`cli/features/support/env.ls` has a cleanup step that calls a docker command which removes all running containers. This should be made a function in `docker-helper.ls` of exosphere-shared to keep all docker related commands in one place.

trushton:
resolves #165 

Moves the function being used in `cli/features/support/env.ls` to exosphere-shared's `DockerHelper`. Additionally, modifies the function to remove **all** containers instead of just those that are currently running.

@kevgo @hugobho

### DIFF
--- a/cli/features/support/env.ls
+++ b/cli/features/support/env.ls
@@ -1,6 +1,6 @@
 require! {
   'child_process'
-  '../../../exosphere-shared' : {kill-child-processes}
+  '../../../exosphere-shared' : {kill-child-processes, DockerHelper}
   'fs-extra' : fs
 }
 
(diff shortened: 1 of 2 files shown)


kevgo (approved):
Cool! Please implement this method synchronously, i.e. remove the `done` callback.

hugoesthere (approved):
It seems like you made it synchronous but forgot to move the `done` callback in line 61

trushton:
@hugobho good catch, thanks!

PR #173, opened 2017-01-24
Don't pull 'exocom:latest' in exo-run

trushton:
@kevgo @hugobho


exo-run/bin/spec_ci
```diff
@@ -3,6 +3,5 @@ set -e
 
 node_modules/o-tools-livescript/bin/build
 node_modules/o-tools/bin/lint
-docker pull originate/exocom
```
kevgo, line 6: This is done in `exo setup` now, right?

trushton: Correct

PR #177, opened 2017-01-24
HH-3: exo-deploy: update version numbers

hugoesthere:
@kevgo @trushton

### DIFF
--- a/exo-deploy/Dockerfile
+++ b/exo-deploy/Dockerfile
@@ -1,4 +1,4 @@
-FROM hashicorp/terraform:0.8.2
+FROM hashicorp/terraform:0.8.4
 
 RUN apk add --update nodejs bash
 RUN npm install -g n
@@ -10,8 +10,10 @@ COPY ./package.json /usr/src/exo-deploy
 WORKDIR /usr/src/exo-deploy
 RUN npm install --production
 
+# Copy source code
 COPY . /usr/src/exo-deploy
 
+# Create directory for generating terraform scripts
 RUN mkdir -p /usr/src/terraform
 
 ENTRYPOINT ["./bin/start-deploy"]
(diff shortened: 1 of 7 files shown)


exo-deploy/Dockerfile
```diff
@@ -10,8 +10,10 @@ COPY ./package.json /usr/src/exo-deploy
 WORKDIR /usr/src/exo-deploy
 RUN npm install --production
 
+# Copy source code
```
kevgo, line 13: This is not a proper sentence, so don't uppercase the first character

exo-deploy/Dockerfile
```diff
@@ -10,8 +10,10 @@ COPY ./package.json /usr/src/exo-deploy
...
 COPY . /usr/src/exo-deploy
 
+# Create directory for generating terraform scripts
```
kevgo, line 16: same here

exo-deploy/package.json
```diff
@@ -4,20 +4,20 @@
...
   "dependencies": {
     "async": "2.1.4",
-    "aws-sdk": "2.7.10",
+    "aws-sdk": "2.7.27",
     "chalk": "1.1.3",
     "dashify": "0.2.2",
-    "fs-extra": "1.0.0",
-    "handlebars": "4.0.5",
-    "inquirer": "1.2.3",
+    "fs-extra": "2.0.0",
```
kevgo, line 10: This is a major version bump, are you sure it works correctly?

exo-deploy/package.json
```diff
@@ -4,20 +4,20 @@
...
+    "handlebars": "4.0.6",
+    "inquirer": "3.0.1",
```
kevgo, line 12: This is a bump of 2 major versions. I know the jump from 1 to 2 had some breaking changes. Are you sure this works still?

hugoesthere: I have been deploying from this branch and I haven't run into any problems. I also just checked the changelogs for `fs-extra` and `inquirer` and none of the breaking changes affect us.

kevgo (approved):

PR #180, opened 2017-01-25
Set logger colors correctly in exo-test

trushton:
`exo-test` had not had its logger updated since we added the `public` and `private` fields to `application.yml`, so colors were not being set for each of the services correctly.

@kevgo @hugobho


kevgo (approved):
Good catch, thanks for fixing this!

PR #183, opened 2017-01-25
(exo-add) Exit if service already exists

trushton:
resolves [Originate/exosphere#106](https://github.com/Originate/exosphere-sdk/issues/106)

@kevgo @hugobho


exo-add/src/cli.ls
```diff
@@ -30,6 +31,9 @@ module.exports = ->
     throw e
   inquirer.prompt(questions).then (answers) ->
     data := merge data, answers
+    current-services = flatten [Object.keys(app-config.services[protection-level]) for protection-level of app-config.services]
+    if current-services.includes(data.service-role)
```
kevgo: omit parentheses

kevgo (dismissed):

hugoesthere:
It looks like tests are red on the `/push` build.

trushton:
@hugobho Yeah, the tests were failing from trying to do `Object.keys` on `null`. The latest commit remedies that.

trushton:
@kevgo @hugobho I have added a feature test for this functionality. Please re-review.

exo-add/features/duplicate-role.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Attempting to add a duplicate service
+
+  As a LiveScript developer adding features to an Exosphere application
```
kevgo: Remove "LiveScript"?

exo-add/features/duplicate-role.feature
```diff
@@ -0,0 +1,28 @@
...
+  I want Exosphere to alert me when attempting to create a duplicate service
+  So that I don't create conflicting services in my application
+
+
+  Scenario: Adding a service-role that already exists
+    Given I am in the root directory of an empty application called "test app"
```
kevgo: Given I am in the root directory of an application containing a "users" service
When running "exo-add service users test-author exoservice-ls user testing" 
Then ....

exo-add/src/cli.ls
```diff
@@ -60,6 +62,15 @@ function help
   console.log help-message
 
 
+function check-for-service(app-config, data)
```
kevgo: Omit parentheses

kevgo: I'd also argue that the order of these parameters is not intuitive, and they should therefore be named parameters.

exo-add/src/cli.ls
```diff
@@ -60,6 +62,15 @@ function help
...
+  current-services = []
+  for protection-level of app-config.services
+    if app-config.services[protection-level]
+      current-services.push Object.keys(that)
+  if flatten current-services |> (.includes data.service-role )
+    console.log red "Service '#{data.service-role}' already exists in this application"
+    process.exit!
```
kevgo: Exit with status code 1 here, since this is an error

exo-add/features/duplicate-role.feature
```diff
@@ -0,0 +1,28 @@
...
+      bus:
+        type: exocom
+        version: 0.16.1
+
+      services:
+        public:
+          users:
+            location: ./users
+      """
+    When running "exo-add service users test-author exoservice-ls user testing" in this application's directory
```
kevgo: When trying to run "...

exo-add/features/duplicate-role.feature
```diff
@@ -0,0 +1,28 @@
...
+    Then it prints "Service 'users' already exists in this application" in the terminal
```
kevgo: Then it exits with code 1
And I see the error "....

exo-add/src/cli.ls
kevgo: Obey the law of demeter, and provide only the data that is needed by this method, not larger object that contain that data. And use more specific prelude methods to solve this more elegantly:

```ls
function check-for-existing-service {service-name, existing-services}
  existing-services |> (.values)
                    |> (.concat)
                    |> (.includes role)

**Comment by trushton** [comment]
@kevgo All comments addressed.

File: exo-add/features/steps/steps-given.ls [hunk 8]
```diff
@@ -17,3 +18,14 @@ module.exports = ->
     @app-dir := path.join process.cwd!, 'tmp'
     fs.empty-dir-sync @app-dir
 
+
+  @Given /^I am in the directory of an application containing a "([^"]*)" service$/, (service-role, done) !->
+    @app-dir := path.join process.cwd!, 'tmp', 'app'
+    @create-empty-app('app', done)
+      .then ~>
```
kevgo: You call `done` here and on line 31. It should be called only once. 
Also, please don't add a new paradigm like promise APIs to our code base. 
We use continuation-style callbacks, and might move to generators in the future.
Promises just add complexity and boilerplate without a clear added benefit.

exo-add/features/steps/steps-then.ls
```diff
@@ -28,8 +28,16 @@ module.exports = ->
...
-     expect(@process.full-output!).to.contain expected-text
+    expect(@process.full-output!).to.contain expected-text
 
   @When /^waiting until I see "([^"]*)" in the terminal$/, timeout: 300_000, (expected-text, done) ->
     @process.wait expected-text, done
 
+
+  @Then /^it exits with code (\d+)$/ (+expected-exit-code) ->
+    @process.on 'ended', ~>
+      expect(@process.exit-code).to.equal expected-exit-code
```
kevgo: You can simplify this to:

```ls
expect(@process.exit-code).to.equal expected-exit-code
```

The previous step waits until the process has finished.

exo-add/features/steps/steps-when.ls
```diff
@@ -42,6 +42,17 @@ module.exports = ->
       ..on 'ended', done
 
 
+  @When /^trying to run "([^"]*)" in this application's directory$/, timeout: 600_000, (command) ->
```
kevgo: Let's standardize the nomenclature: 
- _starting_ a process means kicking it off and continuing the test while the process is running
- _running_ a process means starting the process and waiting until it has finished. The process should finish without errors, meaning if it finishes with an error, this step fails.
- _trying to run_ a task means running a process, waiting until it is finished, and verifying that it has failed. If the process finished without errors, this step fails.

So, with that in mind, this step must wait for the process to finish, and then verify that `@process.exit-code` is > 0.

exo-add/src/cli.ls
```diff
@@ -30,6 +31,11 @@ module.exports = ->
     throw e
   inquirer.prompt(questions).then (answers) ->
     data := merge data, answers
+    existing-services = []
+    for protection-level of app-config.services
+      if app-config.services[protection-level]
+        existing-services.push Object.keys that
+    check-for-service {existing-services, service-role: data.service-role}
```
kevgo: This logit should be extracted into a method.

```ls
check-for-service service-role: data.service-role, services: app-config.services
```

then in `check-for-services` you massage it using the code from my last comment.

exo-add/features/steps/steps-given.ls
trushton: Ah, the promise was in there matching the implementation of the `I am in the directory of an empty app` step, I've fixed it so that neither of them, use promises.

exo-add/src/cli.ls
```diff
@@ -60,6 +62,15 @@ function help
   console.log help-message
 
 
+function check-for-service {service-role, services}
+  existing-services = []
+  for protection-level of services
+    if services[protection-level]
+      existing-services.push Object.keys that
+  if flatten existing-services |> (.includes service-role)
```
trushton: @kevgo Are you opposed to using `flatten` here? I tried using the 
```
existing-services |> (.values!)
                  |> (.concat!) 
                  |> (.includes service-role)
```
that you had mentioned, but that doesn't seem to work as intended.

kevgo: I'm not opposed to using flatten here, but don't like that there is so much low-level logic in this high-level method. Ideally this method calls a helper with the data that the helper needs, and the helper massages it.

exo-add/features/steps/steps-given.ls
```diff
@@ -10,10 +11,20 @@ module.exports = ->
   #       This is because the node_modules folder in there can contain a lot of files.
   @Given /^I am in the root directory of an empty application called "([^"]*)"$/, timeout: 20_000, (app-name, done) !->
     @app-dir := path.join process.cwd!, 'tmp', app-name
-    @create-empty-app(app-name, done).then -> done!
+    @create-empty-app(app-name, done)
```
kevgo: omit parentheses

exo-add/features/steps/steps-when.ls
```diff
@@ -42,6 +42,19 @@ module.exports = ->
...
+    args = command.split ' '
+    args[0] = path.join process.cwd!, 'bin', args[0]
+    if process.platform is 'win32'
+      args[0] += '.cmd'
+    @process = new ObservableProcess(args,
+                                     cwd: @app-dir,
+                                     stdout: dim-console.process.stdout
+                                     stderr: dim-console.process.stderr)
+      ..on 'ended', ~>
+        | @process.exit-code > 0  =>  done!
```
kevgo, line 55: What happens if the exit code is 0? It should end with an error, not time out after 10 sec.

exo-add/src/cli.ls
kevgo: Maybe add `flatten` in there:

```ls
existing-services |> (.values!)
                  |> (.concat!) 
                  |> (.flatten!)
                  |> (.includes service-role)

**Comment by trushton** [comment]
@kevgo This should be a bit more cleaned up

File: exo-add/features/steps/steps-when.ls [hunk 15]
```diff
@@ -42,6 +42,20 @@ module.exports = ->
...
+        | otherwise               =>  throw new Error "Undesirably exited with code 0"
```
kevgo, line 56: Let's make the error message: "Expected failure but exited with code 0"

exo-add/src/cli.ls
```diff
@@ -60,6 +62,19 @@ function help
   console.log help-message
 
 
+function check-for-service {service-role, existing-services}
+  if existing-services.includes service-role
+    console.log red "Service '#{service-role}' already exists in this application"
```
kevgo, line 67: Please match the existing formatting: remove the quotes and make the service role cyan

kevgo (approved):

PR #184, opened 2017-01-25
Replace morula with install script

hugoesthere:
@kevgo @trushton


yarn-install.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+cd ./cli && yarn &&
```
trushton: The trailing `&&` at the end of each line should be able to be removed.

trushton (approved):

package.json
```diff
@@ -35,6 +36,6 @@
     "url": "git+https://github.com/Originate/exosphere.git"
   },
   "scripts": {
-    "postinstall": "morula all 'yarn install'"
+    "postinstall": "./yarn-install.sh"
```
kevgo: This won't work on Windows. Can you write a LiveScript file that lives in `bin` and does this?

bin/yarn-install
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env node
+require('shelljs/global')
+
+exec('cd cli && yarn && cd ..')
```
hugoesthere: I'm not sure if we should use `yarn` or `npm` here. Using `yarn` would require the user to download another dependency, and there could be issues because we're installing `exosphere-sdk` via `npm` but the subproject dependencies via `yarn`. Although I've installed `exosphere-sdk` on my machine when the `post-install` script was using `yarn`, and at least the commands `exo-setup`, `exo-run`, and `exo-deploy` work.

On the other hand, I tried installing `exosphere-sdk` locally when all these commands were using `npm` instead of `yarn`, and it was extremely slow... 🐢

morula.yml
```diff
@@ -1,5 +1,7 @@
 before-all: 'exosphere-shared'
 after-all: ""
 always: ""
-never: "documentation"
+never:
+  - documentation
+  - bin
```
hugoesthere, line 6: This will not work until I merge the PR that uses the latest version of Morula on Travis: #192

hugoesthere:
@kevgo good for another look.

bin/yarn-install
kevgo: Let's nudge the user into the right direction, and try to use yarn as much as possible everywhere. Ideally, all of Exosphere only uses yarn.

bin/yarn-install
```diff
@@ -0,0 +1,14 @@
...
+exec('cd exo-add && yarn && cd ..')
+exec('cd exo-clone && yarn && cd ..')
```
kevgo: This `&&` stuff might not work on Windows. Please run each command individually. Also, ObservableProcess allows you to set the working directory in which you want to run the command, so you don't need all this `cd` madness here.

hugoesthere:
@kevgo using observable process.

bin/yarn-install
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env node
+ObservableProcess = require('observable-process')
+
+SUBPROJECTS_TO_PUBLISH = [
```
kevgo: These are subprojects to set up, not to publish.

bin/yarn-install
```diff
@@ -0,0 +1,25 @@
...
+  'exo-deploy',
+  'exo-lint',
+  'exo-run',
+  'exo-setup',
+  'exo-sync',
+  'exo-test',
+  'exosphere-shared'
+]
+
+for(let project of SUBPROJECTS_TO_PUBLISH)
```
kevgo: Good idea to use ES6 here instead of LS

bin/yarn-install
```diff
@@ -0,0 +1,25 @@
...
+  new ObservableProcess('yarn', { cwd: `./${project}`}).on('ended', function(code){
```
kevgo: Add space before `{`

package.json
```diff
@@ -25,7 +27,8 @@
     "exo-sync/dist",
     "exo-test/dist",
     "exosphere-shared/dist",
-    "exosphere-shared/templates"
+    "exosphere-shared/templates",
+    "bin/yarn-install"
```
kevgo: sort this alphabetically

kevgo (approved):

PR #185, opened 2017-01-25
Only push local services

trushton:
resolves [#228](https://github.com/Originate/exosphere-sdk/issues/228)

When compiling the list of images to push to dockerhub, we now check that the service has the `location` field in `application.yml`, indicating that it is a local service that can be pushed.

@kevgo @hugobho

### DIFF
--- a/exo-deploy/src/docker-hub.ls
+++ b/exo-deploy/src/docker-hub.ls
@@ -39,11 +39,11 @@ class DockerHub
     names = []
     for service-type of @app-config.services
       for name, config of @app-config.services[service-type]
-        service-config = yaml.safe-load fs.read-file-sync(path.join(process.cwd!, config.location, 'service.yml'), 'utf8')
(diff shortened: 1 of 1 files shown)


exo-deploy/src/docker-hub.ls
```diff
@@ -39,11 +39,11 @@ class DockerHub
     names = []
     for service-type of @app-config.services
       for name, config of @app-config.services[service-type]
-        service-config = yaml.safe-load fs.read-file-sync(path.join(process.cwd!, config.location, 'service.yml'), 'utf8')
-        names.push do
-          author: service-config.author
-          name: dashify service-config.type
-          #TODO: get image name if location is docker on dockerhub
+        if config.location
```
kevgo:
```ls
| !config.location  =>  continue

kevgo (approved):

PR #190, opened 2017-01-27
exo-deploy: remove hard-coded exocom version

hugoesthere:
@kevgo @trushton


kevgo (approved):
Nice!

hugoesthere:
Yarn is down, which is causing the tests to be red. Shipping anyways because there are no tests written for `exo-deploy` as it is.

PR #194, opened 2017-01-30
exo-create: Use pkg.version as default exocom version

trushton:
resolves [Originate/exosphere-sdk#232](https://github.com/Originate/exosphere-sdk/issues/232)

@kevgo @hugobho


kevgo (approved):
Beautiful!

PR #197, opened 2017-01-30
exo-create: Allow for creation of reusable services

trushton:
**Partially** resolves [Originate/exosphere-sdk#223](https://github.com/Originate/exosphere-sdk/issues/223)

Updates exo-create to prompt for a `service-type` so that generic/reusable services can be created.

@kevgo @hugobho

### DIFF
--- a/exo-create/features/create-service.feature
+++ b/exo-create/features/create-service.feature
@@ -10,7 +10,7 @@ Feature: create a reusable service
 
   Scenario: create reusable service
     Given I am in the root directory of an empty application called "empty app"
-    When executing "exo-create service users-service test-author exoservice-es6-mongodb user manage users"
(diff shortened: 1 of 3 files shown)


hugoesthere:
Hey @kevgo, now that I see this PR, I remember discussing not having reusable services in the sibling directory of the application. Instead, weren't we going to mandate that all reusable services be external ones on DockerHub, or something like that? That's why we moved `users` and `tweets` into the application directory in Space Tweet.

Does that mean `exo-create services` should be deprecated?

kevgo (approved):

kevgo:
Correction: please don't ship this yet. This no longer works in our new architecture. Let's postpone this feature until we have architected this better. Let's talk at our next standup about this.

PR #200, opened 2017-01-30
Remove version number from sub-directory package.json

hugoesthere:
We are using lock-step versioning based off the root package.json

@kevgo @trushton

### DIFF
--- a/exo-deploy/src/docker-runner.ls
+++ b/exo-deploy/src/docker-runner.ls
@@ -1,8 +1,6 @@
 require! {
   './docker-hub' : DockerHub
   '../../exosphere-shared' : {DockerHelper}
-  'fs'
-  'js-yaml' : yaml
   'observable-process' : ObservableProcess
   'path'
 }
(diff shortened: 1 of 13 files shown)


exo-deploy/src/docker-runner.ls
```diff
@@ -13,8 +13,7 @@ class Docker
   (@app-config, @logger) ->
     process.env.AWS_ACCESS_KEY_ID ? throw new Error "AWS_ACCESS_KEY_ID not provided"
     process.env.AWS_SECRET_ACCESS_KEY ? throw new Error "AWS_SECRET_ACCESS_KEY not provided"
-    @version = (yaml.safe-load fs.read-file-sync(path.join(__dirname, '../package.json'), 'utf8')) |> (.version)
-
+    @version = (yaml.safe-load fs.read-file-sync(path.join(__dirname, '../../package.json'), 'utf8')) |> (.version)
```
kevgo: I think you can just `require` package.json files, it returns them as a hash.

kevgo (approved):

kevgo:
Yes @hugobho!!

PR #205, opened 2017-02-02
publish script: push exocom updates

hugoesthere:
@kevgo @trushton 

A bug came to me in the middle of the night..
The updated Exocom dependencies should be pushed after tests have run successfully, so that in the event they fail we can discard the changes that were made to update dependencies.

### DIFF
--- a/bin/publish
+++ b/bin/publish
@@ -41,6 +41,7 @@ function main
   update-exocom-dependencies!
   build-subprojects!
   run-tests!
+  push-exocom-updates!
   bump-version-number!
   push-version-number!
   publish-to-npm!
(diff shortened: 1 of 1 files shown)


kevgo:
Shouldn't we just abort the script when the tests fail, and not push anything?

hugoesthere:
It does. First it updates the ExoCom dependencies, then runs the tests with the new ExoCom dependencies. Only after those tests pass does it push the updated ExoCom dependencies. That's what this PR is addressing. If those tests fail then the publish script exits before the changes can be pushed.

kevgo (approved):
Okay cool!

PR #206, opened 2017-02-02
Persistent databases in dev  mode (exo-run)

trushton:
resolves [#239](https://github.com/Originate/exosphere-sdk/issues/239)

Overhauls the process of checking on and running dependency docker containers, now using ObservableProcess. This change is only in effect for exo-run and a separate PR will be submitted relaying those changes into exo-test. 

This has been tested to be working with the SpaceTweet application. 
Dependencies in `application.yml` of the application were updated to include a `volume` and `online-text` field.

@kevgo @hugobho


exo-run/src/docker-runner.ls
```diff
@@ -63,11 +65,20 @@ class DockerRunner extends EventEmitter
...
-    for dependency of @docker-config.dependencies
-      app-dependency = "#{@docker-config.app-name}-#{dependency}"
-      DockerHelper.ensure-container-is-running app-dependency, dependency
-      @docker-config.env[dependency.to-upper-case!] = DockerHelper.get-docker-ip app-dependency
+  _check-dependency-containers: (done) ~>
+    dependencies = []
+    for image, vars of @docker-config.dependencies
+      container-name = "#{@docker-config.app-name}-#{image}"
+      if vars.docker_flags?
+        volume = Handlebars.compile(that.volume)({"EXO_DATA_PATH": "/Users/originate/Desktop/data"})
```
hugoesthere: This should be whatever `~/.exosphere` path we decide to go with

trushton: Correct, the one I used here was just a placeholder for testing it out locally. This prevents tests from passing on CI.

EDIT: From stand-up on 2/2/2017, the path should be `~/.exosphere/{{application}}/{{dependency}}/data`

exosphere-shared/src/docker-helper.ls
```diff
@@ -43,14 +44,19 @@ class DockerHelper
...
-    if container is \test-mongo
-      child_process.exec-sync "docker run -d --name=#{container} -p 27017:27017 #{image}"
+  @run-image = (container, done) ~>
+    if container.container-name is \test-mongo
+      child_process.exec-sync "docker run -d --name=#{container.container-name} -p 27017:27017 #{container.image}"
     else
-      child_process.exec-sync "docker run -d --name=#{container} #{image}"
+      new ObservableProcess("docker run #{if container.volume? then that} --name=#{container.container-name} #{container.image}",
+                            stdout: {write: (text) -> },
+                            stderr: {write: (text) -> })
```
hugoesthere: We should still figure out a way to handle this. We could be using `run-image` to start a container for which we want to see the output.

exo-run/src/docker-runner.ls
```diff
@@ -63,11 +65,20 @@ class DockerRunner extends EventEmitter
...
 
 
-  _check-dependency-containers: ~>
-    for dependency of @docker-config.dependencies
-      app-dependency = "#{@docker-config.app-name}-#{dependency}"
-      DockerHelper.ensure-container-is-running app-dependency, dependency
-      @docker-config.env[dependency.to-upper-case!] = DockerHelper.get-docker-ip app-dependency
+  _check-dependency-containers: (done) ~>
+    dependencies = []
+    for image, vars of @docker-config.dependencies
```
kevgo: Let's use correct variable names here: `dependency-name` and `dependency-config` or something similar.

exo-run/src/docker-runner.ls
kevgo: Yes, let's use [os.homedir](https://nodejs.org/api/os.html#os_os_homedir) here instead of a hard-coded path.

exo-run/src/docker-runner.ls
```diff
@@ -63,11 +65,20 @@ class DockerRunner extends EventEmitter
...
+      container-name = "#{@docker-config.app-name}-#{image}"
+      if vars.docker_flags?
+        volume = Handlebars.compile(that.volume)({"EXO_DATA_PATH": "/Users/originate/Desktop/data"})
+        online-text = that.online_text
+      dependencies.push {container-name, image, volume, online-text}
+
+    async.map-series dependencies, DockerHelper.ensure-container-is-running, (err) ~>
+      | err  =>  @emit 'error', err
```
kevgo: Shouldn't we also abort here if there is an error? There is no point in trying to continue if there is an error. The app should stop right here.

exosphere-shared/src/docker-helper.ls
```diff
@@ -13,10 +14,10 @@ class DockerHelper
     child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}') |> (.to-string!) |> (.split '\n') |> (.includes "#{container-name}/Up")
 
 
-  @ensure-container-is-running = (container-name, image) ->
-    | @container-is-running container-name  =>  return
-    | @container-exists container-name      =>  @start-container container-name
-    | otherwise                             =>  @run-image container-name, image
+  @ensure-container-is-running = (container, done) ~>
+    | @container-is-running container.container-name  =>  return done!
+    | @container-exists container.container-name      =>  @start-container container.container-name; done!
```
kevgo: I think `@start-container` is an async function, and should receive `done` as a callback parameter here.

exosphere-shared/src/docker-helper.ls
```diff
@@ -43,14 +44,19 @@ class DockerHelper
     child_process.exec-sync "docker rm -f #{container}" if @container-exists container
 
 
-  @run-image = (container, image) ->
-    if container is \test-mongo
-      child_process.exec-sync "docker run -d --name=#{container} -p 27017:27017 #{image}"
+  @run-image = (container, done) ~>
+    if container.container-name is \test-mongo
```
kevgo: Why do we have special handling for `test-mongo` here? Can't we start it using `ObservableProcess` below?

exosphere-shared/src/docker-helper.ls
```diff
@@ -43,14 +44,19 @@ class DockerHelper
...
+      child_process.exec-sync "docker run -d --name=#{container.container-name} -p 27017:27017 #{container.image}"
     else
-      child_process.exec-sync "docker run -d --name=#{container} #{image}"
+      new ObservableProcess("docker run #{if container.volume? then that} --name=#{container.container-name} #{container.image}",
```
kevgo: You should be able to simplify the expression in the first string interpolation expression to 
```
#{container.volume}
```

exosphere-shared/src/docker-helper.ls
kevgo: See the [ObservableProcess documentation](https://github.com/Originate/observable-process/blob/master/features/stdout.feature#L9) on how to disable logging.

exosphere-shared/src/docker-helper.ls
```diff
@@ -43,14 +44,19 @@ class DockerHelper
...
+                            stdout: {write: (text) -> },
+                            stderr: {write: (text) -> })
+        ..on 'ended', (exit-code) ->
+          | exit-code > 0  =>  throw new Error "Dependency #{container.container-name} failed to run"
```
kevgo: We should not throw exceptions. Call the `done` method with the error here, and return from the method. The caller must handle the error and exit.

exo-run/src/docker-runner.ls
trushton: This `error` event being emitted triggers the shutdown process for the entire app, so this should be good.

kevgo: Okay, good for now. We have to re-think the overall architecture of this, but that can be done later.

trushton:
@kevgo As #211 is an extension of this PR, is it safe to assume that this PR is also good to go?

kevgo:
Tests are red  :(

trushton:
@kevgo I merged the other PR into here, all green now :)

kevgo (approved):
Looks good, thanks for finishing this up! This is a very important feature.

PR #207, opened 2017-02-02
exosphere-shared: mongo template bug fix

hugoesthere:
Attempting to resolve: https://github.com/Originate/exosphere-sdk/issues/230

My suspicion is that `db.close` isn't finishing before the next steps are run (and it closes the database connection in the next step rather than in the previous one).

@kevgo @trushton


trushton (approved):

kevgo (approved):
I think I ran into a similar issue. Let's see if this helps!

PR #210, opened 2017-02-04
Ensure in root directory before publish

hugoesthere:
Ensure user is in root directory before publishing, otherwise all the commands will be run in the wrong subdirectories.

@kevgo

### DIFF
--- a/bin/package.json
+++ b/bin/package.json
@@ -7,6 +7,7 @@
     "inquirer": "3.0.1",
     "livescript": "1.5.0",
     "o-tools-livescript": "1.2.3",
+    "path": "0.12.7",
     "replace-in-file": "2.2.2",
     "semver": "5.3.0",
     "shelljs": "0.7.6"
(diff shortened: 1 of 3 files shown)


bin/package.json
```diff
@@ -7,6 +7,7 @@
     "inquirer": "3.0.1",
     "livescript": "1.5.0",
     "o-tools-livescript": "1.2.3",
+    "path": "0.12.7",
```
kevgo: You should be able to use the built-in [path](https://nodejs.org/api/path.html#path_path_basename_path_ext) module in Node here.

bin/publish
```diff
@@ -193,6 +195,10 @@ function ensure-on-master
   if current-branch.trim! isnt 'master' then console.log red 'You must be on the master branch to publish' ; process.exit 1
 
 
+function ensure-in-root-dir
+  if path.basename(process.cwd!) isnt 'exosphere' then console.log red 'You must be in the root directory to publish' ; process.exit 1
```
kevgo: Remove `then` and move the code after it to the next line, indented

hugoesthere:
@kevgo comments addressed

kevgo (approved):
Alright!

PR #211, opened 2017-02-06
Update exo-test dependency containers

trushton:
Extension of #206 

Addresses one of the comments left in the mentioned PR about running dependency containers during the `exo-test` command.

@kevgo


exo-test/src/app-tester.ls
```diff
@@ -13,12 +13,19 @@ class AppTester extends EventEmitter
...
   start-testing: ->
-    testers = []
+    @services = []
     for protection-level of @app-config.services
       for service-role, service-data of @app-config.services[protection-level]
-        service-dir = path.join process.cwd!, service-data.location
-        testers.push (new ServiceTester {role: service-role, config: {root: service-dir}, @logger})
-    async.series [tester.start for tester in testers], (err, exit-codes) ~>
+        @services.push do
+          {
```
kevgo: You don't need `do` and curly braces at the same time. Either one works here. I'd remove the curly braces.

exo-test/src/service-tester.ls
```diff
@@ -42,9 +43,18 @@ class ServiceTester extends EventEmitter
 
 
   _start-dependencies: (done) ~>
-    for dep of @service-config.dependencies
-      DockerHelper.ensure-container-is-running "test-#{dep}", dep
-    wait 500, done
+    dependencies = []
+    for dependency-name, dependency-config of @service-config.dependencies
+      dependencies.push do
+        {
```
kevgo: Remove curly braces

exo-test/src/service-tester.ls
```diff
@@ -42,9 +43,18 @@ class ServiceTester extends EventEmitter
...
+          container-name: "test-#{dependency-name}"
+          port: '-p 27017:27017'
+          dependency-name: dependency-name
+          online-text: dependency-config.docker_flags?.online_text
+        }
+    async.each-series dependencies, DockerHelper.ensure-container-is-running, (err) ~>
+      | err  => throw new Error err
```
kevgo: Are you sure throwing an error is the right course of action here? We should return it via the `done` function.

exosphere-shared/src/docker-helper.ls
```diff
@@ -16,7 +16,7 @@ class DockerHelper
 
   @ensure-container-is-running = (container, done) ~>
     | @container-is-running container.container-name  =>  return done!
-    | @container-exists container.container-name      =>  @start-container container, done
+    | @container-exists container.container-name      =>  @start-container container, (err) ~> done err
```
kevgo: What is the difference between the old and new code here?

exosphere-shared/src/docker-helper.ls
```diff
@@ -45,24 +45,22 @@ class DockerHelper
 
 
   @run-image = (container, done) ~>
-    if container.container-name is \test-mongo
-      child_process.exec-sync "docker run -d --name=#{container.container-name} -p 27017:27017 #{container.image}"
-    else
-      new ObservableProcess("docker run #{container.volume} --name=#{container.container-name} #{container.dependency-name}",
+      #TODO: Clean up the flags to the docker run command
+      new ObservableProcess("docker run #{if container.volume then that else ''} #{if container.port then that else ''} --name=#{container.container-name} #{container.dependency-name}",
```
kevgo: You can say `container.volume or ''` here, and similarly `container.port or ''`

trushton:
@kevgo Comments have been addressed. Strangely enough the tests are passing locally, and on CI they are failing with this mismatch error that doesn't actually seem to be a mismatch:
screen shot 2017-02-08 at 8 24 27 am

kevgo:
The changes look good to me, but please don't merge this while the tests are broken. Maybe there is a white-space issue, like some trailing spaces after `mongo:`?

trushton:
@kevgo My mistake, it seems that the error wasn't coming from where I thought it was. An update has been made to fix the tests and they should all be green now.

kevgo (approved):
Nice!

PR #212, opened 2017-02-08
Fix global install

kevgo:
The post-install script was running all `yarn` installs in parallel. This was causing random filesystem errors, since Yarn uses a shared cache and isn't thread-safe.

@hugobho @trushton

### DIFF
--- a/bin/yarn-install
+++ b/bin/yarn-install
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
-ObservableProcess = require('observable-process')
+
+const ObservableProcess = require('observable-process'),
+      async = require('async')
 
 SUBPROJECTS_TO_SETUP = [
   'cli',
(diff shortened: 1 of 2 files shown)


bin/yarn-install
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
-ObservableProcess = require('observable-process')
+
+const ObservableProcess = require('observable-process'),
```
kevgo, line 3: In ES6 you have to provide `const` or `var`, otherwise the variable exists in the global scope.

package.json
```diff
@@ -6,11 +6,13 @@
     "exo": "cli/bin/exo"
   },
   "dependencies": {
+    "async": "2.1.4",
     "observable-process": "3.2.2"
   },
   "description": "SDK for the Exosphere framework",
   "devDependencies": {},
   "engines": {
+    "node": ">= 6.0.0",
```
kevgo, line 15: Since our install-script uses ES6, we require Node 6+.

kevgo:
Testing of this btw via `npm i -g .`

PR #213, opened 2017-02-08
Make publish work

kevgo:
The publish script uses NPM modules that aren't downloaded to the publishing machine.

@hugobho @trushton


hugoesthere (approved):
Should these dependencies also be removed from `bin/package.json` then?

PR #217, opened 2017-03-14
Fix broken inline HTML

letsbreelhere:
cf. [this](http://stackoverflow.com/a/29378250) SO answer – apparently the fix is just adding newlines around the text content of a tag.


kevgo:
Thanks @letsbreelhere! Sorry for the delay! I am now watching this repo so that I see PRs when they are created!

PR #218, opened 2017-03-28
Handle version command first

trushton:
resolves [exosphere-sdk#248](https://github.com/Originate/exosphere-sdk/issues/248)

The `require` statement that pulls in the exo commands seems to be causing some lag. In the case that we run `exo version`, we want that to return quickly, so the code for handling that command was lifted to the top of the file.

@kevgo

### DIFF
--- a/cli/src/cli.ls
+++ b/cli/src/cli.ls
@@ -1,27 +1,11 @@
 require! {
-  '../../package.json' : pkg
-}
-
-command-name = process.argv[2]
-if command-name is \version # Version handled before all others for performance
-  return console.log "Exosphere version #{pkg.version}"
-
-require! {
   'abbrev'
   'chalk' : {red}
-  '../../exo-add' : add
-  '../../exo-clone' : clone
-  '../../exo-create' : create
-  '../../exo-deploy' : deploy
-  '../../exo-lint' : lint
-  '../../exo-run' : run
-  '../../exo-setup' : setup
-  '../../exo-sync' : sync
-  '../../exo-test' : test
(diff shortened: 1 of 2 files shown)


trushton:
It also seems that the last time the sdk was updated, the version number of `exorelay` did not receive the update. The tests are looking to install version `0.21.5` but the latest available version of `exorelay` on npm is `0.21.4`

kevgo (commented):
Thanks for looking into this! Which require exactly is the slow one? They should all be pretty fast, no?

kevgo:
If requiring all the commands is slow, maybe we should only require the one that is actually executed?

kevgo:
After thinking more about this, there is no need to statically require all commands all the time. Let's only require the command that is being executed.

trushton:
@kevgo This works much better now

cli/src/cli.ls
```diff
@@ -20,24 +11,23 @@ require! {
...
+  run: "../../exo-run"
+  setup: "../../exo-setup"
+  sync: "../../exo-sync"
+  test: "../../exo-test"
 
 command-name = process.argv[2]
 full-command-name = complete-command-name command-name
 if command-name is \version
-  console.log "Exosphere version #{pkg.version}"
+  return console.log "Exosphere version #{pkg.version}"
```
kevgo, line 30: If you add a `return` here, you can remove the `else` on the next line.

kevgo (approved):
Nice! The tests are red now. Feel free to ship when they are green!

PR #229, opened 2017-05-12
Pass port mapping to external dependency containers

hugoesthere:
resolves https://github.com/Originate/exosphere-sdk/issues/252

In exosphere-shared/docker-helper.ls:

Also adds a check in `@run-image` to gracefully retry `@ensure-container-is-running` if the process fails with a message indicating that the container is already in use. This is to address an edge case that happens when multiple services use the same dependency and Exosphere tries to boot them up concurrently:
 - `@ensure-container-is-running` first checks if the dependency for `service1` is running. It's not, so it goes to start the container from an image.
- `@ensure-container-is-running` checks if the same dependency for `service2` is running. It's not because Docker is still starting the container from the previous step. It goes to start the same dependency from an image.
- `@run-image` tries to start the dependency for `service2`, but by now the same dependency for `service1` has already been booted up. It errors with a message saying that the container is already in use, which it is.

The logic I put in will re-call `@ensure-container-is-running` with the same container, and this time the method will properly exit after verifying that the dependency has been booted up.

This isn't the cleanest way to handle this problem, but it's the quickest solution for now. We could implement some sort of booting queue, but that would require a larger refactor (which I'm not against)

### DIFF
--- a/exosphere-shared/example-apps/external-dependency/mongo/server.ls
+++ b/exosphere-shared/example-apps/external-dependency/mongo/server.ls
@@ -11,6 +11,4 @@ module.exports =
       done!
 
 function get-mongo-address
-  if process.env.MONGO
-    return "mongodb://#{process.env.MONGO}"
-  return "mongodb://localhost:27017"
+  return "mongodb://#{process.env.MONGO or 'localhost:27017'}"
--- a/exosphere-shared/example-apps/external-dependency/mongo/server.ls
+++ b/exosphere-shared/example-apps/external-dependency/mongo/server.ls
@@ -1,0 +1,16 @@
+require! {
+  'mongodb' : {MongoClient}
+  'nitroglycerin' : N
+}
+
+module.exports =
+
+  before-all: (done) ->
+    MongoClient.connect get-mongo-address!, N (mongo-db) ->
+      console.log("MongoDB connected")
+      done!
+
+function get-mongo-address
+  if process.env.MONGO
+    return "mongodb://#{process.env.MONGO}"
+  return "mongodb://localhost:27017"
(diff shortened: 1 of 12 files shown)


exosphere-shared/example-apps/external-dependency/mongo/server.ls
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports =
+
+  before-all: (done) ->
+    MongoClient.connect get-mongo-address!, N (mongo-db) ->
+      console.log("MongoDB connected")
+      done!
+
+function get-mongo-address
+  if process.env.MONGO
```
kevgo:
```ls
return "mongodb://#{process.env.MONGO or 'localhost:27017'}"

File: exosphere-shared/src/docker-helper.ls [hunk 2]
```diff
@@ -45,12 +45,15 @@ class DockerHelper
...
-                          stderr: false)
-      ..on 'ended', (exit-code, killed) ->
-        | exit-code > 0 and not killed  =>  return done "Dependency #{container.container-name} failed to run, shutting down"
-      ..wait container.online-text, done
+    process = new ObservableProcess("docker run #{container.volume or ''} #{container.port or ''} --name=#{container.container-name} #{container.dependency-name}"
+                                    stdout: false, 
+                                    stderr: false)
+      ..on 'ended', (exit-code, killed) ~>
+        | exit-code > 0 and not killed  =>  
+          if /container name ".*" is already in use by container/.test process.full-output!
```
kevgo: Please add a comment explaining what you do here, for example:

```ls
# if the image has already been started by another service, use the existing instance
if /container..../

kevgo (approved):
Great investigation and bug fix, @hugobho! I think we need to evolve the `DockerHelper` into a `DockerManager` to make this more organized:

- if somebody wants to boot up a Docker image, they tell the DockerManager to boot this image up
- it also keeps a list of all running images
- the DockerManager knows that each image should run only once
- so when it is asked a second time, it uses the running instance from the first image
- it also provides methods to shut down a Docker image, or restart it

PR #231, opened 2017-05-13
Fix the tutorial

kevgo:
Makes the tutorial run again. I am switching to longer but more unambiguous command-line parameters for exo-add. I think that's fine since they will only be used in scripts.

### DIFF
--- a/exo-add/src/cli.ls
+++ b/exo-add/src/cli.ls
@@ -82,11 +82,10 @@ function get-existing-services services
 function parse-command-line command-line-args
   data = {}
   questions = []
-  service-role = command-line-args["role"]
-  entity-name = command-line-args["_"]
-  service-type = command-line-args["type"]
+  service-role = command-line-args["service-role"]
+  service-type = command-line-args["service-type"]
   author = command-line-args["author"]
-  template-name = command-line-args['template']
+  template-name = command-line-args['template-name']
   model-name = command-line-args['model']
   description = command-line-args['description']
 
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -14,6 +14,8 @@ git:
   depth: 2
 
 before_install:
+  - sudo apt-get update
+  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
```
kevgo: updates to the latest docker

hugoesthere (approved):

exo-add/src/cli.ls
```diff
@@ -82,9 +82,8 @@ function get-existing-services services
 function parse-command-line command-line-args
   data = {}
   questions = []
-  service-role = command-line-args["role"]
-  entity-name = command-line-args["_"]
-  service-type = command-line-args["type"]
+  service-role = command-line-args["service-role"]
+  service-type = command-line-args["service-type"]
   author = command-line-args["author"]
   template-name = command-line-args['template']
```
trushton: ~~command-line-args['template']~~ 👎 

command-line-args['template-name'] 👍

trushton (changes requested):

trushton:
@kevgo I fixed the one error that was causing failures, but now its back to where things were before with the `write EPIPE` errors

trushton (approved):

exo-add/src/cli.ls
kevgo: @trushton thanks for fixing this. Do the tests run locally on your machine? On my machine they fail now.

kevgo: That's why I didn't do this change before.

kevgo: Correction: Travis worked, so the setup must have been on my machine. Upgrading Node fixed the issue. This works now. Thanks for the contribution!

PR #232, opened 2017-05-16
Implement versions for service dependencies

hugoesthere:
This PR allows services to use specific versions of dependencies, so if an organization releases a breaking change to Dockerhub the service using that dependency will not suffer.

### DIFF
--- a/exo-run/src/service-runner.ls
+++ b/exo-run/src/service-runner.ls
@@ -76,15 +76,13 @@ class ServiceRunner extends EventEmitter
     dependencies = []
     for dependency-name, dependency-config of @service-config.dependencies or {}
       container-name = "#{@config.app-name}-#{dependency-name}"
-      if dependency-config?.version?
-        dependency-name += ":#{dependency-config.version}"
       if dependency-config?.docker_flags?
         data-path = path.join os.homedir!, '.exosphere', @config.app-name, dependency-name, 'data'
         mkdir '-p', data-path
         volume = Handlebars.compile(that.volume)({"EXO_DATA_PATH": data-path})
         online-text = that.online_text
         port = that.port
-      dependencies.push {container-name, dependency-name, volume, online-text, port}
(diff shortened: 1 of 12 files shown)


kevgo (approved):
Yay!

exo-run/src/service-runner.ls
```diff
@@ -76,6 +76,8 @@ class ServiceRunner extends EventEmitter
     dependencies = []
     for dependency-name, dependency-config of @service-config.dependencies or {}
       container-name = "#{@config.app-name}-#{dependency-name}"
+      if dependency-config?.version?
+        dependency-name += ":#{dependency-config.version}"
```
trushton: I believe this could be simplified to 

`dependency-name += ":{that}"`

as `that` would be referring to `dependency-config.version` already

exo-test/src/service-tester.ls
```diff
@@ -46,6 +46,8 @@ class ServiceTester extends EventEmitter
     dependencies = []
     for dependency-name, dependency-config of @service-config.dependencies
       container-name = "test-#{dependency-name}"
+      if dependency-config?.version?
+        dependency-name += ":#{dependency-config.version}"
```
trushton: same as above

trushton (approved):

exo-run/src/service-runner.ls
kevgo: Good catch. You might want to remove the trailing `?` from line 79 to make sure you get the actual version then.

hugoesthere: Good catch on the good catch!

hugoesthere:
@kevgo @trushton 

I moved the `:version` concatenation to `docker-helper.ls` since modifying the `dependency-name` variable directly was affecting how environment variables are being set. Can you guys take another look? I'll ship for now so I can continue debugging broken tests.

PR #233, opened 2017-05-16
Fix get-docker-ip function

hugoesthere:
The command returns a buffer that needs to be formatted properly.

### DIFF
--- a/exosphere-shared/src/docker-helper.ls
+++ b/exosphere-shared/src/docker-helper.ls
@@ -29,7 +29,8 @@ class DockerHelper
 
 
   @get-docker-ip = (container) ->
-    child_process.exec-sync("docker inspect --format '{{ .NetworkSettings.IPAddress }}' #{container}", "utf8") if DockerHelper.container-exists container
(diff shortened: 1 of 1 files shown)


exosphere-shared/src/docker-helper.ls
```diff
@@ -29,7 +29,8 @@ class DockerHelper
 
 
   @get-docker-ip = (container) ->
-    child_process.exec-sync("docker inspect --format '{{ .NetworkSettings.IPAddress }}' #{container}", "utf8") if DockerHelper.container-exists container
+    if DockerHelper.container-exists container
+      child_process.exec-sync("docker inspect --format '{{ .NetworkSettings.IPAddress }}' #{container}", "utf8") |> (.to-string!) |> (.replace '\n', '')
```
kevgo, line 33: Better to call [trim](https://www.w3schools.com/jsref/jsref_trim_string.asp) on the string rather than removing line breaks manually. On windows they look differently.

kevgo (approved):

trushton (approved):

PR #234, opened 2017-05-16
Update service dependency and exocom versions

hugoesthere:
Tests pass for all subprojects locally now (except for `website/`).


kevgo (approved):
Very awesome! Please disable the website tests for now, I'll fix them later.

hugoesthere:
I think they have been disabled for a while now. Hence why tests are green right now :)

PR #235, opened 2017-05-17
Fix container-is-running method

hugoesthere:
When containers have been running for more than a few seconds, their statuses look like so:
```
exocom/Up Less than a second
todo-app-mongo/Up 9 seconds
```
Instead of just `exocom/Up`. We need to check `.includes` against each of the strings in the array rather than the array itself.

### DIFF
--- a/exosphere-shared/src/docker-helper.ls
+++ b/exosphere-shared/src/docker-helper.ls
@@ -11,7 +11,10 @@ class DockerHelper
 
 
   @container-is-running = (container-name) ->
-    child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}') |> (.to-string!) |> (.split '\n') |> (.includes "#{container-name}/Up")
+    running-containers = child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}') |> (.to-string!) |> (.split '\n') 
+    for container in running-containers
+      if container.includes "#{container-name}/Up" then return yes
+    return no
 
 
   @ensure-container-is-running = (container, done) ~>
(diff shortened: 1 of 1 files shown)


exosphere-shared/src/docker-helper.ls
```diff
@@ -11,7 +11,10 @@ class DockerHelper
 
 
   @container-is-running = (container-name) ->
-    child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}') |> (.to-string!) |> (.split '\n') |> (.includes "#{container-name}/Up")
+    running-containers = child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}') |> (.to-string!) |> (.split '\n')
```
kevgo: Let's split by https://nodejs.org/api/os.html#os_os_eol instead of hard-conding the string. It is different on Windows.

exosphere-shared/src/docker-helper.ls
```diff
@@ -11,7 +11,10 @@ class DockerHelper
 
 
   @container-is-running = (container-name) ->
-    child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}') |> (.to-string!) |> (.split '\n') |> (.includes "#{container-name}/Up")
+    running-containers = child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}') |> (.to-string!) |> (.split '\n') 
+    for container in running-containers
```
kevgo:
```ls
running-containers |> any (.includes "#{container-name}/Up")

kevgo: You can even simplify this all into one expression

kevgo (commented):
Good stuff!

hugoesthere:
@kevgo good for another look

trushton (approved):
Nice and clean 👍

kevgo (approved):
🎉

PR #236, opened 2017-05-17
Fix callback being called twice

hugoesthere:
Addresses https://github.com/Originate/exosphere-sdk/issues/264

Temporary fix for when `done` callback is being called twice. This logic will be cleaned up when we a extract a proper `DockerManager` class.


kevgo (approved):
Good!

PR #237, opened 2017-05-18
Print crash report for dependency containers

hugoesthere:
Partly addresses: https://github.com/Originate/exosphere-sdk/issues/261

While not a permanent fix, this change will at least allow the developer not to be completely in the dark when a dependency fails. When we eventually switch to a DockerManger class this might change again. Until that happens we will at least have crash reports for the next release.

### DIFF
--- a/exosphere-shared/src/docker-helper.ls
+++ b/exosphere-shared/src/docker-helper.ls
@@ -87,8 +87,9 @@ class DockerHelper
 
   # Prints the last n number of lines from a given text
   @_print-last-lines = (text, number-of-lines) ->
-    split-text = text.split /\r?\n/
-    sliced-text = split-text.slice Math.max(split-text.length - number-of-lines - 1, 0)
-    sliced-text.join (if process.platform == 'win32' then '\r\n' else '\n')
+    text
+      .split \\n
+      .[-number-of-lines to]
+      .join \\n
 
 module.exports = DockerHelper
--- a/exosphere-shared/src/docker-helper.ls
+++ b/exosphere-shared/src/docker-helper.ls
@@ -54,6 +54,7 @@ class DockerHelper
           # if the image has already been started by another service, use the existing instance
           if /container name ".*" is already in use by container/.test process.full-output!
             return @ensure-container-is-running container, done
+          console.log @_print-last-lines(process.full-output!, 20)
           return done "Dependency #{container.container-name} failed to run, shutting down"
       ..wait container.online-text, done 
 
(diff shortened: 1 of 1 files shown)


exosphere-shared/src/docker-helper.ls
```diff
@@ -84,5 +85,10 @@ class DockerHelper
       child_process.exec-sync 'docker rm -f $(docker ps -aq)'
 
 
+  # Prints the last n number of lines from a given text
+  @_print-last-lines = (text, number-of-lines) ->
+    split-text = text.split /\r?\n/
+    sliced-text = split-text.slice Math.max(split-text.length - number-of-lines - 1, 0)
+    sliced-text.join (if process.platform == 'win32' then '\r\n' else '\n')
```
alexdavid: I think this can be
```livescript
@_print-last-lines = (text, number-of-lines) ->
  text
    .split \\n
    .[-number-of-lines to]
    .join \\n

**Review by alexdavid — APPROVED** [review]
Looks good!

File: exosphere-shared/src/docker-helper.ls [hunk 2]
```diff
@@ -84,5 +85,11 @@ class DockerHelper
       child_process.exec-sync 'docker rm -f $(docker ps -aq)'
 
 
+  # Prints the last n number of lines from a given text
+  @_print-last-lines = (text, number-of-lines) ->
+    text
+      .split \\n
```
kevgo, line 91: use `os.EOL` here

hugoesthere: @alexdavid made a good point: since Windows machines split on `\r\n`, if we split on `\n` and then rejoin on `\n` we won't need to use `os.EOL`

kevgo: Such hackery leads to all sorts of unpredictable bugs, messes, and headaches. Even if it works in this particular example, the moment you change something it might no longer work, meaning because we don't do things properly here, other changes will have unintended side effects. 

Here are a few of the fun things waiting for you in the future:
- lines now have a dangling `\r` while being processed. If you print a line for debugging, it yanks your cursor to the beginning of the line in the terminal, and all output is getting messed up, overwriting text. Fun times.
- when joining things back together, the last line might end up only with a `\r`, without an `\n`. 
- if you decide to sanitize code, for example via `trim`, then it suddenly doesn't work on Windows anymore

Because of all this crazy shit, it's better to do things properly right from the start. Simple, straightforward, robust, easy to understand code that isn't full of hidden trap doors and issues that are just waiting to blow up in your face, possibly in other parts of the code base.

hugoesthere: Addressed: https://github.com/Originate/exosphere/pull/239

PR #242, opened 2017-05-24
WIP: Switch exo-setup to docker-compose

hugoesthere:
Tests won't be green yet, still need to reflect these changes in `exo-run` and `exo-test`.

I wanted to break this PR up into a couple smaller ones, but unfortunately a lot of these changes are tied to each other:

1) Switch `exo-setup` to using `docker-compose` instead of cli commands:
    - generate `docker-compose.yml` file in root dir
    - use `docker-compose build` to build all images at once. This means also copying the Dockerfile over with each of the service templates, which was being done in a different step previously.
2) Change `service.yml` files (used to build `docker-compose.yml`) to a less flag-oriented format.
3) Add step to test that `exo-setup` was exiting properly. We were getting false positives related to this.
4) In `exosphere-shared/src/compile-service-routes.ls`, only compile internal services. This check had to be added after some logic was removed in `exo-setup/app-setup.ls`


exo-setup/src/app-setup.ls
```diff
@@ -1,17 +1,22 @@
...
-  'chalk' : {red}
   './docker-setup' : DockerSetup
   'events' : {EventEmitter}
-  './exocom-setup' : ExocomSetup
+  '../../exosphere-shared' : {DockerHelper, compile-service-routes}
+  'fs-extra' : fs
+  'lodash/assign' : assign
   'path'
   './service-setup' : ServiceSetup
+  'js-yaml' : yaml
```
trushton: Alphabetize the imports. `js-yaml` between `fs-extra` and `lodash/assign`

alexdavid (approved):

exo-setup/src/app-setup.ls
```diff
@@ -23,36 +28,66 @@ class AppSetup extends EventEmitter
...
           ..on 'output', (data) ~> @emit 'output', data
     async.map-series setups, (-> &0.start &1), (err) ~>
       | err  =>  throw new Error err
+      done!
+
+
+  _get-exocom-docker-config: ->
+    exocom-docker-config =
+      exocom:
+        image: "originate/exocom:#{@app-config.bus.version}"
```
trushton, line 54: Do we want to genericize this now as preparation to [allow for different message buses](https://github.com/Originate/exosphere-sdk/milestone/26)?

exo-setup/src/app-setup.ls
```diff
@@ -23,36 +28,66 @@ class AppSetup extends EventEmitter
             location: service-data.location
             docker-image: service-data.docker_image
 
+    @_setup-services ~>
+      @_get-exocom-docker-config!
+      @_get-service-docker-config!
+      @_render-docker-compose!
+      @_setup-docker-images (err) ~>
+        | err => throw new Error err 
+        @logger.log role: 'exo-setup', text: 'setup complete'
```
trushton: `@write 'setup complete'`

exo-setup/src/app-setup.ls
```diff
@@ -23,36 +28,66 @@ class AppSetup extends EventEmitter
...
+
+  _render-docker-compose: ->
+    docker-compose-path = path.join(process.cwd!, 'docker-compose.yml') 
+    fs.ensure-file-sync docker-compose-path 
+    fs.write-file-sync docker-compose-path, yaml.safe-dump(@docker-compose-config)
+
+
+  _setup-docker-images: (done) ->
+    DockerHelper.pull-all-images {@write}, (exit-code, killed) ~>
+      | exit-code => @logger.log role: 'exo-setup', text: 'Docker setup failed'; done exit-code
```
trushton: You have the `@write` method, may as well use it here too.

exo-setup/src/app-setup.ls
```diff
@@ -23,36 +28,66 @@ class AppSetup extends EventEmitter
...
+      | otherwise =>
+        DockerHelper.build-all-images {@write}, (exit-code, killed) ~>
+          | exit-code => @logger.log role: 'exo-setup', text: 'Docker setup failed'; done exit-code
+          | otherwise => @logger.log role: 'exo-setup', text: 'Docker setup finished'; done!
```
trushton: and on these two lines

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,82 @@
...
-  '../../exosphere-shared' : {templates-path, call-args, DockerHelper}
-  'fs'
+  'handlebars' : Handlebars
+  'fs-extra' : fs
+  'prelude-ls' : {Obj, map}
   'js-yaml' : yaml
-  'observable-process' : ObservableProcess
   'path'
-  'shelljs' : {cp}
+  'os'
```
trushton: alphabetize

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,82 @@
...
+    | otherwise       => @_setup-external-service!
+
+
+  _setup-service: ->
+    docker-config = {}
+    docker-config[@role] =
+      Obj.compact do
+        build: @service-location 
+        command: @service-config.startup.command
+        ports: @service-config.docker?.ports or null
```
trushton: is the `or null` needed as the `compact` is being performed on this?

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,82 @@
...
+    for dependency, dependency-config of @service-config.dependencies
+      docker-config[dependency + dependency-config.dev.version] = @_setup-service-dependencies dependency, dependency-config.dev
+    docker-config
+
+
+  _get-service-links: ->
+    links = []
+    for dependency, dependency-config of @service-config.dependencies
+      links.push "#{dependency + dependency-config.dev.version}:#{dependency}"
+    if links.length then links else null
```
trushton: could use `that` in place of `links` after the if statement if you want.

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,82 @@
...
+      ROLE: @role
+      EXOCOM_HOST: 'exocom'
+      EXOCOM_PORT: '$EXOCOM_PORT'
+    for dependency of @service-config.dependencies
+      env-vars[dependency.to-upper-case!] = dependency
+    env-vars
+
+
+  _get-service-dependencies: ->
+    dependencies = ['exocom']
```
trushton: Is it possible that we have some services that aren't at all connected to `exocom` (or any other message bus) in the future? Not that it would need to be addressed right now, just something to consider.

@kevgo thoughts?

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,82 @@
...
+    for dependency, dependency-config of @service-config.dependencies
+      dependencies.push (dependency + dependency-config.dev.version)
+    dependencies
+
+
+  _setup-service-dependencies: (dependency-name, dependency-config) ->
+    if dependency-config.volumes
+      data-path = path.join os.homedir!, '.exosphere', @app-name, dependency-name, 'data'
+      fs.ensure-dir-sync data-path
+      rendered-volumes =  map ((volume) -> Handlebars.compile(volume)({"EXO_DATA_PATH": data-path})), dependency-config.volumes
```
trushton: remove a space between `=   map`

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,82 @@
...
+
+    dependency =
+      image: "#{dependency-config.image}:#{dependency-config.version}"
+      ports: dependency-config.ports 
+      volumes: rendered-volumes or null
+    Obj.compact dependency
```
trushton: Could this be
```
dependency = 
  Obj.compact do
    image: "#{dependency-config.image}:#{dependency-config.version}"
    ports: dependency-config.ports 
    volumes: rendered-volumes

File: exo-setup/src/docker-setup.ls [hunk 12]
```diff
@@ -1,67 +1,82 @@
...
+    
+  _setup-external-service: ->
+    throw new Error red "No location or docker-image specified" unless @docker-image
+    docker-config = {}
+    docker-config[@role] =
+      image: @docker-image
+      depends_on: ['exocom']
+    docker-config
+
```
trushton: 3 blank lines here

exo-setup/src/service-setup.ls
```diff
@@ -9,6 +9,7 @@ require! {
 }
 
 
+# Runs build command in each of the services
 class ServiceSetup extends EventEmitter
```
trushton, line 13: If all testing and running of services is being done inside of docker containers, do we still need this setup process run for each service? 
All it's doing is installing dependencies on the local machine. Those dependencies aren't even copied into the docker image that gets built (due to architecture differences between your local machine and the docker image), the dependencies are installed to the image in the build process itself.

 Speaking of which, currently the Dockerfile runs `npm install --production` whereaas ideally we would be running whatever setup operations are in `service.yml`.

Having this is still helpful if you want to manually run specific tests, but that could be built into `exo-test` in the future.

@kevgo @alexdavid thoughts?

exosphere-shared/src/compile-service-routes.ls
```diff
@@ -8,12 +8,13 @@ module.exports = (app-config, base-path) ->
   service-routes = []
   for protection-level of app-config.services
     for role, service-data of app-config.services["#{protection-level}"]
```
trushton: `"#{protection-level}" -> protection-level`

exosphere-shared/src/compile-service-routes.ls
```diff
@@ -8,12 +8,13 @@ module.exports = (app-config, base-path) ->
...
-      service-config = yaml.safe-load fs.read-file-sync(path.join(base-path ? process.cwd!, service-data.location, 'service.yml'), 'utf8')
-      service-routes.push do
-        {
-          role: role
-          receives: service-config.messages.receives
-          sends: service-config.messages.sends
-          namespace: service-data.namespace
-        }
+      if service-data.location
```
trushton: How do we get the routes of external services?

cli/features/tutorial.feature
```diff
@@ -101,10 +101,11 @@ Feature: Following the tutorial
     dependencies:
 
     docker:
-      publish:
+      ports:
     """
     When running "exo setup" in this application's directory
-    Then it has created the folders:
+    Then it finishes with exit code 0
```
kevgo: This is implicit in the previous step, because we say `When runing ...`. This step only passes if the command exits with exit code 0. If we said `When trying to run ...`, the step only passes if it exits with exit code > 0.

So, please make sure the `running ...` step checks for the exit code, and then remove this line.

cli/features/tutorial.feature
```diff
@@ -159,14 +160,17 @@ Feature: Following the tutorial
 
       dependencies:
         mongo:
-          version: '3.4.0'
-          docker_flags:
-            volume: '-v {{EXO_DATA_PATH}}:/data/db'
-            online_text: 'waiting for connections'
-            port: '-p 27017:27017'
+          dev:
```
kevgo, line 166: I'm still torn whether we should move the definition of dependencies to `application.yml` or leave it here. 

Ideally a service just says: "I need MongoDB version 3.4.0`. Or even better, it says "I need MongoDB version `3.x.x`. Then Exosphere collects all the requirements of all services and boots the smallest possible set of MongoDB image that satisfies all requirements of all services.

exo-setup/features/setup.feature
```diff
@@ -20,14 +21,77 @@ Feature: Setup of Exosphere applications
       | dashboard/Dockerfile     |
       | mongo-service/Dockerfile |
       | web-server/Dockerfile    |
+    And it has generated the file "docker-compose.yml" with the content:
+      """
+      version: '3'
```
kevgo: Are we checking this file into Github? If yes, then it should have a comment on top saying:

> This file is auto-generated. Please don't modify it.

If no, it shouldn't be generated in this directory. We can still generate it in a temp directory, but shouldn't bother the user with it.

exo-setup/src/app-setup.ls
```diff
@@ -23,36 +28,67 @@ class AppSetup extends EventEmitter
             location: service-data.location
             docker-image: service-data.docker_image
 
+    @_setup-services ~>
+      @_get-exocom-docker-config!
+      @_get-service-docker-config!
+      @_render-docker-compose!
+      @_setup-docker-images (err) ~>
+        | err => throw new Error err
```
kevgo: Are we sure we want to throw here? Why not exit the method with an error?

exo-setup/src/app-setup.ls
kevgo: It's probably better to do that all in one PR and keep this already massive PR as simple as possible.

exo-setup/src/app-setup.ls
```diff
@@ -23,36 +28,67 @@ class AppSetup extends EventEmitter
...
+        app-name: @app-config.name
+        role: service.role
+        logger: @logger
+        service-location: service.location
+        docker-image: service.docker-image
+      @docker-compose-config.services `assign` docker-setup.get-service-docker-config!
+
+
+  _render-docker-compose: ->
+    docker-compose-path = path.join(process.cwd!, 'docker-compose.yml')
```
kevgo: omit parentheses

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,85 @@
...
+        depends_on: @_get-service-dependencies!
+    for dependency, dependency-config of @service-config.dependencies
+      docker-config[dependency + dependency-config.dev.version] = @_setup-service-dependencies dependency, dependency-config.dev
+    docker-config
+
+
+  _get-service-links: ->
+    links = []
+    for dependency, dependency-config of @service-config.dependencies
+      links.push "#{dependency + dependency-config.dev.version}:#{dependency}"
```
kevgo: You could call `map` on `@service-config.dependencies` to do this on 2 lines instead of 3.

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,85 @@
...
+    if links.length then links else null
```
kevgo: Why do we return `null` instead of an empty array? That seems counter-intuitive. This method has the name `_get-service-links`, it should return an array of service links, sometimes empty, sometimes not.

exo-setup/src/docker-setup.ls
kevgo: ExoCom is defined in `application.yml`. It is an application dependency, not a service dependency. So it does not necessarily have to show up here.

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,85 @@
...
+  _get-service-dependencies: ->
+    dependencies = ['exocom']
+    for dependency, dependency-config of @service-config.dependencies
+      dependencies.push (dependency + dependency-config.dev.version)
+    dependencies
+
+
+  _setup-service-dependencies: (dependency-name, dependency-config) ->
+    if dependency-config.volumes
+      data-path = path.join os.homedir!, '.exosphere', @app-name, dependency-name, 'data'
```
kevgo: Extract a `global-exosphere-directory` helper function

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,85 @@
...
+
+    dependency =
+      image: "#{dependency-config.image}:#{dependency-config.version}"
+      container_name: dependency-name + dependency-config.version
+      ports: dependency-config.ports 
+      volumes: rendered-volumes or null
+    Obj.compact dependency
+    
+  _setup-external-service: ->
+    throw new Error red "No location or docker-image specified" unless @docker-image
```
kevgo:
```ls
| !@docker-image  =>  throw new Error red "No location or docker-image specified"
```

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,85 @@
...
+    docker-config = {}
+    docker-config[@role] =
+      image: @docker-image
+      container_name: @role
+      depends_on: ['exocom']
+    docker-config
```
kevgo, line 89: This method's name implies that it sets up an external service, but all it does is return a hash variable.

exo-setup/src/service-setup.ls
kevgo: It must be possible to write my own services that run in a particular Docker base image and get set up by whatever command I provide to Exosphere. Also, if I modify the `package.json` file of the service, it needs to run `yarn install` again inside the Docker image. So this command must be specified somewhere, and we need a `ServiceSetup` tool to run them. 

How does dependency installation work inside Docker images? Does Docker run this automatically when booting up the image?

kevgo (commented):
A good start. I think we need 3 waves of starts:
1. start application dependencies
2. start service dependencies
3. start services

Btw, this PR is way too big. It's reckless to make such huge changes. Neither you nor any reviewer can be sure these changes make sense or are complete in their entirety. It's even impossible to review this in one setting and keep all the changes made in one's mind. In addition, this PR makes a whole number of independent changes. Please try to keep your PRs more concise.

exo-setup/src/service-setup.ls
hugoesthere: The setup commands that @kevgo is talking about can be defined in `service.yml`, and then ran when the image is built. It can be part of our Dockerfile template to just take this command from `service.yml`

Whenever `package.json` or `service.yml` is modified then `exo setup` has to be re-run so the Docker image can be rebuilt. Otherwise build and dependency commands can be run during the build phase of a Docker image and this file can be removed.

I've made a ticket to track this: https://github.com/Originate/exosphere-sdk/issues/283

exosphere-shared/src/compile-service-routes.ls
hugoesthere: We run `docker run --rm=true #{image} cat service.yml`. `compile-service-routes.ls` was being used on local services only anyways, this just centralizes the `if service-data.location` check and makes this function more robust.

exo-setup/src/docker-setup.ls
hugoesthere: I agree, these dependencies should depend on whatever message bus is defined in `application.yml`.  I think this change can be made once we actually provide support for different message buses. But the message bus does need to be listed as a service dependency so that `docker-compose` knows to boot up these services only after the message bus is boot up.

The issue may be that the word 'dependency' is overloaded here.

exo-setup/features/setup.feature
hugoesthere: I'll put it in a tmp dir. This config file should be accessible to the user for reading but it shouldn't be modified nor checked into GitHub.

exo-setup/src/docker-setup.ls
hugoesthere: How would I be able to access both `dependency` and `dependency-config` fields?

exo-setup/src/docker-setup.ls
hugoesthere: `_setup-service` uses `Obj.compact` which does not exclude a field if it contains a value which is an empty array. I could move this logic up into `_setup-service` but that would mean we couldn't inline `links: @_get-service-dependencies!`

cli/features/tutorial.feature
hugoesthere: As we discussed, we will list dependencies and their meta data in `application.yml`. Services will just list `mongo3.4.0` or something as a dependency. Then a user could even run `exo add #{dependency-name}` to add a dependency to `application.yml`. @kevgo can you write the ticket for this? I want to make sure all the details and requirements are correct.

hugoesthere:
@kevgo all comments addressed

cli/features/tutorial.feature
kevgo: I have created https://github.com/Originate/exosphere-sdk/issues/284 to track this.

exo-setup/features/setup.feature
```diff
@@ -11,7 +11,8 @@ Feature: Setup of Exosphere applications
   Scenario: set up the "test" application
     Given a freshly checked out "test" application
     When running "exo-setup" in this application's directory
-    Then it has created the folders:
+    Then it finishes with exit code 0
```
kevgo: Please remove this line everywhere from this PR. Exit code 0 is checked by the previous line `When running ...`.

cli/features/steps/steps.ls
```diff
@@ -158,3 +160,7 @@ module.exports = ->
       @browser.assert.success!
       expect(@browser.text 'body').to.include expected-content.replace(/\n/g, '')
       done!
+
+
+  @Then /^it finishes with exit code (\d+)$/ (+expected-exit-code) ->
```
kevgo: remove

exo-setup/features/setup.feature
```diff
@@ -20,14 +21,77 @@ Feature: Setup of Exosphere applications
...
+      | test_users       |
+      | test_web         |
+      | originate/exocom |
 
   Scenario: set up an application with external Docker images
     Given a freshly checked out "app-with-external-docker-images" application
     When running "exo-setup" in this application's directory
-    Then it has acquired the Docker images:
-      | test-web-server |
+    Then it finishes with exit code 0
```
kevgo: remove

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,96 @@
...
+  get-service-docker-config: ~> 
+    | @service-config => @_get-service-docker-config!
+    | otherwise       => @_get-external-service-docker-config!
+
+
+  # builds the Docker config for a service and its dependencies
+  _get-service-docker-config: ->
+    docker-config = {}
+    docker-config[@role] =
+      Obj.compact do
```
kevgo: Can you merge this line with the previous line?

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,96 @@
...
+        build: path.join '..', @service-location 
+        container_name: @role
+        command: @service-config.startup.command
+        ports: @service-config.docker?.ports or null
```
kevgo: Does `or null` make sense here? Is `null` so much better than `undefined`?

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,96 @@
...
+      volumes: rendered-volumes or null
+    
+
+  _get-external-service-docker-config: ->
+    | !@docker-image => throw new Error red "No location or docker-image specified"
+    docker-config = {}
+    docker-config[@role] =
+      image: @docker-image
+      container_name: @role
+      depends_on: ['exocom']
```
kevgo, line 88: External services might not depend on exocom

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,96 @@
...
+    docker-config
+
+
+  _global-exosphere-directory: (dependency-name) ->
```
kevgo: This helper should not be in the DockerSetup class. Make it a global method?

exo-setup/src/docker-setup.ls
hugoesthere: Correct. There is a ticket tracking the need to extract such config information out of external services: https://github.com/Originate/exosphere-sdk/issues/274. This is also something we should worry about when we begin providing support for message buses other than exocom

hugoesthere:
@kevgo all comments addressed.

exo-setup/features/setup.feature
```diff
@@ -20,14 +20,76 @@ Feature: Setup of Exosphere applications
       | dashboard/Dockerfile     |
       | mongo-service/Dockerfile |
       | web-server/Dockerfile    |
+    And it has generated the file "tmp/docker-compose.yml" with the content:
```
kevgo: This file is not getting checked into Git, so it isn't relevant to the user. It is a technical implementation detail. Let's remove it from this spec.

exo-setup/features/setup.feature
```diff
@@ -20,14 +20,76 @@ Feature: Setup of Exosphere applications
...
+      | test_users       |
+      | test_web         |
+      | originate/exocom |
 
   Scenario: set up an application with external Docker images
     Given a freshly checked out "app-with-external-docker-images" application
     When running "exo-setup" in this application's directory
-    Then it has acquired the Docker images:
-      | test-web-server |
+    Then it has generated the file "tmp/docker-compose.yml" with the content:
```
kevgo: same here

exo-setup/features/steps/steps-then.ls
```diff
@@ -24,5 +25,7 @@ module.exports = ->
       expect(docker-images).to.include row[0]
 
 
-  @Then /^it finishes with exit code (\d+)$/ (+expected-exit-code) ->
-    expect(@process.exit-code).to.equal expected-exit-code
+  @Then /^it has generated the file "([^"]*)" with the content:$/ (filename, expected-content, done) ->
```
kevgo: This step can go away as well

kevgo (commented):
Once you remove the DockerCompose files from the specs, this is good to go!

cli/features/steps/steps.ls
```diff
@@ -158,3 +160,5 @@ module.exports = ->
       @browser.assert.success!
       expect(@browser.text 'body').to.include expected-content.replace(/\n/g, '')
       done!
+
+
```
ben-goldberg: looks like you added some extra newlines here

exo-run/src/service-restarter.ls
```diff
@@ -0,0 +1,52 @@
...
+        @logger.log {role: 'exo-run', text: "Restarting service '#{@role}' because #{removed-path} was deleted"}
+        @_restart!
+
+
+  _restart: ->
+    @watcher.close!
+    cwd = @docker-config-location 
+    DockerHelper.kill-container {service-name: @role, cwd, @write}, (exit-code) ~>
+      | exit-code => @emit 'error', "Docker failed to kill container #{@role}"
+      @write "Docker container stopped"
```
ben-goldberg: I don't know if we have style preferences for this, but don't we usually prefer single quotes to double, when possible? If so, same would go for line 40.

exo-setup/features/steps/steps-then.ls
```diff
@@ -24,5 +25,5 @@ module.exports = ->
       expect(docker-images).to.include row[0]
 
 
-  @Then /^it finishes with exit code (\d+)$/ (+expected-exit-code) ->
-    expect(@process.exit-code).to.equal expected-exit-code
+  @Then /^it finishes with exit code (\d+)$/ (+expected-exit-code) ->
```
ben-goldberg: unnecessary whitespace here and on next line

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,92 @@
...
+    docker-config[@role] = Obj.compact do
+      build: path.join '..', @service-location 
+      container_name: @role
+      command: @service-config.startup.command
+      ports: @service-config.docker?.ports or undefined
+      links: @_get-docker-links!
+      environment: @_get-docker-env-vars! 
+      depends_on: @_get-service-dependencies!
+    for dependency, dependency-config of @service-config.dependencies
+      docker-config[dependency + dependency-config.dev.version] = @_get-service-dependency-docker-config dependency, dependency-config.dev
```
ben-goldberg, line 36: prefer string interpolation over `+`?

exo-setup/src/docker-setup.ls
```diff
@@ -1,67 +1,92 @@
...
+    docker-config
+
+
+  # compiles list of container links from service to dependency
+  # returns undefined if length is 0 so it can be ignored with Obj.compact
+  _get-docker-links: ->
+    links = []
+    for dependency, dependency-config of @service-config.dependencies
+      links.push "#{dependency + dependency-config.dev.version}:#{dependency}"
```
ben-goldberg, line 45:
```
links.push "#{dependency}#{dependency-config.dev.version}:#{dependency}"
```
if you think that looks better

exo-setup/src/docker-setup.ls
hugoesthere: I think `+` makes sense here, since we aren't using literal string characters it doesn't seem necessary to enter into a string just to escape it again.

exo-setup/src/docker-setup.ls
hugoesthere: I'll keep this as `+` just to stay consistent with line 36 above.

exo-setup/src/docker-setup.ls
ben-goldberg: Sure! It's a ruby convention to prefer interpolation for concatenation instaed of using +, but I wasn't sure if that was a convention in livescript too

PR #243, opened 2017-05-25
WIP: switching exo-run to use docker-compose

hugoesthere:
Switch `exo-run` to using `docker-compose` commands. Tests pass for the exo-run subproject, but the end to end tests don't pass yet.


exo-run/features/steps/steps-given.ls
```diff
@@ -18,4 +19,4 @@ module.exports = ->
                                        cwd: @app-dir,
                                        stdout: dim-console.process.stdout
                                        stderr: dim-console.process.stderr)
-      done!
+      wait 10_000, done
```
hugoesthere, line 22: Since we're using `docker-compose` to boot up our services, we can't individually monitor each service in its own ObservableProcess, meaning we can't wait for each of them to come online and then finally emit a message such as 'all services online'. That control is something we hand over when switching to `docker-compose`. 

I'm not sure how this step definition was working previously when we weren't waiting for any sort of online text. Regardless, that's why we're waiting for 10 seconds instead of for some sort of online text to ensure that an application has come online.

Open to suggestions for alternatives though.

exo-run/features/steps/steps-then.ls
```diff
@@ -2,32 +2,37 @@ require! {
...
+  '../../../exosphere-shared' : {compile-service-routes}
+  'fs'
   'jsdiff-console'
   'nitroglycerin' : N
-  'prelude-ls' : {last}
+  'prelude-ls' : {any, last}
+  'os'
+  'path'
   'request'
+  'jsonic'
```
hugoesthere: We can get rid of this once this ticket has been resolved: https://github.com/Originate/exosphere-sdk/issues/279

exosphere-shared/src/docker-helper.ls
```diff
@@ -79,13 +117,13 @@ class DockerHelper
...
-      ..on 'ended', (exit-code, killed) ->
-        | exit-code > 0 and not killed  =>  return done "Dependency #{container.container-name} failed to start, shutting down"
-      ..wait container.online-text, done
+  # @start-container = (container, done) ~>
+  #   new ObservableProcess("docker start -a #{container.container-name}",
+  #                           stdout: false,
+  #                           stderr: false)
+  #     ..on 'ended', (exit-code, killed) ->
+  #       | exit-code > 0 and not killed  =>  return done "Dependency #{container.container-name} failed to start, shutting down"
+  #     ..wait container.online-text, done
```
hugoesthere, line 134: Planning on cleaning up the rest of this `docker-helper.ls` file in another PR. Just wanted to make sure this function wasn't needed elsewhere before completely getting rid of it.

exo-run/features/steps/steps-then.ls
```diff
@@ -2,32 +2,37 @@ require! {
...
 module.exports = ->
 
-  @Then /^ExoCom uses this routing:$/ timeout: 10_000, (table, done) ->
-    expected-routes = {}
+  @Then /^ExoCom uses this routing:$/ (table) ->
+    expected-routes = []
     for row in table.hashes!
-      expected-routes[row.ROLE] or= {}
+      service-routes = {}
+      service-routes.role = row.ROLE
```
ben-goldberg: these two lines could be:
```
service-routes = {role: row.ROLE}
```
or whatever clean livescript syntax is to do that

exo-run/features/steps/steps-given.ls
trushton: The first alternative that comes to mind, though I'm not sure its a great option by any means, is that we have a docker image built `FROM` something tiny like the `alpine` image (5MB). that simply prints a message such as `all services online`, and `depends_on` all other services of the application so that it would only be run after all other containers are ready.

I'll keep looking into better options.

exo-run/features/steps/steps-then.ls
```diff
@@ -2,32 +2,37 @@ require! {
...
   'jsdiff-console'
   'nitroglycerin' : N
-  'prelude-ls' : {last}
+  'prelude-ls' : {any, last}
+  'os'
+  'path'
   'request'
+  'jsonic'
+  'js-yaml' : yaml
   'wait' : {wait}
```
trushton: Alphabetize `require`s

exo-run/features/steps/steps-then.ls
```diff
@@ -47,13 +52,12 @@ module.exports = ->
...
+    @process.wait /ExoCom WebSocket listener online at port/, done
 
 
-  @Then /^my machine is running the services:$/ timeout: 10_000, (table, done) ->
-    async.each [row['NAME'] for row in table.hashes!],
-               ((name, cb) ~> @process.wait "'#{name.to-lower-case!}' is running", cb),
-               done
+  @Then /^my machine is running the services:$/ (table) ->
+    for row in table.hashes!
+      expect(child_process.exec-sync('docker ps --format {{.Names}}/{{.Status}}') |> (.to-string!) |> (.split os.EOL) |> any (.includes "#{row.NAME}/Up")).to.be.true
```
trushton, line 59: Use `docker-helper`s  `container-is-running` method here.
```
expect( DockerHelper.container-is-running( row.NAME )).to.be.true
```

exo-run/src/app-runner.ls
```diff
@@ -1,74 +1,44 @@
...
+  './service-watcher' : ServiceWatcher
 }
 
 
 # Runs the overall application
 class AppRunner extends EventEmitter
 
   ({@app-config, @logger}) ->
+    @env =
+      EXOCOM_PORT: 80 or process.env.EXOCOM_PORT
```
trushton: I think this should be `process.env.EXOCOM_PORT or 80` otherwise the `80` will always succeed first and be used.

exo-run/src/app-runner.ls
```diff
@@ -1,74 +1,44 @@
...
-            ROLE: 'exocom'
-          publish:
-            EXOCOM_PORT: "#{@exocom-port}:#{@exocom-port}"
-        @exocom = new DockerRunner {role: 'exocom', @docker-config, @logger}
-          ..start-service!
-          ..on 'error', (message) ~> @shutdown error-message: message
+  start: ->
+    @watch-services!
+    DockerHelper.run-all-images {@env, @write}, (exit-code, killed) ~>
+      | exit-code => return @shutdown error-message: 'Failed to run images'
```
trushton: What if it was `killed`?

exo-run/src/service-watcher.ls
```diff
@@ -0,0 +1,48 @@
...
+        @_restart!
+      ..on 'unlink', (removed-path) ~>
+        @logger.log {role: 'exo-run', text: "Restarting service '#{@role}' because #{removed-path} was deleted"}
+        @_restart!
+
+
+  _restart: ->
+    @watcher.close!
+    DockerHelper.kill-container {service-name: @role, @write}, (exit-code) ~>
+      | exit-code => @write "Docker failed to kill container #{@role}"; process.exit exit-code
```
trushton: Shouldn't the error be bubbled up and trigger the `shutdown` of the application?

exo-run/src/service-watcher.ls
```diff
@@ -0,0 +1,48 @@
...
+      @write "Docker container stopped"
+
+      DockerHelper.create-new-container {service-name: @role, @env, @write}, (exit-code) ~>
+        | exit-code => @write "Docker image failed to rebuild"; process.exit exit-code
```
trushton: And here

exo-run/src/service-watcher.ls
```diff
@@ -0,0 +1,48 @@
...
+        @write "Docker image rebuilt"
+
+        DockerHelper.start-container {service-name: @role, @env, @write}, (exit-code) ~>
+          | exit-code => @write "Docker container failed to restart"; process.exit exit-code
```
trushton: And here

trushton (changes requested):
This looks great overall! I'm liking this move to `docker-compose` already.

exo-run/features/steps/steps-given.ls
hugoesthere: That won't work because `depends_on` only checks if a container has been started, not if it has properly booted up and is ready to accept connections.

exo-run/features/steps/steps-then.ls
hugoesthere: 👍 We are getting rid of the old `DockerHelper` command line methods and switching to `dockerode`. I'm working on a PR that does this, so this line will eventually reflect that.

exo-run/src/app-runner.ls
hugoesthere: Good catch 👍

exo-run/src/app-runner.ls
hugoesthere: If this process was killed, `cli.ls` already has a listener for `SIGINT` and `shutdown` will be called. I can remove that extra param so there isn't any confusion.

exo-run/src/service-watcher.ls
hugoesthere: Good call 👍

exo-run/features/steps/steps-given.ls
kevgo: Docker Compose gives us output that contains the output of each service, multiplexed. If we create a parser for it's output, we can de-multiplex it and create a bunch of individual streams.

hugoesthere: Yes, as discussed this will be handled in a later PR.

hugoesthere:
@kevgo all comments have been addressed

exo-run/src/service-watcher.ls
```diff
@@ -0,0 +1,52 @@
+require! {
+  'chokidar' : {watch}
+  'events' : {EventEmitter}
+  '../../exosphere-shared' : {DockerHelper}
+  'path'
+}
+
+
+# Watches local services for changes and restarts them
+class ServiceWatcher extends EventEmitter
```
kevgo: Better name for this class: `ServiceRestarter`

kevgo (approved):

PR #244, opened 2017-05-25
WIP: switch exo-test to use dockerode

hugoesthere:
Switch `exo test`, the remaining command that runs Docker processes, to using `dockerode`. This PR also cleans up the rest of the previous DockerHelper class, and extracts a DockerCompose class. DockerHelper will use `dockerode` and DockerCompose will use `docker-compose`

Tests for the `exo test` subproject pass; working on another PR to get end-to-end tests passing again.


exosphere-shared/src/docker-helper.ls
```diff
@@ -1,146 +1,49 @@
...
+  @start-container = ({Image, name, HostConfig}, done) ->
+    DockerHelper._container-is-running name, (err, is-running) ->
+      | err => done err
+      | is-running => done!
+      | otherwise =>
+        docker.create-container {Image, name, HostConfig}, (err, container) -> 
+          | err => done err
+          container.start (err, data) ->
+            | err => done err
+            wait 2_000, done
```
hugoesthere, line 21: I put this in as a temporary fix until https://github.com/Originate/exosphere-sdk/issues/277 is resolved. Open to alternatives.

exosphere-shared/src/docker-compose.ls
```diff
@@ -0,0 +1,59 @@
...
+  @run-all-images = ({env, write}, done) ->
+    new ObservableProcess('docker-compose up'
+                          env: env
+                          stdout: {write}
+                          stderr: {write})
+      ..on 'ended', done
+
+
+  @kill-container = ({service-name, write}, done) ->
+    new ObservableProcess("docker-compose kill #{service-name}"
```
kevgo: Let's follow the lessons learned from the DockerHelper and provide the arguments as an array, so that we don't run into issues with escaping strings for shell usage.

```ls
new ObservableProcess(['docker-compose', 'kill', service-name]

File: exosphere-shared/src/docker-helper.ls (hunk 1 above)
Comment (kevgo) [inline exosphere-shared/src/docker-helper.ls:21]: This seems slow and brittle. Is there no better way to do this?

File: exosphere-shared/src/docker-helper.ls [hunk 3]
```diff
@@ -1,146 +1,49 @@
...
-  @pull-all-images = ({write}, cb) ->
-    new ObservableProcess('docker-compose pull'
-                          stdout: {write}
-                          stderr: {write})
-      ..on 'ended', cb
 
+  @remove-container = ({name}, done) ->
+    docker.list-containers {name}, (err, containers) ->
+      | err => done err
+      DockerHelper._force-remove-containers containers, done
```
kevgo, line 27: This is the same code as `@remove-all-containers`. Are you sure this is correct? What's the difference between the two methods?

kevgo: Are you sure you want to always force-remove containers?

exosphere-shared/src/docker-helper.ls
```diff
@@ -1,146 +1,49 @@
...
-  @kill-all-containers = ({write}, cb) ->
-    new ObservableProcess('docker-compose down'
-                          stdout: {write}
-                          stderr: {write})
-      ..on 'ended', cb
 
+  @_container-is-running = (name, done) ->
+    docker.list-containers {name}, (err, containers) ->
+      | err => done err
+      done null, any (.Names.includes name), containers
```
kevgo: Are you calling `done` with 3 parameters here?

exosphere-shared/src/docker-helper.ls
```diff
@@ -1,146 +1,49 @@
...
-
-  @get-build-command = (image, build-flags) ->
-    return "docker build -t #{image.author}/#{image.name} #{if build-flags then build-flags else ""} ."
-
-
-  @get-config = (image) ->
-    child_process.exec-sync("docker run --rm=true #{image} cat service.yml", 'utf8') |> (.to-string!)
-
-
-  @get-docker-ip = (container) ->
```
kevgo, line 91: What about this method? No longer used anywhere?

exosphere-shared/src/docker-helper.ls
trushton: `@remove-container` additionally passes a `{name}` to `docker.list-containers`, meaning that only containers matching the name will be passed, and therefore removed in the `_force-remove-containers` call. 

However, I think these two could be combined by having a default value for `name` of an empty string?
```
@remove-containers = ({name = ''}, done) -> 
    docker.list-containers {name}, (err, containers) ->
        | err                 => done err
        | !containers.length  => done!
        | _                   => DockerHelper._force-remove-containers containers, done
```

trushton (commented):
LGTM other than the comments above that need addressing, will approve once that's done

exo-run/src/app-runner.ls
```diff
@@ -1,7 +1,7 @@
 require! {
   'chalk': {red}
   'events' : {EventEmitter}
-  '../../exosphere-shared' : {DockerHelper}
+  '../../exosphere-shared' : {DockerCompose}
```
kevgo, line 4: This change does not affect `exo-test`, but `exo-run`, and it switches it to use Docker Compose, not Dockerode. Please separate these changes into a separate PR.

exosphere-shared/src/docker-helper.ls
hugoesthere: I agree, I just put this in as a temporary fix. In the future we want to add multiple listeners to this process as we discussed here: https://github.com/Originate/exosphere-sdk/issues/280

But I agree, I am uncomfortable with this implementation and am open to any other ideas.

kevgo: Let's hope that after #277 and #280 are shipped, this can be cleaned up.

exosphere-shared/src/docker-helper.ls
kevgo: One should be named `remove-container(name)`, the other `remove-containers`.

exosphere-shared/src/docker-helper.ls
hugoesthere: No, there's no need to manually query container IPs in order to link them to each other. `docker-compose` handles linking between containers now.

exo-run/src/app-runner.ls
hugoesthere: These changes are tied to this PR. This PR is really just implementing Dockerode, but DockerHelper had to be renamed to DockerCompose to distinguish it from the Dockerode processes which uses that name. There has been no extra logic added to DockerCompose.

hugoesthere:
@kevgo all comments have been addressed.

kevgo (approved):

trushton (approved):

PR #245, opened 2017-05-26
Adding service prompts for protection level

ben-goldberg:
resolves https://github.com/Originate/exosphere-sdk/issues/229

When adding a new service, the user is now prompted to give a protection level, either 'public' or 'private'. This choice is then reflected in the application.yml file.

@kevgo @hugobho @trushton

### DIFF
--- a/exo-add/features/steps/steps-then.ls
+++ b/exo-add/features/steps/steps-then.ls
@@ -11,7 +11,6 @@ module.exports = ->
 
   @Then /^my application contains the file "([^"]*)" with the content:$/, (file-path, expected-content, done) ->
     fs.read-file path.join(@app-dir, file-path), N (actual-content) ->
-      console.log actual-content.to-string!
       try
         jsdiff-console actual-content.to-string!trim!, expected-content.trim!
       catch
--- a/exo-add/package.json
+++ b/exo-add/package.json
@@ -11,7 +11,7 @@
     "nitroglycerin": "1.1.2",
     "prelude-ls": "1.1.2",
     "tmplconv": "3.2.1",
-    "yaml-cutter": "^0.2.1"
+    "yaml-cutter": "0.2.1"
   },
   "description": "The 'exo-add' command",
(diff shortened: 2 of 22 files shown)


trushton (approved):
Just need to update the feature tests in `exo-create` and it should all work

exo-add/features/steps/steps-then.ls
```diff
@@ -11,7 +11,6 @@ module.exports = ->
 
   @Then /^my application contains the file "([^"]*)" with the content:$/, (file-path, expected-content, done) ->
     fs.read-file path.join(@app-dir, file-path), N (actual-content) ->
-      console.log actual-content.to-string!
```
kevgo, line 14: Thanks for cleaning this up!

exo-add/package.json
```diff
@@ -11,7 +11,7 @@
     "nitroglycerin": "1.1.2",
     "prelude-ls": "1.1.2",
     "tmplconv": "3.2.1",
-    "yaml-cutter": "0.1.3"
+    "yaml-cutter": "^0.2.1"
```
kevgo: Please remove the `^` from here.

exo-setup/src/cli.ls
```diff
@@ -14,7 +14,7 @@ module.exports = ->
 
   app-config = yaml.safe-load fs.read-file-sync('application.yml', 'utf8')
   console.log "Setting up #{green app-config.name} #{cyan app-config.version}\n"
-  logger = new Logger flatten [Object.keys(app-config.services[type]) for type of app-config.services]
+  logger = new Logger flatten [Object.keys(app-config.services[type]) unless app-config.services[type] is null for type of app-config.services]
```
kevgo: We don't care if the services are `null`, `undefined`, or an empty array. Hence, this change can be simplified from `unless app-config.services[type] is null` to `if app-config.services[type]`

kevgo (approved):
This is cool, thanks!

PR #247, opened 2017-05-30
Bg dry exo service creation

ben-goldberg:
resolves https://github.com/Originate/exosphere-sdk/issues/282

There was a bunch of shared code between `exo create service` and `exo add`. This PR DRYs up some of that.

@kevgo @hugobho @trushton


exo-add/features/without-arguments.feature
```diff
@@ -9,14 +9,14 @@ Feature: interactive scaffolding
...
-      | Author                        | tester  |
-      | Template                      |         |
-      | Name of the data model        | web     |
-      | Protection level              | public  |
+      | FIELD                         | INPUT          |
+      | Role of the service to create | web            |
+      | Type of the service to create | web            |
+      | Description                   | testing        |
+      | Author                        | tester         |
+      | Template                      | exoservice-es6 |
```
kevgo, line 17: Does this work? Afaik there is no way to select things in the wizard, because one would have to hit the _down_ cursor key a number of times and then hit Enter.

exo-create/src/entities/service.ls
```diff
@@ -16,7 +16,7 @@ service = ->
 
   console.log 'We are about to create a new Exosphere service!\n'
 
-  {data, questions} = parse-command-line process.argv
+  {data, questions} = ServiceAdder.parse-command-line process.argv.slice 3
```
kevgo: Why `.slice 3` at the end?

exosphere-shared/src/service-adder.ls
```diff
@@ -0,0 +1,93 @@
+require! {
+  'fs'
+  'minimist'
+  'path'
+}
+
+
+class ServiceAdder
```
kevgo: Please add documentation for this class

kevgo (approved):
This is a very awesome cleanup!

exo-add/features/without-arguments.feature
ben-goldberg: Yup! it was unhappy with leaving it empty, and this is the default choice

exo-create/src/entities/service.ls
ben-goldberg: I guess it's not strictly necessary! I'll rip that part out

trushton (approved):

PR #249, opened 2017-05-31
Organize web server templates

trushton:
resolves [#65](https://github.com/Originate/exosphere-sdk/issues/65#issuecomment-305054269)

Not sure what happened with this ticket from last year, but this should address the changes that needed to be made.

@kevgo @hugobho


kevgo (approved):
Thanks for fixing this!

trushton:
@kevgo Had to fix the some things, `exprestive` was expecting a certain file/directory structure by default

kevgo:
Good to go once tests are green

PR #254, opened 2017-06-01
Implement online text listener for internal services 

hugoesthere:
partially resolves https://github.com/Originate/exosphere-sdk/issues/286

TODO has been tracked on the same issue (#286)


trushton (approved):
👍

exo-run/features/run.feature
```diff
@@ -52,5 +52,5 @@ Feature: running Exosphere applications
 
 
   Scenario: a service crashes during startup
-    Given a running "crashing-service" application
+    When running a "crashing-service" application
```
kevgo: This app crashes, so we should say `When trying to run the "crashing-service" application`

exo-run/features/steps/steps-when.ls
```diff
@@ -29,6 +29,15 @@ module.exports = ->
       ..on 'ended', -> done!
 
 
+  @When /^running a "([^"]*)" application$/ timeout: 600_000, (@app-name, done) ->
+    @checkout-app @app-name
+    @app-dir := path.join process.cwd!, 'tmp', @app-name
+    @setup-app @app-dir, ~>
+      command = \exo-run
+      if process.platform is 'win32' then command += '.cmd'
+      @run-app {command}, done
```
kevgo: Check that the error code is 0 here.

kevgo (commented):
Cool stuff!

exo-run/features/steps/steps-when.ls
hugoesthere: The exit code should actually be non-zero here. But I don't think I should check for an error code at all since the step is "When trying to run the '...' application" (after I address the comment above that is).

charlierudolph: This functions looks to be a near exact copy of `Given a running "crashing-service" application` aside from the online-text. Could that be dried up?

exo-run/src/app-runner.ls
```diff
@@ -39,6 +54,16 @@ class AppRunner extends EventEmitter
     DockerCompose.kill-all-containers {cwd: @docker-config-location, @write}, -> process.exit exit-code
 
 
+  _compile-online-text: ->
+    online-texts = {}
+    for protection-level of @app-config.services
+      for role, service-data of @app-config.services[protection-level]
+        if service-data.location #TODO: compile online text for external services
```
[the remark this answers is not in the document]
charlierudolph: @hugobho does this still need to be fixed?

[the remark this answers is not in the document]
hugoesthere: Yes, but unfortunately it's blocked on this: https://github.com/Originate/exosphere-sdk/issues/278

I've made sure to track this in this ticket: https://github.com/Originate/exosphere-sdk/issues/274

PR #255, opened 2017-06-01
application dependencies

charlierudolph:
Addresses a https://github.com/Originate/exosphere-sdk/issues/284

Todo: update exo-deploy once terraform refactor is complete


kevgo (approved):

hugoesthere (approved):
LGTM 👍

exo-setup/src/docker-setup.ls
```diff
@@ -50,42 +50,48 @@ class DockerSetup
...
     | !@docker-image => throw new Error red "No location or docker-image specified"
+    dependencies = []
+    for dependency-config in @app-config.dependencies
+      dependencies.push "#{dependency-config.type}#{dependency-config.version}"
     docker-config = {}
     docker-config[@role] =
       image: @docker-image
       container_name: @role
-      depends_on: ['exocom']
+      depends_on: dependencies
```
charlierudolph: Could this just be `@_get-service-dependencies!`?

charlierudolph: Also does this need the env variables as well? (or anything else in `_get-service-docker-config`)?

trushton (approved):

exo-setup/src/docker-setup.ls
hugoesthere: `_get-service-dependencies` accesses local service config which an external service doesn't have, but I can extract a `_get-app-dependencies` for both to use.

We have yet to provide env vars and other config for external images has yet to be implemented: https://github.com/Originate/exosphere-sdk/issues/274

bin/publish
```diff
@@ -93,16 +93,15 @@ function update-exocom-package-versions
     process.exit 1
 
 
-function update-exocom-bus-number
+function update-exocom-dependency-number
   try
     replace.sync do
       files: ['features/**/*.feature', 'exosphere-shared/**/*.yml']
-      replace: /bus:(\s*)type: exocom(\s*)version: \d+.\d+.\d+/g
+      replace: /type: exocom(\s*)version: \d+.\d+.\d+/g
```
hugoesthere: We should probably also add a check for this now: https://github.com/Originate/exosphere/blob/12d35cdf89eabdbfff35bac237a67a0a464e11dd/exo-run/features/steps/steps-then.ls#L33

cli/features/tutorial.feature
```diff
@@ -36,9 +36,9 @@ Feature: Following the tutorial
       description: A todo application
       version: 0.0.1
 
-      bus:
-        type: exocom
-        version: 0.21.7
+      dependencies:
+        - type: exocom
+          version: 0.21.7
```
hugoesthere, line 41: Something else to consider...if we eventually want to support configuring these app wide depenencies, should we format it like so:
```
dependencies:
  - type: exocom
      version: 0.21.7
  - type: mongo3.4.0
      version: 3.4.0
      ports:
        - '27017:27017'
      environment-variables:
         - EXOCOM_PORT: 3100
      volumes:
         - {some_mount..}
```

in which case we might even consider changing `type` to `name`
```
dependencies:
  - name: exocom
      version: 0.21.7
  - name: mongo3
      version: 3.4.0
      ports:
        - '27017:27017'
      environment-variables:
         - EXOCOM_PORT: 3100
      volumes:
         - {some_mount..}
```

kevgo: +1 on using `name` here

kevgo (commented):
Looks great!

bin/publish
charlierudolph: Fixed

cli/features/tutorial.feature
charlierudolph: Sticking with `type` and the name is going to be the type and version concatenated

hugoesthere (approved):
🚀

charlierudolph:
merged in c75f0fc

PR #256, opened 2017-06-02
exo-run: update cucumber

charlierudolph:


hugoesthere (approved):
Looks good, but seems like tests are still red :(

Travis is complaining about the cucumber run command: 
```
error: unknown option '--snippet-syntax'
```

PR #257, opened 2017-06-02
update publish script

charlierudolph:


hugoesthere (approved):
I'm assuming this is to be shipped before #255 right?

bin/publish
```diff
@@ -96,13 +96,13 @@ function update-exocom-package-versions
...
-      files: ['features/**/*.feature', 'exosphere-shared/**/*.yml']
-      replace: /bus:(\s*)type: exocom(\s*)version: \d+.\d+.\d+/g
+      files: ['**/*.feature', 'exosphere-shared/**/*.yml']
+      replace: /bus:\n(\s*)type: exocom\n(\s*)version: \d+.\d+.\d+/g
       with:
         """
         bus:
-          type: exocom
-          version: #{@target-version}
+        $1type: exocom
```
kevgo, line 104: Are you sure this change is correct?

yarn.lock
```diff
@@ -0,0 +1,718 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
kevgo: Why do you add this file here?

bin/publish
charlierudolph: `$1` is the captured whitespace in front of type. I ran this function locally and it appeared correct

charlierudolph:
The top level was missing a `yarn.lock`. I'll break this into its own PR.

yarn.lock
charlierudolph: The top level was missing a yarn.lock. I'll break this into its own PR.

PR #259, opened 2017-06-02
Wait for online text in test dependencies

hugoesthere:


exosphere-shared/src/docker-helper.ls
```diff
@@ -1,24 +1,26 @@
...
       | err                              => done err
       | running-containers.includes name => done!
       | otherwise                        =>
         docker.create-container {Image, name, HostConfig}, (err, container) -> 
           | err => done err
-          container.start (err, data) ->
-            | err => done err
-            wait 2_000, done
+          container.attach {stream: true, stdout: true, stederr: true}, (err, stream) ->
+            stream = new TextStreamSearch stream
```
charlierudolph: can this move down to line 23 so its created just before its used? Could also join the two lines into one
```livescript
new TextStreamSearch(stream).wait online-text, done
```

exosphere-shared/src/docker-helper.ls
```diff
@@ -1,24 +1,26 @@
...
       | running-containers.includes name => done!
       | otherwise                        =>
         docker.create-container {Image, name, HostConfig}, (err, container) -> 
           | err => done err
-          container.start (err, data) ->
-            | err => done err
-            wait 2_000, done
+          container.attach {stream: true, stdout: true, stederr: true}, (err, stream) ->
+            stream = new TextStreamSearch stream    
+            container.start (err, data) ->
```
charlierudolph: data appears unused

charlierudolph (approved):

exosphere-shared/src/docker-helper.ls
hugoesthere: Is it possible then to miss data that's been passed through the stream between calling `container.start` and when `TextStreamSearch` is constructed in the callback of `container.start`? I figured attaching to the stream before starting the container would ensure capturing all output from the process.

charlierudolph: Auhh correct. Can we avoid overriding the `stream` variable and name this `testStreamSearch`?

hugoesthere: Yes, I like that

PR #265, opened 2017-06-07
go: simplify exoservice template

charlierudolph:
With some learnings from trying to build a service


charlierudolph:
bump please review

PR #266, opened 2017-06-07
upgrade dependency-lint

charlierudolph:
Some had their own version installed while others were using the o-tools version


exo-add/package.json
```diff
@@ -3,6 +3,7 @@
   "author": "Kevin Goslar",
   "dependencies": {
     "chalk": "1.1.3",
+    "dependency-lint": "5.0.1",
```
hugoesthere: This needs to be moved to devDependencies

hugoesthere (approved):
Thanks for doing this!

PR #268, opened 2017-06-07
Compile external service routing

hugoesthere:
Resolves: https://github.com/Originate/exosphere-sdk/issues/278


exosphere-shared/src/docker-helper.ls
```diff
@@ -50,6 +51,24 @@ class DockerHelper
...
+      docker.modem.follow-progress stream, (err, output) ->
+        | err       => done err
+        | otherwise => console.log "'#{image}' download complete"; done!
+
+
+  # Runs cat on file in Docker container to print its content
+  @cat-file = ({image, file-name}, done) ->
+    DockerHelper.pull-image {image}, ->
+      stdout-stream = new stream.PassThrough
+      text-stream-search = new TextStreamSearch stdout-stream
```
charlierudolph: This seems like an odd instance to use a TextStreamSearch. I would think something like [text-stream-accumulator](https://github.com/Originate/node-text-stream-accumulator) makes more sense here

exosphere-shared/src/docker-helper.ls
```diff
@@ -50,6 +51,24 @@ class DockerHelper
...
+      docker.run image, ['cat', file-name], stdout-stream, (err, data, container) ->
+        | err       => console.log err; done err
```
charlierudolph: Why are we logging the error here?

exosphere-shared/src/compile-service-routes.ls
```diff
@@ -17,4 +21,31 @@ module.exports = (app-config, base-path) ->
...
+        external-services.push {image: service-data['docker-image'], file-name: 'service.yml'}
+        
+  # compile service.yml in each external Docker containers
+  async.map-series external-services, (-> DockerHelper.cat-file &0, &1), (err, external-service-configs) ->
+    | err => return done null 
+
+    # compile external service routes
+    external-service-index = 0
+    for protection-level of app-config.services
+      for role, service-data of app-config.services[protection-level]
```
charlierudolph: Thoughts on adding the role and service data into `external-services` above and that way you don't need to loop through the app-config again?

exosphere-shared/src/compile-service-routes.ls
```diff
@@ -17,4 +21,31 @@ module.exports = (app-config, base-path) ->
             sends: service-config.messages.sends
             namespace: service-data.namespace
           }
-  service-routes
+
+  # compile list of external service config
+  external-services = []
+  for protection-level of app-config.services
+    for role, service-data of app-config.services[protection-level]
```
charlierudolph: Thoughts on using the following:
```livescript
for , protection-data of app-config.services
  for role, service-data of protection-data
```

exo-setup/src/app-setup.ls
```diff
@@ -27,15 +27,16 @@ class AppSetup extends EventEmitter
...
 
     @_setup-services ~>
-      @_get-dependencies-docker-config!
-      @_get-service-docker-config!
-      @_render-docker-compose!
-      @_setup-docker-images (exit-code) ~>
-        | exit-code => @write 'setup failed'; process.exit exit-code
-        @write 'setup complete'
+      @_get-dependencies-docker-config (err) ~>
+        | err => @write 'setup failed'; process.exit
```
charlierudolph: This doesn't actually call `process.exit`. Also the default exit code is zero (success). We should probably supply 1. Also thoughts on adding a test for this case where we would exit on this line? (If you agree, try to add the test before fixing this so we can see it go from red to green)

exosphere-shared/src/docker-helper.ls
hugoesthere: We shouldn't be, that's a debugging statement that should b e removed.

exo-setup/src/app-setup.ls
hugoesthere: Will update to exit with code 1. This is part of the larger setup process, which is tested in `setup.feature`,  so if this step exits then the setup process won't finish and those tests won't pass.

Although when I implemented the docker-compose stuff I did add a test diff-ing the `docker-compose.yml` file which was created, which included checking the Docker config set in this function `@_get-dependencies-docker-config`. @kevgo made the point that since the file is always auto-generated it doesn't concern the user and shouldn't be in a feature test. Thoughts on adding that back in? Or maybe unit testing that instead?

charlierudolph: I'm fine with unit testing this (especially since it would probably take a lot of setup)

hugoesthere: I'll write a ticket for that, this seems like it should be done in a separate PR

charlierudolph: I think part of this PR makes sense since it added this line. It was added incorrectly at first so I think its worth making sure we add the correct line

hugoesthere:

exo-setup/features/steps/steps-then.ls
```diff
@@ -1,8 +1,9 @@
 require! {
   'chai' : {expect}
-  '../../../exosphere-shared' : {DockerHelper}
+  '../../../exosphere-shared' : {DockerHelper, compile-service-routes}
```
charlierudolph: `compile-service-routes` doesn't appear used

exo-setup/features/steps/steps-then.ls
```diff
@@ -28,3 +29,20 @@ module.exports = ->
...
+        (service-routes.receives or= []).push message
+      for message in row.SENDS.split(', ')
+        (service-routes.sends or= []).push message
+      if row.NAMESPACE
+        service-routes.namespace = row.NAMESPACE
+      expected-routes.push service-routes
+    docker-config = yaml.safe-load fs.read-file-sync(path.join(@current-dir, 'tmp', 'docker-compose.yml'))
+    actual-routes = JSON.parse docker-config.services['exocom0.21.8'].environment.SERVICE_ROUTES
+    jsdiff-console actual-routes, expected-routes
+
```
charlierudolph: this empty line should be removed. I have an editor setting that makes sure a file ends with only a single trailing newlines

charlierudolph (approved):

PR #269, opened 2017-06-07
add buffer to exo-run file watcher

cheahuychou:
resolves #271

Make the watcher wait until the size of target file is stable before triggering a response and make the restarter wait for a few seconds before killing and restarting Docker containers.


exo-run/src/service-restarter.ls
```diff
@@ -30,19 +30,20 @@ class ServiceRestarter extends EventEmitter
...
-
-      DockerCompose.create-new-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
-        | exit-code => @emit 'error', "Docker image failed to rebuild #{@role}"
-        @write "Docker image rebuilt"
-
-        DockerCompose.start-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
-          | exit-code => @emit 'error', "Docker container failed to restart #{@role}"
-          @watch!
-          @logger.log {role: 'exo-run', text: "'#{@role}' restarted successfully"}
+    set-timeout (~>
```
hugoesthere: Do you think
```
set-timeout ( ~>
  cwd = @docker-config-location
  ....
    ....
      ....)
, 5000
```
is a bit easier to read? If not we can just keep it as it.

exo-run/src/service-restarter.ls
```diff
@@ -30,19 +30,20 @@ class ServiceRestarter extends EventEmitter
...
+        @write "Docker container stopped"
+
+        DockerCompose.create-new-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
+          | exit-code => @emit 'error', "Docker image failed to rebuild #{@role}"
+          @write "Docker image rebuilt"
+
+          DockerCompose.start-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
+            | exit-code => @emit 'error', "Docker container failed to restart #{@role}"
+            @watch!
+            @logger.log {role: 'exo-run', text: "'#{@role}' restarted successfully"}), 5000
```
hugoesthere: Let's change this to 2500 for now. We can play around with these values once we get more user feedback.

hugoesthere (approved):
Great work! 🚢

exo-run/src/service-restarter.ls
```diff
@@ -30,19 +30,21 @@ class ServiceRestarter extends EventEmitter
...
+
+        DockerCompose.create-new-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
+          | exit-code => @emit 'error', "Docker image failed to rebuild #{@role}"
+          @write "Docker image rebuilt"
+
+          DockerCompose.start-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
+            | exit-code => @emit 'error', "Docker container failed to restart #{@role}"
+            @watch!
+            @logger.log {role: 'exo-run', text: "'#{@role}' restarted successfully"})
+    , 2500
```
ben-goldberg: 2500, meaning 2.5 seconds? From experience working on an exosphere project, I'm worried this is too long. Then again, it's hard to guess what the right amount of time would be. Maybe this is fine for now and we can tune it later? 

In either case, it seems like you're intentionally using the same number here and for the `stabilityThreshold` above. Maybe we should set this to a class variable in the constructor, like `@timeDelay = 2500` or something?

cheahuychou: That was not intended. I just changed it according to @hugobho 's suggestion. I made a class variable for that but we might want to have two different variables later. I agree that 2.5 seconds is quite long.

hugoesthere: We can start with 1 second for this timeout and for the `stabilityThreshold` above? Yeah these are arbitrary values, but @ben-goldberg you're the one who's actually been using the sdk the most, let us know what you think is an appropriate wait time.

exo-run/src/service-restarter.ls
```diff
@@ -30,19 +31,21 @@ class ServiceRestarter extends EventEmitter
...
-
-      DockerCompose.create-new-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
-        | exit-code => return @emit 'error', "Docker image failed to rebuild #{@role}"
-        @write "Docker image rebuilt"
-
-        DockerCompose.start-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
-          | exit-code => return @emit 'error', "Docker container failed to restart #{@role}"
-          @watch!
-          @logger.log {role: 'exo-run', text: "'#{@role}' restarted successfully"}
+    set-timeout ( ~>
```
charlierudolph: We will still end up with multiple restarts. I think we should use a [debounced](https://www.npmjs.com/package/debounce) version of `_restart` (that we can create in the constructor) in the `watch` function

charlierudolph: I also think the delay time can drop to 100 milliseconds. We are mainly after saving multiple files at the same time.

cheahuychou: I have separated the delay variables and made the watcher use the debounced version of `_restart` but I'm still waiting for everyone to agree on actual delay times. Now that we're using debounced restart, I don't know if there is a need for this `set-timeout` function anymore. The point was to kill the watcher for the service directory and make the restarter wait for `@time-delay` and restart only once with the changes made (to all other files) during those couple seconds of the initial file change. Actually, I'm not entirely sure how this is different from debounced restart in terms of the final effect if the wait times are equal.

charlierudolph: I believe we should remove `set-timeout` from the `_restart` function now that we are using the debounced restart in the watcher.

hugoesthere: Let's stick with `debounce` and bring the delay down to 500ms, and we can tweak it from there. I think the point that @charlierudolph  is making is that with `set-timeout`, if the user makes saves outside of the 2 second window or whatever time we alot, then there will still be multiple restarts. Whereas with `debounce`, the user can make as many consecutive saves as they want just as long as they're 500ms within each other. Maybe doesn't make that big of a difference effectively but `debounce` does seem like a more idiomatic way to deal with this issue.

exo-run/src/service-restarter.ls
```diff
@@ -10,37 +11,40 @@ require! {
...
       @write "Docker container stopped"
 
       DockerCompose.create-new-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
-        | exit-code => return @emit 'error', "Docker image failed to rebuild #{@role}"
+        | exit-code => @emit 'error', "Docker image failed to rebuild #{@role}"
         @write "Docker image rebuilt"
 
         DockerCompose.start-container {service-name: @role, cwd, @env, @write}, (exit-code) ~>
-          | exit-code => return @emit 'error', "Docker container failed to restart #{@role}"
+          | exit-code => @emit 'error', "Docker container failed to restart #{@role}"
```
hugoesthere: Add `return` back in

exo-run/src/service-restarter.ls
```diff
@@ -10,37 +11,40 @@ require! {
 class ServiceRestarter extends EventEmitter
 
   ({@role, @service-location, @env, @logger}) ->
+    @stable-delay = 500
+    @debounce-delay = 1000
```
hugoesthere: Bring this down to `500`, we can play with the values later

hugoesthere (approved):

exo-run/src/service-restarter.ls
```diff
@@ -10,37 +11,40 @@ require! {
...
+        @debounced-restart!
 
 
   _restart: ->
     @watcher.close!
-    cwd = @docker-config-location
+    cwd = @docker-config-location 
     DockerCompose.kill-container {service-name: @role, cwd, @write}, (exit-code) ~>
-      | exit-code => return @emit 'error', "Docker failed to kill container #{@role}"
+      | exit-code => @emit 'error', "Docker failed to kill container #{@role}"
```
charlierudolph: Add return back in

exo-run/src/service-restarter.ls
```diff
@@ -10,37 +11,40 @@ require! {
...
       ..on 'unlink', (removed-path) ~>
         @logger.log {role: 'exo-run', text: "Restarting service '#{@role}' because #{removed-path} was deleted"}
-        @_restart!
+        @debounced-restart!
 
 
   _restart: ->
     @watcher.close!
-    cwd = @docker-config-location
+    cwd = @docker-config-location
```
charlierudolph: Please remove the trailing whitespace. Ideally your editor should trim trailing whitespace when you save files.

charlierudolph (commented):
Please sync with master. One of the updates you are missing should cause the build to fail as you added a dependency but the yarn.lock file was not updated to reflect it. Please install dependencies with `yarn` instead of `npm`

PR #275, opened 2017-06-07
ci: add yarn check

charlierudolph:
validates that our lockfiles are up to date


hugoesthere (approved):
Never knew this existed. Very useful!

PR #277, opened 2017-06-08
parse exo-run output

cheahuychou:
resolves #294

Parse exo-run output so that the logger prints out messages for individual services. For example:
```
   exo-run  Creating web
   exo-run  Creating users
   exo-run  Attaching to exocom0.21.8, web, users
    exocom  Exocom 0.21.8
    exocom  
    exocom  ExoCom HTTP service online at port 80
    exocom  ExoCom WebSocket listener online at port 80
       web  web server running at port 4000
```
However, this increases the time it takes to launch/setup the app. The tests sometimes fail due to timeout. We might need to increase the wait time or find a more efficient way to parse the output.


hugoesthere (approved):
💯

exosphere-shared/src/logger.ls
```diff
@@ -42,9 +46,32 @@ class Logger
...
 
-  _pad: (text) ->
-    "               #{text}".slice -@length
+  _parse-line: (role, line, done) ->
+    segments = [segment.trim! for segment in line / /\s+\|\s*/]
+    if segments.length == 2
+      service = @_parse-service segments[0]
+      done service, (@_reformat-line(segments[1]))
+    else
+      done role, line
```
charlierudolph: Why is this function passed a done function? It appears to be synchronous.

exosphere-shared/src/logger.ls
```diff
@@ -42,9 +46,32 @@ class Logger
   @_default_colors = [blue, cyan, magenta, yellow]
 
 
-  _pad: (text) ->
-    "               #{text}".slice -@length
+  _parse-line: (role, line, done) ->
```
charlierudolph: Can you please add some unit tests for this function? It appears to be doing a lot of complex logic.

charlierudolph:
@cheahuychou I added some unit test infrastructure to your PR and gave an example of some tests I would like to see in order to understand the code a bit better.

cheahuychou:
@charlierudolph I have written some tests. The test inputs were taken directly from the stringified output of `exo-run` so they contain the ugly ansi escape codes and strange spacing that I was trying to manipulate.

cheahuychou:
I haven't experienced any timeout errors since I created this PR so I think the timeout errors last time weren't caused by my changes. It makes sense since we're not doing anything computationally expensive here.

exosphere-shared/src/logger.spec.ls
```diff
@@ -0,0 +1,54 @@
...
+
+  describe '_parse-line', ->
+    logger = new Logger [role]
+    role = 'exo-run'
+
+    specify 'should parse non-service log message correctly' ->
+      line = 'Attaching to exocom0.21.8, web'
+      logger._parse-line role, line, (left, right) ->
+        expect(left).to.eql(role)
+        expect(right).to.eql(line)
```
charlierudolph: Can we update `_parse-line` to return an object? I'm envisioning the interface to look like the following. I think thats a better way to return multiple arguments then passing a callback.
```livescript
result = logger._parse-line role, line
expect(result.left).to.eql #...
expect(result.right).to.eql #...
```

exosphere-shared/src/logger.spec.ls
```diff
@@ -0,0 +1,54 @@
...
+
+    specify 'should parse service log message correctly' ->
+      line = '\u001b[33mweb             |\u001b[0m web server running at port 4000'
+      logger._parse-line role, line, (left, right) ->
+        expect(left).to.eql('\u001b[33mweb')
+        expect(right).to.eql('\u001b[0mweb server running at port 4000')
```
charlierudolph: Thoughts on having `_parse-line` completely strip out the colors? We can add back in our own colors after parsing.

exosphere-shared/src/logger.spec.ls
```diff
@@ -0,0 +1,54 @@
...
+        expect(left).to.eql('\u001b[36mexocom')
+        expect(right).to.eql('\u001b[0mExoCom HTTP service online at port 80')
+
+
+  describe '_pad', ->
+    role = 'exo-run'
+    logger = new Logger [role]
+
+    specify 'should pad non-styled string correctly' ->
+      expect-padded = ' ' * (logger.length - role.length) + role
```
charlierudolph: Can we place make the `expect-padded` be hardcoded instead of programmatically created? I think that makes the input / output easier to understand.

exosphere-shared/src/logger.spec.ls
hugoesthere: I'm for this - color preservation aren't a priority.

cheahuychou: This makes parsing a lot simpler. I have made the changes.

exosphere-shared/src/logger.spec.ls
```diff
@@ -0,0 +1,37 @@
...
+      expect(result.right).to.eql('ExoCom HTTP service online at port 80')
+
+
+  describe '_pad', ->
+    role = 'exo-run'
+    logger = new Logger [role]
+
+    specify 'should return a string with the correct length' ->
+      expect-padded = '   exo-run'
+      expect(logger._pad role).to.eql(expect-padded)
```
charlierudolph: Can we add more test cases for this? I'm still a bit lost at what exactly this is suppose to do. Based on the implementation I believe its suppose to pad with spaces to the left based on the maximum size of the roles passed in initially. We should probably switch to using a library for this like: https://www.npmjs.com/package/pad-left instead of how it is currently implemented.

exosphere-shared/src/logger.spec.ls
charlierudolph: This test needs to be updated to the new interface

exosphere-shared/src/logger.spec.ls
cheahuychou: You're right. Initially, I had to change the function to preserve the colors. We're no longer doing that so the function is now the same as master version. Do you still want to switch the library above?

charlierudolph: Yes. I don't like that the current implementation has a hardcoded number of spaces which I think breaks down if we have a role that is longer than those number of spaces

exosphere-shared/src/logger.spec.ls
```diff
@@ -0,0 +1,37 @@
...
+      expect(result.right).to.eql('ExoCom HTTP service online at port 80')
+
+
+  describe '_pad', ->
+    roles = ['exo-run', 'mongodb-es6']
+    logger = new Logger roles
+
+    specify 'should return a padded string of length equal to the length of the longest role' ->
+      expect-padded = '    exo-run'
+      expect(logger._pad 'exo-run').to.eql(expect-padded)
```
charlierudolph: I think the `expect-padded` variable can be removed and we just inline the expected value

charlierudolph: Lets add another expectation for what `logger._pad` should return when passed in the longest role.

charlierudolph (approved):

exosphere-shared/src/logger.spec.ls
cheahuychou: @charlierudolph I've added that line.

charlierudolph: Looks good!

PR #280, opened 2017-06-08
migrate to cucumber 2

cheahuychou:
Update `exo-` sub-projects to use cucumber 2 so that we can see useful error messages.

### DIFF
--- a/cli/features/support/env.ls
+++ b/cli/features/support/env.ls
@@ -1,16 +1,19 @@
 require! {
+  'cucumber': {defineSupportCode}
+  './world': World
   'child_process'
   '../../../exosphere-shared' : {kill-child-processes, DockerHelper}
   'fs-extra' : fs
 }
 
 
-module.exports = ->
+defineSupportCode ({After, set-default-timeout, set-world-constructor})
 
-  @set-default-timeout 2000
+  set-default-timeout 2000
+  set-world-constructor World
 
 
-  @After tags: ['~@e2e'], (scenario, done) ->
+  After tags: ['~@e2e'], (scenario, done) ->
     if @app-dir
       fs.remove-sync @app-dir
     kill-child-processes ->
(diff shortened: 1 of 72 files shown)


cli/features/support/env.ls
```diff
@@ -1,22 +1,26 @@
...
-module.exports = ->
+defineSupportCode ({After, set-default-timeout, set-world-constructor}) ->
 
-  @set-default-timeout 2000
+  set-default-timeout 2000
+  set-world-constructor World
 
 
-  @After tags: ['~@e2e'], (scenario, done) ->
+  After tags: '~@e2e', (scenario, done) ->
```
charlierudolph: The tag syntax changed in cucumber 2.0. `~@e2e` should become `not @e2e`. Same for `~@docker-cleanup` below

exo-add/features/steps/steps-then.ls
```diff
@@ -1,15 +1,17 @@
 require! {
   'chai' : {expect}
+  'cucumber': {defineSupportCode}
+
```
charlierudolph: please remove this newline

exo-clone/features/steps/steps-when.ls
```diff
@@ -1,18 +1,20 @@
 require! {
   'chai' : {expect}
+  'cucumber': {defineSupportCode}
+
```
charlierudolph: Please remove this newline

charlierudolph (commented):
Please also remove the use of the `--strict` flag. This was removed in cucumber 2.0 as strict is now the default

charlierudolph (approved):
Thanks a lot for doing this! Good to ship once the comments are addressed if the build stays green

exo-clone/features/steps/steps-when.ls
hugoesthere: This comment was undressed

exo-sync/features/steps/steps-given.ls
```diff
@@ -1,24 +1,25 @@
-require! {
+require! {
```
hugoesthere: Remove trailing whitespace

hugoesthere (approved):
Awesome, thanks! 🚀 once tests are green

PR #281, opened 2017-06-08
Update nats bus config

hugoesthere:
`get-docker-config` uses callbacks now, forgot to update this method when Exocom's driver was switched.


charlierudolph:
Should we add a feature test for nats?

hugoesthere:
@charlierudolph added feature tests for app dependencies, which also included adding some extra logic to `app-runner.ls` to check for a dependency's online text.

charlierudolph (approved):

PR #282, opened 2017-06-09
Update CHANGELOG.md

charlierudolph:
We need to keep this up to date in order for it to be worth anything. We could potentially add a step to the publish script that checks that the version you want to publish appears in the CHANGELOG.md as a validation that it has been updated


hugoesthere (approved):
+1 to adding check in publish script

PR #285, opened 2017-06-10
port `exo-clean` to go

charlierudolph:
Ported this command since it is the smallest.

Still todo: 
* ~~update the CLI tests to use the go binary~~
* ~~get the tests passing on travis~~
* ~~add travis deploy script to compile the binaries and add to the release (use git-town as an example)~~
* ~~add a yarn install script to pull in the OS specific binary and put it in the proper place~~


charlierudolph:
Okay. This is now ready for review. Notes:
* As part of the install script it will download the operating system specific exo-clean binary and put it in the proper place.
* The go dependencies are installed on the top level so each command can share them as needed.
* When another subcommand is ported to go, the following updates need to occur:
  * rewrite subcommand in `go`
    * in the subcommand directory, `bin/setup` should output the built command to `bin/<command>`
  * update `cli/src/cli.ls` for which commands are go 
  * add folder to the `bin/cross-compile` script
  * add line in `bin/fetch-go-binaries` script

The last bullet should probably be added to the documentation somewhere. Any ideas where to put it?

package.json
```diff
@@ -1,20 +1,19 @@
 {
   "name": "exosphere-sdk",
-  "version": "0.22.1",
+  "version": "0.22.2-test.1",
```
charlierudolph, line 3: I'll revert this after this has been reviewed. You can checkout this branch and test out the script that fetches the go binary for yourself.

bin/cross-compile
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -ex
+
+
+# This script cross-compiles
+# the Git Town source code
```
kevgo, line 6: What is our friend Git Town doing here?

bin/fetch-go-binaries
```diff
@@ -0,0 +1,58 @@
...
+    case 'freebsd':
+      return 'freebsd'
+    case 'linux':
+      return 'linux'
+    case 'openbsd':
+      return 'openbsd'
+    case 'win32':
+      return 'windows'
+    default:
+      throw new Error('Unsupported operating system. Please open an issue with your operating system and system architecture.')
```
kevgo, line 22: Let's print the value of `process.platform` here, so that we get good bug reports from users.

bin/fetch-go-binaries
```diff
@@ -0,0 +1,58 @@
...
+function getGoBinaryArchitecure() {
+  switch(process.arch) {
+    case 'arm':
+      return 'arm'
+    case 'i32':
+      return '386'
+    case 'x64':
+      return 'amd64'
+    default:
+      throw new Error('Unsupported system architecture system. Please open an issue with your operating system and system architecture.')
```
kevgo, line 35: Print the value of `process.arch` here.

bin/fetch-go-binaries
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+const DOWNLOADS_BASE_URL = `https://github.com/Originate/exosphere/releases/download/v${version}/`
+const GO_BINARY_SUFFIX = `-${getGoBinaryOS()}-${getGoBinaryArchitecure()}`
```
kevgo, line 40: Instead of definining these two constants, create a `getBinaryPath` function that returns the full path of the binary?

bin/fetch-go-binaries
```diff
@@ -0,0 +1,58 @@
...
+  var fileStream = fs.createWriteStream(filePath)
+  var url = `${DOWNLOADS_BASE_URL}${command}${GO_BINARY_SUFFIX}`
+  console.log(`Fetching '${command}' from ${url}`)
+  request.get(url)
+    .on('error', (err) => {
+      console.error(`Failed to fetch '${command}'. Error: ${err.toString()}`)
+    })
+    .pipe(fileStream)
+    .on('finish', () => {
+      fs.chmodSync(filePath,  0o777)
```
kevgo, line 54: No need to make this world writable. Let's go with 744 here.

cli/package.json
```diff
@@ -7,6 +7,7 @@
   "dependencies": {
     "abbrev": "1.1.0",
     "chalk": "1.1.3",
+    "cross-spawn": "^5.1.0",
```
kevgo, line 10: no `^`

bin/fetch-go-binaries
```diff
@@ -0,0 +1,58 @@
...
+      return 'amd64'
+    default:
+      throw new Error('Unsupported system architecture system. Please open an issue with your operating system and system architecture.')
+  }
+}
+
+const DOWNLOADS_BASE_URL = `https://github.com/Originate/exosphere/releases/download/v${version}/`
+const GO_BINARY_SUFFIX = `-${getGoBinaryOS()}-${getGoBinaryArchitecure()}`
+
+function fetchGoBinary(command) {
```
kevgo, line 42: What do yo think about bundling all the Go binaries for all platforms with the NPM module? Then decide at runtime wich one to execute. This could save us a lot of downloading logic.

exo-clean/.gitignore
```diff
@@ -1,3 +1,2 @@
-dist/
-node_modules/
 tmp/
+bin/exo-clean
```
trushton, line 2: We should be ignoring the `vendor` directory here too

bin/cross-compile
trushton, line 6: `the exosphere source code`

trushton (approved):

kevgo (commented):
Some preliminary comments

exo-clean/bin/features
```diff
@@ -1,10 +0,0 @@
-#!/usr/bin/env bash
```
kevgo, line 1: I would like to keep a script that allows to run a single feature without running the linters first.

exo-clean/main_test.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		err = startAndRemoveContainer(dockerClient, imageName)
+		if err != nil {
+			return fmt.Errorf("Error starting and removing container up: %v", err)
+		}
+		return nil
+	})
+
+	s.Step(`^running "([^"]*)" in the terminal$`, func(argumentsStr string) error {
+		argumentsStr = strings.Replace(argumentsStr, "exo-clean", path.Join(cwd, "bin", "exo-clean"), 1)
```
kevgo, line 117: Maybe replace `exo-` here, instead of `exo-clean` to support other exo commands in the future?

exo-clean/test_helpers/copy_directory.go
```diff
@@ -0,0 +1,113 @@
...
+	"os"
+	"path/filepath"
+)
+
+// CopyFile copies the contents of the file named src to the file named
+// by dst. The file will be created if it does not already exist. If the
+// destination file exists, all it's contents will be replaced by the contents
+// of the source file. The file mode will be copied from the source and
+// the copied data is synced/flushed to stable storage.
+func CopyFile(src, dst string) (err error) {
```
kevgo, line 18: Does Go really not implement a file copy method?? I read through the newsgroups, and their reasoning is ridiculous.

kevgo (approved):
Looks good to me. We can also build the Go version as a single Cobra monolith. The break-up into dedicated commands didn't hold up, since the CI server always has to run all commands.

charlierudolph:
@mcharawi. There were a lot of unaddressed comments in this PR. Why was it merged?

kevgo:
@charlierudolph we decided to address them in separate PRs to keep momentum

charlierudolph:
Is there an issue for addressing these comments? Now that I'm back, I'm happy to tackle them.

kevgo:
We didn't create issues. There are only a few comments in here, they can be addressed directly. Moving forward, let's create issues again. This was a special case due to delays and change in developers.

PR #286, opened 2017-06-10
cleanup: remove use of cucumber-js '--strict'

charlierudolph:
After updating to cucumber 2.0, this option is no longer valid.

I mentioned this in #280 but my comment was missed.


cheahuychou (approved):
Thanks for fixing this. I thought `--strict` only appeared in `cucumber.js` files as shown in the diff.

hugoesthere (approved):

PR #291, opened 2017-06-12
allow users to hide logs

cheahuychou:
Depends on #277

Allow users to hide logs from a service or dependency (i.e. exocom) by adding `silent: true` in `application.yml` like so:

```
dependencies:
  -  type: exocom
     version: 0.22.1
     silent: true

services:
  public:
    web:
      location: ./web-server
      silent: true
```


exo-run/src/cli.ls
```diff
@@ -16,7 +16,10 @@ module.exports = ->
 
   app-config = yaml.safe-load fs.read-file-sync('application.yml', 'utf8')
   console.log "Running #{green app-config.name} #{cyan app-config.version}\n"
-  logger = new Logger flatten [Object.keys(app-config.services[type]) for type of app-config.services]
+  services = flatten [Object.keys(app-config.services[type]) for type of app-config.services]
+  silenced-services = [service if app-config.services[type][service].silent for type of app-config.services for service of app-config.services[type]]
```
hugoesthere: It might be worth doing a single nested for loop to build `services` and `silenced-services`, makes it a bit easier to understand what's going on.

exo-setup/src/cli.ls
```diff
@@ -14,7 +14,10 @@ module.exports = ->
 
   app-config = yaml.safe-load fs.read-file-sync('application.yml', 'utf8')
   console.log "Setting up #{green app-config.name} #{cyan app-config.version}\n"
-  logger = new Logger flatten [Object.keys(app-config.services[type]) if app-config.services[type] for type of app-config.services]
+  services = flatten [Object.keys(app-config.services[type]) if app-config.services[type] for type of app-config.services]
+  silenced-services = [service if app-config.services[type][service].silent for type of app-config.services for service of app-config.services[type]]
```
hugoesthere: Same here, might make more sense in a single nested loop.

charlierudolph:
Please add some tests for this. I'm thinking some feature tests where we have a service that sends some output and verify that by default it gets printed and when silenced it does not.

exosphere-shared/src/logger.ls
```diff
@@ -23,16 +23,18 @@ class Logger
     text = text.trim! if trim
     for line in text.split '\n'
       parsed-line = @_parse-line role, line
-      color = @colors[parsed-line.left] ? reset
-      console.log color("#{bold "#{@_pad "#{parsed-line.left}"} "} #{parsed-line.right}")
+      if (@silenced-roles.indexOf parsed-line.left) == -1
```
hugoesthere: What about this? 
```
if !@silenced-roles.includes parsed-line.left
  ...
```

hugoesthere (approved):

cheahuychou:
Reopening, the PR somehow got closed after #277 was merged.

cheahuychou:
@charlierudolph  I have added the feature tests.

exo-run/features/run.feature
```diff
@@ -54,3 +54,13 @@ Feature: running Exosphere applications
   Scenario: a service crashes during startup
     When trying to run the "crashing-service" application
     Then it prints "crasher exited with code 1" in the terminal
+
+
+  Scenario: hiding logs from certain service and dependency
```
hugoesthere: `hiding logs from certain services or dependencies`

exo-run/features/run.feature
```diff
@@ -54,3 +54,13 @@ Feature: running Exosphere applications
...
+    Given a running "silenced-running" application
+    Then my machine is running the services:
+      | NAME  |
+      | web   |
+      | users |
```
hugoesthere: Let's list `exocom#{some-version}` here as well, since we're testing that it doesn't print logs for exocom.

exo-run/features/run.feature
```diff
@@ -54,3 +54,13 @@ Feature: running Exosphere applications
...
+    And it does not print "'users' is running" in the terminal
+    And it does not print "'exocom' is running" in the terminal
```
hugoesthere: Can we check that online text for the web service is printed? So we know the logger doesn't just hide all output text.

exo-run/features/run.feature
hugoesthere: You might also have to add a step before line 65 that checks that 'all services online' is printed. That way we're not checking for the absence of text before the application has been fully booted up.

exo-setup/features/setup.feature
```diff
@@ -39,3 +39,8 @@ Feature: Setup of Exosphere applications
     And ExoCom uses this routing:
       | ROLE             | SENDS                       | RECEIVES                    | NAMESPACE |
       | external-service | users.list, users.create    | users.listed, users.created |           |
+
+  Scenario: set up the application with silenced services and dependencies
```
hugoesthere: Didn't realize that output was being silenced for `exo-setup` as well. I think that can be a potential source of confusion for the user. For now we should always log output when setting up the application and only allow the option to hide output during `exo-run`.

But it's good to know that if we do want this feature in the future it's really easy to implement.

exo-setup/features/steps/steps-then.ls
```diff
@@ -46,3 +46,7 @@ defineSupportCode ({Then}) ->
     docker-config = yaml.safe-load fs.read-file-sync(path.join(@current-dir, 'tmp', 'docker-compose.yml'))
     actual-routes = JSON.parse docker-config.services['exocom0.22.1'].environment.SERVICE_ROUTES
     jsdiff-console actual-routes, expected-routes
+
+
+  Then /^it does not print "([^"]*)" in the terminal$/ (expected-text) ->
+    expect(@process.full-output!).to.not.contain expected-text
```
hugoesthere: See above, this change will be reverted.

exo-setup/src/cli.ls
```diff
@@ -14,10 +14,19 @@ module.exports = ->
...
-  logger = new Logger flatten [Object.keys(app-config.services[type]) if app-config.services[type] for type of app-config.services]
+  services = []
+  silenced-services = []
+  for type of app-config.services
+    for service of app-config.services[type]
+      services.push service
+      if app-config.services[type][service].silent
+        silenced-services.push service
+  silenced-dependencies = [dependency.type if dependency.silent for dependency in app-config.dependencies]
+  logger = new Logger services, silenced-services ++ silenced-dependencies
```
hugoesthere: See above, this change will be reverted.

hugoesthere (approved):
Awesome! Good to go after comments are addressed and tests are green.

PR #292, opened 2017-06-14
unit test docker-compose.yml generation

cheahuychou:
Add unit tests for the code in `exo-setup` that generates `docker-compose.yml`


exosphere-shared/src/docker-helper.ls
```diff
@@ -83,18 +83,6 @@ class DockerHelper
...
-    DockerHelper.pull-image {image}, (err) ->
-      | err       => done err
-      | otherwise =>
-        stdout-stream = new stream.PassThrough
-        text-stream-search = new TextStreamAccumulator stdout-stream
-        docker.run image, ['cat', file-name], stdout-stream, (err, data, container) ->
-          | err       => done err
-          | otherwise => done null, text-stream-search.to-string!
-
-
```
hugoesthere, line 97: I think this was removed when resolving merge conflicts somewhere. Please add it back in :)

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,123 @@
...
+
+
+describe 'AppSetup', ->
+
+  describe 'app with external dependencies' (...) ->
+    @timeout 600000
+
+    before (done) ~>
+      @app-name = 'external-dependency'
+      @app-dir = path.join process.cwd!, 'test', @app-name
```
hugoesthere: Let's build the app in `tmp` like we do with the rest of the tests.

exo-setup/.gitignore
```diff
@@ -1,3 +1,4 @@
 dist/
 node_modules/
 tmp/
+test/*/
```
hugoesthere: This can be removed once comment below has been addressed.

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,123 @@
...
+}
+
+
+describe 'AppSetup', ->
+
+  describe 'app with external dependencies' (...) ->
+    @timeout 600000
+
+    before (done) ~>
+      @app-name = 'external-dependency'
```
hugoesthere: Let's test against a more complex app so we know that all edge cases are covered, like the todo app that's generated in `tutorial.feature`. I will send you an example.

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,123 @@
...
+      @app-dir = path.join process.cwd!, 'test', @app-name
+      @docker-compose-location = path.join @app-dir, 'tmp', 'docker-compose.yml'
+      @exocom-name = 'exocom0.22.1'
+      @service-name = 'mongo'
+      @mongo-name = 'mongo3.4.0' # external dependency
+      @process = checkout-setup-app @app-name, @app-dir
+        ..on 'ended' done
+
+    specify 'should create docker-compose.yml at the expected location', (done) ~>
+      fs.exists @docker-compose-location, (exists) ~>
```
hugoesthere: `fs.exists` is depricated, please use `fs.stat` or `fs.access`: https://nodejs.org/api/fs.html#fs_fs_exists_path_callback

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,123 @@
...
+        expect(exists).to.be.true
+        @docker-compose = yaml.safe-load fs.read-file-sync(@docker-compose-location, 'utf8')
+        done!
+
+    specify 'should list all services and dependencies under \'services\'', (done) ~>
+      expect(Object.keys @docker-compose.services).to.eql([@exocom-name, @service-name, @mongo-name])
+      done!
+
+    specify 'should generate an image name for each dependency', (done) ~>
```
hugoesthere: Once we switch to the more complex example app: `'should generate an image name for each dependency and external service'`

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,123 @@
...
+      expect(@docker-compose.services[@exocom-name].image).to.not.be.empty
```
hugoesthere: Let's use `.to.exist` if that works, that makes this a little more readable.

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,123 @@
...
+      expect(@docker-compose.services[@mongo-name].image).to.not.be.empty
+      done!
+
+    specify 'should generate a container name for each service/dependency equal to the service/dependency name', (done) ~>
```
hugoesthere: Just `should generate a container name for each service and dependency` should be fine. Then we can check for equality in the steps below like you're already doing.

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,123 @@
...
+      expect(@docker-compose.services[@service-name].container_name).to.eql(@service-name)
+      expect(@docker-compose.services[@mongo-name].container_name).to.eql(@mongo-name)
+      expect(@docker-compose.services[@exocom-name].container_name).to.eql(@exocom-name)
+      done!
+
+    specify 'should have the correct command for each service/dependency that can be built', (done) ~>
```
hugoesthere: `should have the correct build command for each service and dependency`

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,123 @@
...
+      expect(@docker-compose.services[@service-name].command).to.eql('node_modules/exoservice/bin/exo-js')
+      expect(@docker-compose.services[@exocom-name].command).to.eql('bin/exocom')
+      done!
+
+    specify 'should have correct build path for each service', (done) ~>
```
hugoesthere: `should have the correct build path for each internal service`

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,123 @@
...
+      expect(mongo-env.EXOCOM_HOST).to.eql(@exocom-name)
+      expect(mongo-env.EXOCOM_PORT).to.eql('$EXOCOM_PORT')
+      done!
+
+    specify 'should generate a volume path for an external dependency that mounts a volume', (done) ~>
+      expect(@docker-compose.services[@mongo-name].volumes).to.not.be.empty
+      done!
+
+  # indirectly tests @cat-file used in exosphere-shared/src/compile-service-routes.ls
+  describe 'app with external docker images' (...) ->
```
hugoesthere: Once we switch to the more complex example app, this test case will be covered and can be removed. We can test `@cat-file` in another PR if we decide to do so.

hugoesthere (commented):
Great work! Thanks for doing this - this is a critical part of our sdk and in dire need of testing.

exosphere-shared/src/docker-helper.ls
cheahuychou: There is already a `@cat-file` on line 66-75. They're the same I think.

hugoesthere: Ah, I was wondering how tests were still passing. Great, thanks for noticing this!

cheahuychou:
@hugobho all comments addressed. I chose to use `to.not.be.empty` because it checks that the string is non-empty and I think `to.exist` checks that it is not `undefined` and `null`.  I can change it if you want.

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,102 @@
...
+      @process = checkout-setup-app @app-name, @app-dir
+        ..on 'ended' done
+
+    specify 'should create docker-compose.yml at the expected location', (done) ~>
+      fs.stat @docker-compose-location, (err, stat) ~>
+        expect(err).to.be.null
+        @docker-compose = yaml.safe-load fs.read-file-sync(@docker-compose-location, 'utf8')
+        done!
+
+    specify 'should list all services and dependencies under \'services\'', (done) ~>
```
hugoesthere: This looks like a synchronous function, no need to provide a callback.

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,102 @@
...
+      all-services = [@exocom-name] ++ @internal-services ++ @external-dependencies ++ @external-services
+      expect(Object.keys @docker-compose.services).to.have.members(all-services)
+      done!
+
+    specify 'should generate an image name for each dependency and external service', (done) ~>
+      expect(@docker-compose.services[@exocom-name].image).to.not.be.empty
+      @external-dependencies.for-each (dependency) ~> expect(@docker-compose.services[dependency].image).to.not.be.empty
```
hugoesthere: Cool syntax!

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,102 @@
...
+        expect(err).to.be.null
+        @docker-compose = yaml.safe-load fs.read-file-sync(@docker-compose-location, 'utf8')
+        done!
+
+    specify 'should list all services and dependencies under \'services\'', (done) ~>
+      all-services = [@exocom-name] ++ @internal-services ++ @external-dependencies ++ @external-services
+      expect(Object.keys @docker-compose.services).to.have.members(all-services)
+      done!
+
+    specify 'should generate an image name for each dependency and external service', (done) ~>
```
hugoesthere: Please remove callback here and in all other synchronous methods.

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,102 @@
...
+      @internal-services.for-each (service) ~> expect(@docker-compose.services[service].container_name).to.not.be.empty
+      done!
+
+    specify 'should have the correct build command for each service and dependency', (done) ~>
+      @internal-services.for-each (service) ~> expect(@docker-compose.services[service].command).to.eql('echo "does not run"')
+      expect(@docker-compose.services[@exocom-name].command).to.eql('bin/exocom')
+      done!
+
+    specify 'should have the correct build path for each internal service', (done) ~>
+      @internal-services.for-each (service) ~> expect(@docker-compose.services[service].build).to.eql("../#service")
```
hugoesthere: Should this not be `to.eql.("../#{service}")`? How is this step passing?

exo-setup/src/app-setup.spec.ls
```diff
@@ -0,0 +1,102 @@
...
+describe 'AppSetup', ->
+
+  describe 'set up a complex app' (...) ->
+    @timeout 600000
+
+    before (done) ~>
+      @app-name = 'complex-setup-app'
+      @app-dir = path.join process.cwd!, 'tmp', @app-name
+      @docker-compose-location = path.join @app-dir, 'tmp', 'docker-compose.yml'
+      @exocom-name = 'exocom0.22.1'
```
hugoesthere, line 22: Let's make this an array called `@application-dependencies` to stay consistent with the other variables.

hugoesthere (approved):
Looks good! Yes, I think `to.not.be.empty` makes more sense here. 

Once you go through and remove the callbacks from the synchronous functions this should be good to go.

exo-setup/src/app-setup.spec.ls
cheahuychou: Just tested it, they return the same string. `"../#service" == "../#{service}"` returns `true`.

exo-setup/src/app-setup.spec.ls
cheahuychou: Without `done`, all the tests will run before `before` finishes and we need `before` to do the setup.

exo-setup/src/app-setup.spec.ls
hugoesthere: Hm, I just ran it through the lsc compiler too, didn't know you could do this. Good to know these tests weren't returning a false positive though.

exo-setup/src/app-setup.spec.ls
hugoesthere: Even though the `before` method is asynchronous? This should be the expected behavior: https://stackoverflow.com/questions/24723374/async-function-in-mocha-before-is-alway-finished-before-it-spec

hugoesthere: Actually, that link isn't that helpful. They have a `done` callback in their test steps as well.

hugoesthere: What happens if the nest the `specify` steps in another `describe` block after the `before` method?

cheahuychou: That works but still need pass in `done` into `before`.

exo-setup/src/app-setup.spec.ls
cheahuychou: I named it `@internal-dependencies`, hope that's ok.

PR #296, opened 2017-06-16
make the publish use exocom latest version instead

cheahuychou:
resolves: https://github.com/Originate/exosphere-sdk/issues/303

update the publish script of exosphere to have the version of exocom be the latest release of exocom instead of being the same version of exosphere we want to publish

### DIFF
--- a/bin/publish
+++ b/bin/publish
@@ -2,6 +2,7 @@
 
 require! {
   'chalk' : {bold, red, cyan, yellow, green}
+  'child_process'
   'inquirer'
   'path'
   'replace-in-file' : replace
@@ -66,9 +67,7 @@ function check-npm-dependencies
 
 
 function get-exocom-version
-  process = new ObservableProcess('npm view exocom-dev version')
-    ..on 'ended', ~>
-      process.full-output!
+  (child_process.exec-sync 'npm view exocom-dev version', {encoding: 'utf-8'}).trim!
 
 
 function update-exocom-dependencies
--- a/bin/publish
+++ b/bin/publish
@@ -45,6 +45,7 @@ function main
   check-npm-dependencies!
 
   # update Exocom version / ensure tests pass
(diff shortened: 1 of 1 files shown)


bin/publish
```diff
@@ -64,6 +65,12 @@ function check-npm-dependencies
   console.log!
 
 
+function get-exocom-version
```
hugoesthere, line 69: This is an asynchronous function that's being called synchronously - we're waiting for the 'ended' event before returning the result. On second thought, let's just use `child_process.exec-sync` (the synchronous version of `exec`): https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options. Especially since we don't care about monitoring the logs of this process.

Sorry about all the confusion with `ObservableProcess`. Feel free to use whatever library you see fit, we use `ObservableProcess` when we need to monitor (observe) the output of a process, like when waiting for certain strings to be logged. By no means it is the only way to run processes. Since we only care about the output here we can use `exec-sync`.

hugoesthere (approved):
Looks good! Just need to make this method synchronous.

Also, please link to the `exosphere-sdk` ticket in the future so we can keep track of which PRs are open for which tickets :)

Resolves: https://github.com/Originate/exosphere-sdk/issues/303

kevgo:
Let's not blindly use the latest possible version of Exocom. If we do that, and ever publish a breaking change for Exocom, all Exosphere apps break the next time they are deployed. Please use the Exocom version specified in `application.yml` of the application that is being deployed.

hugoesthere:
@kevgo This is the publish script for the `exosphere-sdk` npm package, not deployment of an application. I'm open to specifying an Exocom version, but we shouldn't reading from any sort of `application.yml` file.

kevgo:
Haha, okay, in this case it's okay to use the latest version of Exocom! Good call!

hugoesthere (approved):

PR #298, opened 2017-06-19
reformat dependencies section of application.yml  and service.yml

cheahuychou:
Make the dependencies section of `application.yml` and `service.yml` have the same format to make reusing dependency definitions (https://github.com/Originate/exosphere-sdk/issues/284)  easier. So the dependencies section now looks like:
```
dependencies:
  - mongo:
    name: 'mongo'
    version: '3.4.0'
    config:
      volumes:
        - '{{EXO_DATA_PATH}}:/data/db'
      ports:
        - '27017:27017'
```
instead of
```
dependencies:
  - type: exocom
    version: 0.22.1
``` 
and 
```
dependencies:
  mongo:
    dev:
      image: 'mongo'
      version: '3.4.0'
      volumes:
        - '{{EXO_DATA_PATH}}:/data/db'
      ports:
        - '27017:27017'
```

### DIFF
--- a/cli/features/tutorial.feature
+++ b/cli/features/tutorial.feature
@@ -162,8 +162,7 @@ Feature: Following the tutorial
           - todo.updated
 
       dependencies:
-        - mongo:
-          name: 'mongo'
+        - name: 'mongo'
           version: '3.4.0'
           config:
             volumes:
--- a/cli/features/tutorial.feature
+++ b/cli/features/tutorial.feature
@@ -37,7 +37,7 @@ Feature: Following the tutorial
       version: 0.0.1
 
       dependencies:
-        - type: exocom
+        - name: exocom
           version: 0.22.1
 
       services:
@@ -65,7 +65,7 @@ Feature: Following the tutorial
       version: 0.0.1
 
       dependencies:
-        - type: exocom
+        - name: exocom
           version: 0.22.1
 
       services:
@@ -162,10 +162,10 @@ Feature: Following the tutorial
           - todo.updated
 
       dependencies:
-        mongo:
-          dev:
-            image: 'mongo'
-            version: '3.4.0'
+        - mongo:
+          name: 'mongo'
+          version: '3.4.0'
+          config:
             volumes:
               - '{{EXO_DATA_PATH}}:/data/db'
             ports:
--- a/exo-setup/src/docker-setup.ls
+++ b/exo-setup/src/docker-setup.ls
@@ -34,8 +34,8 @@ class DockerSetup
       depends_on: @_get-service-dependencies!
     if @service-config.dependencies
       for dependency in @service-config.dependencies
-        console.log dependency
-        docker-config[dependency.name + dependency.version] = @_get-service-dependency-docker-config dependency.name, dependency.config
+        console.log 'dependency' dependency
+        docker-config[dependency.name + dependency.version] = @_get-service-dependency-docker-config dependency.name, dependency.version, dependency.config
     docker-config
 
 
@@ -43,7 +43,7 @@ class DockerSetup
   # returns undefined if length is 0 so it can be ignored with Obj.compact
   _get-docker-links: ->
     links = []
-    console.log @service-config.dependencies
+    # console.log @service-config.dependencies
     if @service-config.dependencies
       for dependency in @service-config.dependencies
         links.push "#{dependency.name + dependency.version}:#{dependency.name}"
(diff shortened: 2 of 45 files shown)


cli/features/tutorial.feature
```diff
@@ -162,10 +162,10 @@ Feature: Following the tutorial
           - todo.updated
 
       dependencies:
-        mongo:
-          dev:
-            image: 'mongo'
-            version: '3.4.0'
+        - mongo:
```
hugoesthere: Can we keep this in the same format that it is in `application.yml`?
```
dependencies:
- name: 'mongo'
  version: '3.4.0'
  config:
     ...
```
Looks like you have that format in a couple other places as well, let's stick with the one above.

exo-setup/src/docker-setup.ls
```diff
@@ -32,17 +32,19 @@ class DockerSetup
       links: @_get-docker-links!
       environment: @_get-docker-env-vars!
       depends_on: @_get-service-dependencies!
-    for dependency, dependency-config of @service-config.dependencies
-      docker-config[dependency + dependency-config.dev.version] = @_get-service-dependency-docker-config dependency, dependency-config.dev
+    if @service-config.dependencies
```
hugoesthere: Is this conditional necessary? The for loop won't execute if there are no dependencies right?

exo-setup/src/docker-setup.ls
```diff
@@ -32,17 +32,19 @@ class DockerSetup
...
     docker-config
 
 
   # compiles list of container links from service to dependency
   # returns undefined if length is 0 so it can be ignored with Obj.compact
   _get-docker-links: ->
     links = []
-    for dependency, dependency-config of @service-config.dependencies
-      links.push "#{dependency + dependency-config.dev.version}:#{dependency}"
+    if @service-config.dependencies
```
hugoesthere: Same as above.

exo-setup/src/docker-setup.ls
```diff
@@ -53,24 +55,26 @@ class DockerSetup
     for dependency-config in @app-config.dependencies
       dependency = ApplicationDependency.build dependency-config
       env-vars = {...env-vars, ...dependency.get-service-env-variables!}
-    for dependency of @service-config.dependencies
-      env-vars[dependency.to-upper-case!] = dependency
+    if @service-config.dependencies
```
hugoesthere: Same as above.

exo-setup/src/docker-setup.ls
```diff
@@ -53,24 +55,26 @@ class DockerSetup
...
+        env-vars[dependency.name.to-upper-case!] = dependency.name
     env-vars
 
 
   # compiles list of names of dependencies a service relies on
   _get-service-dependencies: ->
     dependencies = @_get-app-dependencies!
-    for dependency, dependency-config of @service-config.dependencies
-      dependencies.push "#{dependency}#{dependency-config.dev.version}"
+    if @service-config.dependencies
```
hugoesthere: Same as above.

cli/features/tutorial.feature
cheahuychou: Yes, that was a mistake. I'll change it.

exo-setup/src/docker-setup.ls
cheahuychou: I think without that if statement, it returns `Cannot read property 'length' of undefined`

cheahuychou:

exo-setup/src/docker-setup.ls
hugoesthere: What if we do 
```
for dependency in @service-config.dependencies or {}
```
What do you think? I feel like it reads a little nicer.

cheahuychou: This is an array so should be `for dependency in @service-config.dependencies or []`. I think that works.

exo-test/src/service-tester.ls
```diff
@@ -44,21 +44,22 @@ class ServiceTester extends EventEmitter
...
     @dependencies = []
-    for dependency-name, dependency-config of @service-config.dependencies
-      if dependency-config.dev
-        @dependencies.push do
-          Image: "#{that.image}:#{that.version}"
-          name: "#{@role}-test-#{dependency-name}"
-          HostConfig: @_get-port-mapping that 
-          online-text: that['online-text']
+    if @service-config.dependencies
+      for dependency in @service-config.dependencies
```
hugoesthere: Let's do `for dependency in @service-config.dependencies or []` here also

hugoesthere (approved):
👍

PR #299, opened 2017-06-19
reuse dependency definitions

cheahuychou:
resolves: https://github.com/Originate/exosphere-sdk/issues/284

Allow users to reuse dependency definitions. For example, `application.yml` could contain
```
dependencies:
  - name: exocom
    version: 0.22.1
  - name: mongo
    version: 3.4.0
    config:
      ports:
        - '5000:5000'
      volumes:
        - '{{EXO_DATA_PATH}}:user/data/db'
      dependency-environment:
        DEPENDENCY_VAR: 'test-db'
      service-environment:
        SERVICE_VAR: 'test-collection'
```
and the `service.yml` of the services that use `mongo` can just specify:
```
dependencies:
  - name: mongo
    version: 3.4.0
```


exo-setup/features/setup.feature
```diff
@@ -39,3 +39,70 @@ Feature: Setup of Exosphere applications
     And ExoCom uses this routing:
       | ROLE             | SENDS                       | RECEIVES                    | NAMESPACE |
       | external-service | users.list, users.create    | users.listed, users.created |           |
+
+  Scenario: set up an application with services shared dependencies
```
hugoesthere: `setting up an application with shared service dependencies`

hugoesthere: I think this test should go in `exosphere-shared` where `docker-compose.yml` is unit tested. This is a cucumber test which is related to more of what the user should be concerned about. Since the `docker-compose.yml` file shouldn't be modified by them anyways (hence why it lives in `tmp/`), then we shouldn't be testing it via cucumber.

exo-setup/features/steps/steps-then.ls
```diff
@@ -21,6 +22,13 @@ defineSupportCode ({Then}) ->
       fs.access-sync path.join(@current-dir, row[0])
 
 
+  Then /^my application contains the file "([^"]*)" with the content:$/, (file-path, expected-fragment, done) ->
+    console.log @current-dir
+    fs.read-file path.join(@current-dir, file-path), N (actual-content) ->
+      expect(actual-content.to-string!).to.contain expected-fragment.trim!
+      done!
+
+
```
hugoesthere: This will all be removed once we switch to unit tests.

exo-setup/src/docker-setup.ls
```diff
@@ -34,7 +34,8 @@ class DockerSetup
       depends_on: @_get-service-dependencies!
     if @service-config.dependencies
       for dependency in @service-config.dependencies
-        docker-config[dependency.name + dependency.version] = @_get-service-dependency-docker-config dependency.name, dependency.version, dependency.config
+        if dependency.config
```
hugoesthere: Can we do `for dependency in @service-config.dependencies or {}` here and remove this `if` statement?

exo-setup/src/docker-setup.ls
```diff
@@ -63,10 +64,11 @@ class DockerSetup
 
   # compiles list of names of dependencies a service relies on
   _get-service-dependencies: ->
-    dependencies = @_get-app-dependencies!
+    dependencies = []
```
hugoesthere, line 66: `_get-service-dependencies` and `_get-app-dependencies` should not have changed. There can be Exosphere applications which use another messaging bus like Nats, which would completely replace ExoCom. The concept of Exosphere needs to be separate from that of ExoCom, so that means we still need to read `application.yml` to see which messaging bus (a.k.a. application dependency) an app is using. 

Sorry I misinterpreted your question when you asked if every service uses ExoCom. When an Exosphere application uses ExoCom, then yes every service depends on ExoCom. But an application could also use another messaging bus in which case every service doesn't rely on ExoCom.

exosphere-shared/example-apps/shared-dependency/mongo-service/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM nodesource/node
+
+# These steps ensure that npm install is only run when package.json changes
+COPY ./package.json .
+RUN npm install --production
+COPY . .
```
hugoesthere: Let's use the same Dockerfile we use here for setting up `html-server`

exosphere-shared/example-apps/shared-dependency/mongo-service/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "version": "0.0.0",
+  "description": "for testing only, do not publish",
+  "dependencies": {
+    "exoservice": "0.22.1"
+  },
+  "devDependencies": {
+    "david": "11.0.0",
+    "o-tools": "0.8.47"
+  }
+}
```
hugoesthere: Are these packages used at all?

exosphere-shared/example-apps/shared-dependency/web-server/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM nodesource/node
+
+# These steps ensure that npm install is only run when package.json changes
+COPY ./package.json .
+RUN npm install --production
+COPY . .
```
hugoesthere: Use same file as `html-server`'s Dockerfile

exosphere-shared/example-apps/shared-dependency/web-server/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "Example app that is running",
+  "devDependencies": {
+    "o-tools": "0.8.47"
+  },
+  "main": "server.ls",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "license": "ISC"
+}
```
hugoesthere: Let's remove all dependencies and scripts if they're not being used, and we should update the description to reflect the fact that this app isn't being run.

exosphere-shared/example-apps/shared-dependency/web-server/service.yml
```diff
@@ -0,0 +1,25 @@
...
+    - users.create
+  receives:
+    - users.listed
+    - users.created
+
+docker:
+  ports:
+    - '4000:4000'
+
+dependencies:
```
hugoesthere: The `web-server` shouldn't really be relying on mongo.

hugoesthere (commented):
One thing that happens with this implementation is if an application lists `mongo` as a dependency, then every service will "rely" on it (it will list `mongo` in the `depends_on` field of every service in `docker-compose.yml`). This makes no effective difference to the services, but now Docker will wait for `mongo` to start up before starting up every service. 

I think the simplest solution would be to go with this implementation for now. And then in the future if we choose to we can list a field that says "always include me as a dependency in every service" versus "only include me as a dependency if I say so". Besides, we are planning on booting up these Docker containers in multiple steps anyways (messaging bus, then dependencies, then services) - this might be a problem we want to solve when we get to implementing that. @kevgo thoughts?

cheahuychou:
> One thing that happens with this implementation is if an application lists mongo as a dependency, then every service will "rely" on it (it will list mongo in the depends_on field of every service in docker-compose.yml). 

That's not true (if I understand what you wrote correctly). For example, in[ here ](https://github.com/Originate/exosphere/blob/cm-shared-dependencies/exo-setup/features/setup.feature#L86)`html-server` service does not have `mongo` under `depends_on`

hugoesthere:
Not yet, but that's because `_get-app-dependencies` in `exo-setup/src/docker-setup.ls` is only returning Exocom when it should be returning a list of all app dependencies. Once you make that change I requested then every service will be listing all application dependencies under `depends_on`.

cheahuychou:
I know, that was what the change was for. The issue is it is hard to tell which dependencies are shared by all services instead of just a few services.

exo-setup/src/docker-setup.ls
cheahuychou: Sorry, this branch just was not synced with its parent yet.

cheahuychou:

exo-setup/src/docker-setup.ls
```diff
@@ -4,7 +4,7 @@ require! {
   'handlebars' : Handlebars
   'js-yaml' : yaml
   'path'
-  'prelude-ls' : {Obj, map}
+  'prelude-ls' : {Obj, map, union, compact}
```
hugoesthere: Doesn't look like `compact` is being used

exo-setup/src/docker-setup.ls
```diff
@@ -33,7 +33,8 @@ class DockerSetup
       environment: @_get-docker-env-vars!
       depends_on: @_get-service-dependencies!
     for dependency in @service-config.dependencies or []
-      docker-config[dependency.name + dependency.version] = @_get-service-dependency-docker-config dependency.name, dependency.version, dependency.config
+      if dependency.config
```
hugoesthere, line 36: Let's add a comment: 
```
# only add dependency to service if 'config' field exists,
# otherwise the dependency has already been listed as an application dependency
```

exo-setup/yarn.lock
```diff
@@ -887,6 +887,10 @@ mz@^2.4.0:
     object-assign "^4.0.1"
     thenify-all "^1.0.0"
 
+nitroglycerin@1.1.2:
+  version "1.1.2"
+  resolved "https://registry.yarnpkg.com/nitroglycerin/-/nitroglycerin-1.1.2.tgz#3bbc98d18999a027e72a9e332c5fde1049420417"
+
```
hugoesthere: This shouldn't be in `yarn.lock` anymore, run `yarn` again to update this file.

exosphere-shared/example-apps/complex-setup-app/mongo-service/server.ls
```diff
@@ -0,0 +1,7 @@
+module.exports =
+
+  'mongo.list': (_, {reply}) ->
+    console.log "'mongo' service received message 'mongo.list'"
+    reply 'mongo.listed' do
+      * name: 'Jean-Luc Picard'
+      * name: 'William Riker'
```
hugoesthere: Is this file needed if we're only using this service for testing setup?

exosphere-shared/yarn.lock
```diff
@@ -1170,3 +1299,12 @@ wrappy@1:
...
   resolved "https://registry.yarnpkg.com/xtend/-/xtend-4.0.1.tgz#a5c6d532be656e23db820efb943a1f04998d63af"
+
+yargs@~3.10.0:
+  version "3.10.0"
+  resolved "https://registry.yarnpkg.com/yargs/-/yargs-3.10.0.tgz#f7ee7bd857dd7c1d2d38c0e74efbd681d1431fd1"
+  dependencies:
+    camelcase "^1.0.2"
+    cliui "^2.1.0"
+    decamelize "^1.0.0"
+    window-size "0.1.0"
```
hugoesthere: There seems to be a lot of changes here, did you make sure to run `yarn` on the latest `package.json`?

exosphere-shared/example-apps/complex-setup-app/mongo-service/service.yml
```diff
@@ -0,0 +1,20 @@
+type: users
+description: generic mongoDB storage service
+author: exospheredev
+
+setup: yarn install
+startup:
+  command: node_modules/exoservice/bin/exo-js
```
hugoesthere: We can just do `echo "does not run"` here as well.

exosphere-shared/example-apps/complex-setup-app/application.yml
```diff
@@ -15,3 +27,6 @@ services:
   private:
     todo-service:
       location: ./todo-service
+    users:
+      namespace: mongo
+      location: ./mongo-service
```
hugoesthere: If you want you can do:
```
users-service:
  location: ./users-service
```
to make your tests easier

hugoesthere (approved):
Looks good!

PR #301, opened 2017-06-23
Update travis.yml

hugoesthere:
`docker-engine` has been depricated so this installation was breaking the build. Docker can be used in Travis by simply listing it under `services` which we already do.


cheahuychou (approved):

kevgo:
That line was added to update Docker to the latest version. If it works with the default version installed on Circle now, all the better!

PR #302, opened 2017-06-27
port `exo-create` to Go

cheahuychou:
Port `exo-create` to Go and modify the command to only support `exo create application` (`exo create service` is no longer supported). The new `exo create application` also adds an empty subdirectory `.exosphere` to the application directory.


cli/features/abbreviations.feature
```diff
@@ -28,24 +28,16 @@ Feature: executing abbreviated Exosphere commands
...
-  Scenario Outline: executing "exo create service" with abbreviated commands
-    When executing the abbreviated command "<CREATE-SERVICE-COMMAND>" in the terminal
-    Then the full command "exo create service" is executed
-  Examples:
-    | CREATE-SERVICE-COMMAND |
-    | exo cr s               |
-    | exo creat serv         |
+    | CREATE-APP-COMMAND    |
+    | exo cr application    |
+    | exo creat application |
```
hugoesthere: Does this work if we keep:
```
| exo cr a      |
| exo creat app |
```

exo-create/features/missing-argument.feature
```diff
@@ -1,12 +1,12 @@
...
+  I want to be able to start the exo-create command without arguments
   So that I can call this command correctly and build up muscle memory.
 
   - calling "exo-create" without an argument displays the list of entities that can be created
 
 
   Scenario: the user calls "exo create" without arguments
     When running "exo-create" in the terminal
-    Then it prints "Error: missing entity for 'create' command" in the terminal
+    Then it prints "Error: missing argument for 'create' command" in the terminal
```
hugoesthere: Since `exo create application` only creates an application now, I wonder if we can shorten it to just `exo create` now? We no longer have to distinguish between `exo create service` and `exo create application` so the extra argument seems unnecessary here.

@kevgo thoughts?

cli/features/abbreviations.feature
cheahuychou: It doesn't because Go doesn't have anything similar to`abbrev` so we can't do [this](https://github.com/Originate/exosphere/blob/master/exo-create/src/cli.ls#L20).

exo-create/features/missing-argument.feature
cheahuychou: I was going to ask about this too.

cli/features/abbreviations.feature
kevgo: If Go doesn't support abbreviations, we can remove that feature. It wasn't used much. For this PR, it is okay that the user has to type `exo create` in full.

exo-create/bin/spec.cmd
```diff
@@ -1,3 +0,0 @@
-call node_modules\o-tools-livescript\bin\build
-call node_modules\o-tools\bin\lint
-node_modules\.bin\cucumber-js %*
```
kevgo, line 3: This file is needed for Windows support. Please try to keep it. Line 3 has to change to `go test`, the rest can stay the same.

exo-create/features/missing-argument.feature
kevgo: Yes, I wanted to say the same thing. Let's simplify this command to just `exo create`. Good idea!

exo-create/helpers/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/Originate/exosphere/exo-create/types"
+)
+
+func CreateApplicationYAML(appConfig types.AppConfig) error {
+	dir, err := os.Executable()
+	if err != nil {
+		return err
+	}
+	templatePath := path.Join(dir, "..", "..", "..", "exosphere-shared", "templates", "create-app", "application-go.yml")
```
kevgo: I'm not sure this will work when installing just the Go binary on a user's computer. Keep in mind that Go compiles the entire source code into a single file. So we don't have access to files of the source code tree anymore, and there will be no other directories like `exosphere-shared` there. 

We have two options here: 

1. hard-code creating the application.yml file here. Meaning you concatenate the content using string variables together here. Something like:

```go
const template = `name: %s
description: %s
version: %s

dependencies:
  %s

environments:
  %s

services:
  public:
  private:
`
content = fmt.sprintf(template, name, ...)
```
2. since this is complicated and needs a ton of user input, use boilr to ask the user all the questions and generate this file.

exo-create/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/Originate/exosphere/exo-create/helpers"
+	"github.com/Originate/exosphere/exo-create/types"
+	"github.com/Originate/exosphere/exo-create/user_input"
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "exo create",
+	Short: "Usage: \"exo create application\"\nCreate a new Exosphere application\n  - options: \"[<app-name>] [<app-version>] [<exocom-version>] [<app-description>]\"",
```
kevgo: - use [string literals](https://golang.org/ref/spec#String_literals) so that you don't have to escape the double quotes
- you don't have to provide the options - Cobra does that for you automatically
- the `Short` text should just quickly describe what the command does. For example "Scaffolds a new Exosphere application"

exo-create/main.go
```diff
@@ -0,0 +1,75 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			fmt.Println("Error: missing argument for 'create' command")
+			return
+		} else if args[0] == "help" {
```
kevgo: Cobra does the "help" command for you automatically. This will all work out of the box once we switch to the "proper" Cobra setup where all commands are in one place.

exo-create/main.go
```diff
@@ -0,0 +1,75 @@
...
+			}
+			return
+		} else if args[0] != "application" {
+			fmt.Println("Error: cannot create '" + args[0] + "'")
+			return
+		}
+		fmt.Print("We are about to create a new Exosphere application\n\n")
+		appConfig := getAppConfig(args)
+		if err := helpers.CreateApplicationYAML(appConfig); err != nil {
+			panic(err)
```
kevgo: Instead of throwing so many panics manually, please use [log.fatalf](https://golang.org/pkg/log/#Fatalf) to print a clean error message and exit with a non-zero exit code.

exo-create/main.go
```diff
@@ -0,0 +1,75 @@
...
+		if err := os.Mkdir(path.Join(appConfig.AppName, ".exosphere"), os.FileMode(0777)); err != nil {
+			panic(err)
+		}
+		fmt.Println("\ndone")
+	},
+}
+
+func getAppConfig(args []string) types.AppConfig {
+	var appName, appVersion, exocomVersion, appDescription string
+	reader := bufio.NewReader(os.Stdin)
```
kevgo: Let's try to use Boilr instead of hard-coding all this user input here.

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/Originate/exosphere/exo-create/test_helpers"
+)
+
+var cmd *exec.Cmd
+var in io.WriteCloser
+var out bytes.Buffer
+var appDir string
+var err error
+
+type AppConfig struct {
```
kevgo: Why does the test file define the AppConfig? I would expect that the application configuration is defined by the application, not its test.

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+	AppName, AppVersion, ExocomVersion, AppDescription string
+}
+
+func run(cwd, command string) error {
```
kevgo: Please add a comment that summarizes this method. Example:

```
// Runs the given command in the given directory
```
I would also make the command the first argument, and the directory the second one.

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+	cmdSegments := strings.Split(path.Join(cwd, "bin", command), " ")
```
kevgo: Good job figuring out how to run processes! Please take a look at [this run method](https://github.com/Originate/git-town/blob/master/src/util/util.go#L44-L52) from the Git Town code base and follow its structure. For example, provide the command to run as a string array (each word separated already). This prevents subtle bugs where arguments that contain spaces are treated as separate arguments

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+func validateTextContains(haystack, needle string) error {
+	if strings.Contains(haystack, needle) {
+		return nil
+	}
+	return fmt.Errorf("Expected:\n\n%s\n\nto include\n\n%s", haystack, needle)
+}
+
+func enterInput(row *gherkin.TableRow) error {
+	_, input := row.Cells[0].Value, row.Cells[1].Value
```
kevgo: Why do you compute `row.Cells[0].Value` when you don't store its content here?

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+	_, err := in.Write([]byte(input + "\n"))
```
kevgo: Nice!

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+	})
+
+	s.Step(`^executing "([^"]*)"$`, func(command string) error {
+		return run(cwd, command)
+	})
+
+	s.Step(`^starting "([^"]*)" in the terminal$`, func(command string) error {
+		appDir = path.Join(cwd, "tmp")
+		if err := testHelpers.EmptyDir(appDir); err != nil {
+			return err
```
kevgo: Let's [wrap the error](https://github.com/pkg/errors) with additional context here:

```go
return errors.Wrap(err, "cannot create temp  dir")

File: exo-create/main_test.go [hunk 15]
```diff
@@ -0,0 +1,162 @@
...
+		}
+		return run(cwd, command)
+	})
+
+	s.Step(`^entering into the wizard:$`, func(table *gherkin.DataTable) error {
+		for _, row := range table.Rows[1:] {
+			if err := enterInput(row); err != nil {
+				return err
```
kevgo: wrap error

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+		}
+		defer in.Close()
```
kevgo: Do we really want to close the input stream here? We might want to enter more things later. Can we leave it open and let the process end while the input stream is open?

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil
+	})
+
+	s.Step(`^running "([^"]*)" in the terminal$`, func(command string) error {
```
kevgo: How is this step different from `starting xxx in the terminal`?

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+		return run(cwd, command)
+	})
+
+	s.Step(`^waiting until I see "([^"]*)" in the terminal$`, func(expectedText string) error {
+		if err := cmd.Wait(); err != nil {
+			return err
+		}
+		childOutput = out.String()
+		return validateTextContains(childOutput, expectedText)
```
kevgo: This step should keep calling `cmd.Wait` until it either finds the given string in the output, or hits a timeout. Please read https://blog.golang.org/go-concurrency-patterns-timing-out-and for some background.

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+	})
+
+	s.Step(`^it prints "([^"]*)" in the terminal$`, func(expectedText string) error {
```
kevgo: This step should also wait until the given text appears in the output.

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	s.Step(`^my application contains the file "([^"]*)" with the content:$`, func(filePath string, expectedContent *gherkin.DocString) error {
+		bytes, err := ioutil.ReadFile(path.Join(appDir, filePath))
+		if err != nil {
+			return err
+		}
+		return validateTextContains(strings.TrimSpace(string(bytes)), strings.TrimSpace(expectedContent.Content))
+	})
+
+	s.Step(`^my workspace contains the file "([^"]*)" with content:$`, func(fileName string, expectedContent *gherkin.DocString) error {
```
kevgo: The difference between "workspace" and "application" in this and the previous step is that "workspace" is the main directory ("tmp"), and "application" is the application directory ("tmp/app"). These two steps seem to check the same directory

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	s.Step(`^my workspace contains the empty directory "([^"]*)"`, func(directory string) error {
+		if _, err := os.Stat(path.Join(appDir, directory)); err == nil {
+			return nil
+		}
+		if os.IsNotExist(err) {
+			return fmt.Errorf("%s does not exist", directory)
+		}
+		return nil
+	})
```
kevgo: You need to return the error here.

exo-create/main_test.go
```diff
@@ -0,0 +1,162 @@
...
+			return err
+		}
+		return validateTextContains(strings.TrimSpace(string(bytes)), strings.TrimSpace(expectedContent.Content))
+	})
+
+	s.Step(`^my workspace contains the empty directory "([^"]*)"`, func(directory string) error {
+		if _, err := os.Stat(path.Join(appDir, directory)); err == nil {
+			return nil
+		}
+		if os.IsNotExist(err) {
```
kevgo: You might want to verify that the file system entry is actually a directory here.

exo-create/test_helpers/empty_directory.go
```diff
@@ -0,0 +1,12 @@
+package testHelpers
+
+import (
+	"os"
+)
+
+func EmptyDir(dir string) error {
```
kevgo: Did you run the linter? Exported functions typically need a comment.

exo-create/types/types.go
```diff
@@ -0,0 +1,6 @@
+package types
+
+// AppConfig represents the configuration for an application
+type AppConfig struct {
```
kevgo: This struct seems to be defined in two places.

exosphere-shared/templates/create-app/application-go.yml
```diff
@@ -0,0 +1,11 @@
+name: {{.AppName}}
```
kevgo: Using explicit templates in files does not work in Go. Let's use strings in files instead.

kevgo (commented):
Hey @cheahuychou, very good job making this all work! Some comments below. We can discuss more later today once I'm out of the business meeting here in SF.

exo-create/main.go
cheahuychou: I don't know if this is possible since we will need to set up a template folder and you said "we don't have access to files of the source code tree anymore, and there will be no other directories like exosphere-shared there."

kevgo: Yes, we have to find a way to tell Boilr about our template.

exo-create/main_test.go
cheahuychou: I've changed `run` to take in a string array but I still need to process `command` in a separate function because it is passed in from cucumber tests as a single string.

exo-create/main.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+		fmt.Print("We are about to create a new Exosphere application\n\n")
+		templatePath, err := helpers.CreateTemplate()
+		if err != nil {
+			log.Fatalf("Failed to create the template")
+		}
+		template, err := template.Get(templatePath)
+		if err != nil {
+			log.Fatalf("Failed to fetch the application template")
+			os.Exit(1)
```
charlierudolph: this can be removed as `log.Fatalf` exits

exo-create/main.go
```diff
@@ -0,0 +1,48 @@
...
+		if len(args) == 1 && args[0] == "help" {
+			if err := cmd.Help(); err != nil {
+				panic(err)
+			}
+			return
+		}
+		fmt.Print("We are about to create a new Exosphere application\n\n")
+		templatePath, err := helpers.CreateTemplate()
+		if err != nil {
+			log.Fatalf("Failed to create the template")
```
charlierudolph: Lets also output the error here so we know what the error is. Something like: `log.Fatal("Failed to create template:", err)`. Same for the other three instances in this file.

exo-create/helpers/helpers.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+)
+
+func createProjectJSON(templateDir string) error {
+	content := "{\n\"AppName\": \"my-app\",\n\"ExocomVersion\": \"0.22.1\",\n\"AppVersion\": \"0.0.1\",\n\"AppDescription\": \"\"\n}"
```
charlierudolph: This can be defined outside the func so we can use actual newlines to make it more readable
```go
var projectJsonContent = `
{
  "AppName": "my-app",
  "ExocomVersion": "0.22.1",
  "AppVersion": "0.0.1",
  "AppDescription": ""
}
`
```
Same for the yml content.

exo-create/helpers/helpers.go
```diff
@@ -0,0 +1,46 @@
...
+
+func CreateTemplate() (string, error) {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return "", err
+	}
+	templateDir := path.Join(cwd, "tmp")
+	appDir := path.Join(templateDir, "template/{{AppName}}")
+	if err := os.MkdirAll(path.Join(appDir, ".exosphere"), os.FileMode(0777)); err != nil {
+		return templateDir, fmt.Errorf("Failed to create the neccessary directories for the template")
```
charlierudolph: Lets wrap the error here so we don't lose information about why it failed. Also lets return an empty string instead of templateDir in the case where we are returning an error.

Same for the two similar lines below

exo-create/helpers/helpers.go
```diff
@@ -0,0 +1,46 @@
...
+	content := "{\n\"AppName\": \"my-app\",\n\"ExocomVersion\": \"0.22.1\",\n\"AppVersion\": \"0.0.1\",\n\"AppDescription\": \"\"\n}"
+	return ioutil.WriteFile(path.Join(templateDir, "project.json"), []byte(content), 0777)
+}
+
+func createApplicationYAML(appDir string) error {
+	content := "name: {{AppName}}\ndescription: {{AppDescription}}\nversion: {{AppVersion}}\n\ndependencies:\n  - name: exocom\n    version: {{ExocomVersion}}\n\nservices:\n  public:\n  private:\n"
+	return ioutil.WriteFile(path.Join(appDir, "application.yml"), []byte(content), 0777)
+}
+
+func CreateTemplate() (string, error) {
```
charlierudolph: Thoughts on naming this `CreateTemplateDir`? Similar idea for `RemoveTemplate`

exo-create/test_helpers/test_helpers.go
```diff
@@ -0,0 +1,26 @@
...
+
+func ReformatCommand(command, cwd string) []string {
+	return strings.Split(path.Join(cwd, "bin", command), " ")
+}
+
+func ValidateTextContains(haystack, needle string) error {
+	if strings.Contains(haystack, needle) {
+		return nil
+	}
+	return fmt.Errorf("Expected:\n\n%s\n\nto include\n\n%s", haystack, needle)
```
charlierudolph: The template can be defined outside the funcs so we can more easily see the formatting
```go
var validateTextContainsErrorTemplate = `
Expected:

%s

to include

%s
`
```

charlierudolph (commented):
Looks great

cheahuychou:
@kevgo @charlierudolph all comments addressed.

exo-create/helpers/helpers.go
```diff
@@ -0,0 +1,67 @@
...
+const projectJsonContent = `
+{
+  "AppName": "my-app",
+  "ExocomVersion": "0.22.1",
+  "AppVersion": "0.0.1",
+  "AppDescription": ""
+}
+`
+
+const applicationYmlContent = `name: {{AppName}}
```
charlierudolph, line 20: I think we add a newline here
```golang
const applicationYmlContent = `
name: {{AppName}}
`
```

charlierudolph (approved):

exo-create/helpers/helpers.go
cheahuychou: While this helps with readability of the code, I think it adds an empty line to the top of `application.yml` file. I don't know if this affects the code that reads the file.

charlierudolph: Auhh okay. Then lets remove the newline we are currently adding from the project.json. Than go ahead and ship if the build stays green

cheahuychou: `project.json` is a tmp file and is only used by `boilr`. `boilr` can parse it fine so I think we can leave it as it is?

charlierudolph: sounds good

PR #303, opened 2017-06-28
Mc add go project structure

mcharawi:


cmd/root.go
```diff
@@ -0,0 +1,88 @@
+// Copyright © 2017 NAME HERE <EMAIL ADDRESS>
```
kevgo: Please remove this legalese boilerplate comment.

cmd/root.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	// Here you will define your flags and configuration settings.
+	// Cobra supports persistent flags, which, if defined here,
+	// will be global for your application.
+	RootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.exo-go.yaml)")
```
kevgo: Comment this out for now, we don't use config files.

cmd/root.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Cobra also supports local flags, which will only run
+	// when this action is called directly.
+	RootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
```
kevgo: comment out

bin/setup
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+
+go get github.com/mitchellh/go-homedir github.com/spf13/viper github.com/mattn/go-shellwords github.com/termie/go-shutil
```
kevgo: Please use [glide](https://github.com/Masterminds/glide) for dependency management, instead of hard-coding the dependencies here.

cmd/root.go
```diff
@@ -0,0 +1,88 @@
...
+		// Search config in home directory with name ".exo-go" (without extension).
+		viper.AddConfigPath(home)
+		viper.SetConfigName(".exo-go")
+	}
+
+	viper.AutomaticEnv() // read in environment variables that match
+
+	// If a config file is found, read it in.
+	if err := viper.ReadInConfig(); err == nil {
+		fmt.Println("Using config file:", viper.ConfigFileUsed())
```
kevgo: I think we can comment out this whole method body, since we don't use config files for now.

exo_go_test.go
```diff
@@ -0,0 +1,62 @@
...
+// the temp dir in which the test repos live
+var testRoot string
+
+// the output of the last command run
+var output string
+
+// Cucumber step definitions
+func FeatureContext(s *godog.Suite) {
+
+	s.Step(`^it prints:$`, func(text *gherkin.DocString) error {
```
kevgo: Let's name the parameter `expected`, so that we know this is the expected text.

exo_go_test.go
```diff
@@ -0,0 +1,62 @@
...
+			return errors.New("output does not match")
+		}
+		return nil
+	})
+
+	s.Step(`^running "([^"]*)"$`, func(command string) (err error) {
+		command = makeCrossPlatformCommand(command)
+		words, err := shellwords.Parse(command)
+		check(err)
+		err = run(words)
```
kevgo: this err is not used

exo_go_test.go
```diff
@@ -0,0 +1,62 @@
...
+		fmt.Println(output)
```
kevgo: remove debug statement

features/version.feature
```diff
@@ -0,0 +1,15 @@
...
+  So that I know whether I am up to date, and can work with customer support on investigating issues.
+
+  - run "exo-go version" to see the version of Exosphere you are running
+
+
+  Scenario: displaying the version
+    When running "exo-go version"
+    Then it prints:
+      """
+      Exosphere-Go v\d+\.\d+(\.\d+)?
```
kevgo, line 14: Woohoo for using a regex here!

helpers_test.go
```diff
@@ -0,0 +1,85 @@
...
+	shutil "github.com/termie/go-shutil"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func commitAllChanges() {
+	err := run([]string{"git", "add", "-A"})
```
kevgo: Use a variadic function to make this interface nicer:

```go
err := run("git", "add", "-A")
```

Example implementation at 
https://github.com/Originate/git-town/blob/master/src/util/util.go#L45

helpers_test.go
```diff
@@ -0,0 +1,85 @@
...
+	checkText(err, output)
```
kevgo: Where does output come from? Do we have to use global variables here? Can you return `output` from `run`?

helpers_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func checkText(e error, text string) {
+	if e != nil {
+		fmt.Println(text)
+		panic(e)
+	}
+}
+
+func makeCrossPlatformCommand(command string) string {
+	if os.PathSeparator == '\\' {
```
kevgo: Alternative, slightly more readable version:

```go
if runtime.GOOS == "windows" {
```

helpers_test.go
```diff
@@ -0,0 +1,85 @@
...
+		command = strings.Replace(command, "/", "\\\\", -1)
```
kevgo: Use https://golang.org/pkg/path/filepath/#FromSlash

helpers_test.go
```diff
@@ -0,0 +1,85 @@
...
+		command = strings.Replace(command, "bin\\spec", "bin\\spec.cmd", -1)
+	}
+	return command
+}
+
+// Runs the given command, stores the output in "output"
+func run(commands []string) (err error) {
+	if commands[0] == "morula" && !containsColorArgument(commands) {
```
kevgo: morula??

helpers_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Returns the project names in the given human-friendly project name list
+func splitProjectNames(projectNames string) (result []string) {
+	for _, projectName := range strings.Split(projectNames, "and") {
+		result = append(result, strings.Trim(projectName, " \""))
+	}
+	return
+}
+
+func switchBranch(branchName string) {
```
kevgo: This doesn't seem to be used. Why do you want to switch Git branches here?

kevgo (commented):
Good stuff! I assume you are going to fill in the command source code later.

PR #305, opened 2017-07-01
go: move to single binary

charlierudolph:


cli/src/cli.ls
```diff
@@ -91,3 +91,20 @@ function print-usage
 
 function command-names
   Object.keys commands
+
+function get-go-binary-os
```
mcharawi, line 95: I was under the impression that we decided to bundle five binaries when we publish and not deal with determining the end user's platform requirements?

exo-go/features/help.feature
```diff
@@ -0,0 +1,43 @@
...
+      """
+      Usage:
+        exo [command]
+
+      Available Commands:
+        clean       Removes dangling Docker images and volumes.
+        create      Create a new Exosphere application
+        version     Exosphere go version number
+
+      Use "exo [command] --help" for more information about a command.
```
mcharawi, line 20: Use `exo help [command]`...

mcharawi (changes requested):

bin/spec
```diff
@@ -1,6 +0,0 @@
-#!/usr/bin/env bash
```
kevgo, line 1: Why does this go away? It would be nice to have a global `spec` command that runs the Node and Go tests.

cli/src/cli.ls
```diff
@@ -91,3 +91,20 @@ function print-usage
...
+  switch process.platform
```
kevgo, line 96: Livescript power as long as we have it!

```ls
switch process.platform
  | 'darwin' => 'darwin'
  | 'linux'  => 'linux'
  | 'win32'  => 'windows'
  | _        => throw new Error('Unsupported operating system. Please open an issue with your operating system and system architecture.')

File: exo-go/bin/lint [hunk 5]
```diff
@@ -8,5 +8,5 @@ fi
 
 # required by gotype to work properly
 # see https://github.com/golang/go/issues/12703#issuecomment-144429128
-go install
-gometalinter.v1 --deadline 60s ./
+go install ./...
+gometalinter.v1 --deadline 60s ./...
```
kevgo, line 12: Does this lint the entire `vendor` folder as well? We should avoid that, it takes forever and causes tons of lint errors that aren't our fault.

kevgo (approved):

mcharawi (approved):

cli/src/cli.ls
charlierudolph: We are. This is used for determining which of those 5 binaries to execute.

exo-go/features/help.feature
charlierudolph: This was generated by the cobra library. I'd prefer we just use whats builtin to cobra instead of implementing our own. Once we switch over to fully using this binary this will be a breaking change

bin/spec
charlierudolph: This was only added for the go binaries since it was top level. Moving it into a folder it was no longer needed. If we want the global spec that can be added separately

exo-go/bin/lint
charlierudolph: The vendor folder is on the top level instead of in exo-go, thus we are already avoiding that

cli/src/cli.ls
charlierudolph: updated

PR #306, opened 2017-07-03
port `exo-add` to Go

cheahuychou:
resolves https://github.com/Originate/exosphere-sdk/issues/306

Ported exo-create to Go and modified the command according to [this ticket](https://github.com/Originate/exosphere-sdk/issues/306). `exo-add` now asks users to select a template from the templates it finds in `.exosphere` subdirectory and uses [boilr](https://github.com/tmrts/boilr) for scaffolding. It also creates a simple `service.yml` if the file isn't provided in the template. So now `exo-add` only accepts templates that follow [boilr template format](https://github.com/tmrts/boilr/wiki/Creating-Templates). I created boilr templates for `exoservice-es6-mongodb` and `htmlserver-express-es6` under `exosphere-shared/boilr-templates`because I had to change the tutorial tests. We might want to do that for the rest of the templates too and update the tutorial documentation.


cli/features/tutorial.feature
```diff
@@ -44,35 +44,37 @@ Feature: Following the tutorial
...
-          version: 0.22.1
-
+      - name: exocom
+        version: 0.22.1
       services:
         public:
           html-server:
             location: ./html-server
-        private:
+        private: {}
```
cheahuychou, line 77: I don't think there is a way to get rid of `{}`. It could be pointer problems but I did try setting it to `nil`. I did manage to make it not initialize `config` map for the dependencies though.

cli/features/steps/steps.ls
```diff
@@ -41,6 +42,11 @@ defineSupportCode ({Given, When, Then}) ->
     fs.write-file-sync path.join(app-dir, filename), file-content
 
 
+  Given /^my application contains the template folder "([^"]*)"$/ (template-dir) ->
+    template-name = (template-dir.split "/")[1]
```
kevgo: Use more traditional formatting:

```ls
template-name = template-dir.split("/")[1]
```

(So nice that Go does not need these types of comments)

cli/features/steps/steps.ls
```diff
@@ -143,6 +149,11 @@ defineSupportCode ({Given, When, Then}) ->
       jsdiff-console actual-content.toString!trim!, expected-content.trim!, done
 
 
+  Then /^my workspace contains the empty directory "([^"]*)"$/, (directory) ->
+    fs.stat path.join(app-dir, directory), (err, stat) ~>
```
kevgo: This is an asynchronous step, it needs a 'done' callback. 

```ls
Then /^my workspace contains the empty directory "([^"]*)"$/, (directory, done) ->
  fs.stat path.join(app-dir, directory), done

File: cli/features/tutorial.feature (hunk 1 above)
Comment (kevgo) [reply cli/features/tutorial.feature:77]: good enough for now

File: exo-add/features/add-service.feature [hunk 4]
```diff
@@ -0,0 +1,47 @@
+Feature: interactive scaffolding
+
+  As a developer not using ExoService
```
kevgo: > As a tech lead
I want to define the templates my team uses for creating new services
So that they don't copy-and-past old code around.

exo-add/features/add-service.feature
```diff
@@ -0,0 +1,47 @@
...
+  I want to have an easy way to use my own service templates
+  So that I don't waste time copy-and-pasting a bunch of code.
+
```
kevgo: Let's add a rules section here:

```
rules
- templates for new services are located in the ".exosphere" folder
- each subdirectory in that folder is a template
- the templates are applied using boilr

File: exo-add/features/no-templates.feature [hunk 6]
```diff
@@ -0,0 +1,13 @@
+Feature: Attempting to add a duplicate service
```
kevgo: Wrong feature name

exo-add/features/no-templates.feature
```diff
@@ -0,0 +1,13 @@
...
+  I want Exosphere to alert me when attempting to create a service without providing templates
+  So that I know that .exosphere folder either does not contain valid boilr templates or is empty
+
+
+  Scenario: Adding a service without providing any templates
+    Given I am in the root directory of an empty application called "test app"
+    And my application contains the empty directory ".exosphere"
+    When starting "exo-add" in this application's directory
+    Then it exits with code 1
+    And I see the error "no templates found"
```
kevgo: Let's provide a more helpful error message here:

```
no templates found

Please add templates to the ".exosphere" folder of your code base.

File: exo-add/helpers/helpers.go [hunk 8]
```diff
@@ -0,0 +1,107 @@
...
+	"path"
+
+	"github.com/Originate/exosphere/exo-add/os_helpers"
+	"github.com/Originate/exosphere/exo-add/types"
+	"github.com/Originate/exosphere/exo-add/user_input"
+	"github.com/tmrts/boilr/pkg/template"
+	"gopkg.in/yaml.v2"
+)
+
+func CheckForService(serviceRole string, existingServices []string) {
```
kevgo: The name `CheckXXX` implies that this method just checks for something. Since this method also exits the app, let's call it `VerifyServiceDoesNotExist`.

exo-add/helpers/helpers.go
```diff
@@ -0,0 +1,107 @@
...
+	if contains(existingServices, serviceRole) {
+		fmt.Printf(`Service %v already exists in this application`, serviceRole)
+		os.Exit(1)
+	}
+}
+
+func GetExistingServices(services types.Services) []string {
+	existingServices := []string{}
```
kevgo: Does this work?

```go
return append(services.Private, services.Public...)

File: exo-add/helpers/helpers.go [hunk 10]
```diff
@@ -0,0 +1,107 @@
...
+	for service, _ := range services.Private {
+		existingServices = append(existingServices, service)
+	}
+	for service, _ := range services.Public {
+		existingServices = append(existingServices, service)
+	}
+	return existingServices
+}
+
+func GetTemplates() []string {
```
kevgo: All these public methods need documentation

exo-add/helpers/helpers.go
```diff
@@ -0,0 +1,107 @@
...
+	templatesDir := ".exosphere"
+	if !osHelpers.DirectoryExists(templatesDir) || osHelpers.IsEmpty(templatesDir) {
+		fmt.Print("no templates found")
```
kevgo: Use `log.fatal`

exo-add/helpers/helpers.go
```diff
@@ -0,0 +1,107 @@
...
+	return templates
+}
+
+func CreateTmpServiceDir(chosenTemplate string) string {
+	templateDir := path.Join(".exosphere", chosenTemplate)
+	template, err := template.Get(templateDir)
+	if err != nil {
+		log.Fatalf("Failed to fetch %s template: %s", chosenTemplate, err)
+	}
+	serviceTmpDir := path.Join("tmp", "service-tmp")
```
kevgo: Use https://golang.org/pkg/io/ioutil/#TempDir to create a temp dir

exo-add/main_test.go
```diff
@@ -0,0 +1,227 @@
...
+func FeatureContext(s *godog.Suite) {
+	var cwd, tmpDir, appDir string
+
+	s.BeforeSuite(func() {
+		var err error
+		cwd, err = os.Getwd()
+		if err != nil {
+			panic(err)
+		}
+		tmpDir = path.Join(cwd, "tmp")
```
kevgo: This can go away once you use https://golang.org/pkg/io/ioutil/#TempDir

exo-add/features/add-service.feature
```diff
@@ -0,0 +1,47 @@
...
+    And entering into the wizard:
+      | FIELD                         | INPUT          |
+      | template                      | 1              |
+      | Name                          | test-service   |
+      | ServiceType                   | web-service    |
+      | Description                   | testing        |
+      | Author                        | tester         |
+      | Protection Level              | 1              |
+    And waiting until the process ends
+    Then my application contains the file "application.yml" with the content:
```
kevgo: Let's use wording to make more clear that this is a `Then` step and not a `Given` step: "Then my application __now__ contains the file...

exo-add/test_helpers/test_helpers.go
```diff
@@ -0,0 +1,52 @@
...
+	`
+
+func EmptyDir(dir string) error {
+	if err := os.RemoveAll(dir); err != nil {
+		return err
+	}
+	return os.Mkdir(dir, 0777)
+}
+
+func ReformatCommand(command string) []string {
```
kevgo: This doesn't seem to be used

exo-add/test_helpers/test_helpers.go
```diff
@@ -0,0 +1,52 @@
...
+func ValidateTextContains(haystack, needle string) error {
+	if strings.Contains(haystack, needle) {
+		return nil
+	}
+	return fmt.Errorf(validateTextContainsErrorTemplate, haystack, needle)
+}
+
+func WaitForText(stdout *bytes.Buffer, text string, duration int) error {
+	interval := time.Tick(100 * time.Millisecond)
+	timeout := time.After(time.Duration(duration) * time.Millisecond)
```
kevgo: You can inline these two variables, they are used only once

kevgo (commented):
Time to move to the unified Go architecture, there is a ton of code here that also exists in the other Go projects! I also want to see @charlierudolph's review of this.

exo-add/helpers/helpers.go
cheahuychou: No it doesn't because each of these is of type `map[string]ServiceConfig`.

exo-add/helpers/helpers.go
cheahuychou: I did that but the step `And I see the error:` timed out  because`expectedText` didn't show up in `out.String()`. This is odd since `log.Fatal` should call `Print()` before `os.Exit(1)`. I have not figured out why that is so now I just call `Print` and `os.Exit(1)` separately.

charlierudolph:
You updated all the built in service templates but it appears the feature tests surrounding them were removed. Was that intentional or should those tests be added back?

charlierudolph:
I think we should update all the built in service templates to `boilr` format and add the tests back in for adding the different types of built in templates.

cheahuychou:
I only updated two templates: `exoservice-es6-mongodb` and `htmlserver-express-es6` because they're used in `cli/features/tutorial.feature`. I could update the rest.

kevgo:
Moving forward the templates won't be inside the tool anymore, but in a separate repo.

cheahuychou:
@kevgo So should I update the templates and add the tests back in this PR?

kevgo:
I think the final setup will be that these templates will be in their own repo, for example in a folder inside https://github.com/originate/exocom. The tests for these templates should live somewhere near that folder as well. They don't really test that Exosphere works, but that the templates work. So they shouldn't be in the Exosphere code base. 

What do you think about this setup:
* template to add an Exorelay-Go service are in https://github.com/originate/exocom/go/exorelay-templates
* tests for this template are in https://github.com/originate/exocom/go/exorelay-template-tests
* template to add an Exoservice-Go service are in https://github.com/originate/exocom/go/exoservice-templates
* tests for this template are in https://github.com/originate/exocom/go/exoservice-template-tests

cli/features/steps/steps.ls
```diff
@@ -143,6 +149,12 @@ defineSupportCode ({Given, When, Then}) ->
       jsdiff-console actual-content.toString!trim!, expected-content.trim!, done
 
 
+  Then /^my workspace contains the empty directory "([^"]*)"$/, (directory, done) ->
+    fs.stat path.join(app-dir, directory), (err, stat) ~>
```
charlierudolph: stat is unneedded

cli/features/support/world.ls
```diff
@@ -13,6 +13,11 @@ World = !->
     fs.copy-sync path.join(example-apps-path, app-name), app-dir
 
 
+  @checkout-service-template = (app-name, template-name) ->
+    template-dir = path.join process.cwd!, 'tmp', app-name, '.exosphere', template-name
+    fs.copy-sync path.join(process.cwd!, '..' 'exosphere-shared' 'templates', 'boilr-templates', template-name), template-dir
```
charlierudolph: thoughts on updating this to
```livescript
src = #...
dest = #....
fs.copy-sync src, dest
```
Having the src inlined makes this very hard for me to read

exo-go/src/app_config_helpers/index.go
```diff
@@ -0,0 +1,54 @@
...
+		log.Fatalf("Failed to unmarshal application.yml: %s", err)
+	}
+	return appConfig
+}
+
+// UpdateAppConfig adds serviceRole to the appConfig object and updates
+// application.yml
+func UpdateAppConfig(serviceRole string, appConfig types.AppConfig) {
+	reader := bufio.NewReader(os.Stdin)
+	switch protectionLevel := userInputHelpers.Choose(reader, "Protection Level:", []string{"public", "private"}); protectionLevel {
```
charlierudolph: do you need the `protectionLevel` variable? I think this could be
```go
switch userInputHelpers.Choose(reader, "Protection Level:", []string{"public", "private"}) {
```

exo-go/src/app_config_helpers/index.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		log.Fatalf("Failed to unmarshal application.yml: %s", err)
+	}
+	return appConfig
+}
+
+// UpdateAppConfig adds serviceRole to the appConfig object and updates
+// application.yml
+func UpdateAppConfig(serviceRole string, appConfig types.AppConfig) {
+	reader := bufio.NewReader(os.Stdin)
```
charlierudolph: Can we pass the reader in to this function instead of using `os.Stdin` inside of the this function? I think that helps make this function more reusable and unit-testable for the future

exo-go/src/os_helpers/index.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	_, err = f.Readdir(1)
+	return err == io.EOF
+}
+
+// IsValidTemplateDir returns true if the directory templateDir is a valid
+// boilr template directory
+func IsValidTemplateDir(templateDir string) bool {
+	return FileExists(path.Join(templateDir, "project.json")) && DirectoryExists(path.Join(templateDir, "template"))
+}
```
charlierudolph: I know I moved a lot of the methods around, but can we move this template_helpers since this is more about templates then a generic os helper?

exo-go/src/service_helpers/index.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/Originate/exosphere/exo-go/src/types"
+	"github.com/Originate/exosphere/exo-go/src/util"
+)
+
+// VerifyServiceDoesNotExist forces the program to exit with status code 1
+// if the service serviceRole already exists in existingServices
+func VerifyServiceDoesNotExist(serviceRole string, existingServices []string) {
+	if util.DoesStringArrayContain(existingServices, serviceRole) {
+		fmt.Printf(`Service %v already exists in this application`, serviceRole)
+		os.Exit(1)
```
charlierudolph: Lets not exit from this inside this method but just return an error or not. Lets leave just the commands to exit when needed

charlierudolph (commented):
I'm going to create an issue to update our existing code from using log.fatal/panic inside helper functions and instead returning the error. Ideally only the commands should be printing errors. This makes the helpers more reusable and unit-testable. Please follow that moving forward.

exo-go/src/template_helpers/index.go
```diff
@@ -0,0 +1,174 @@
...
+
+// CreateServiceYML creates service.yml for the service serviceRole by creating
+// a boilr template for service.yml, making boilr do the scaffolding and finally
+// remove the template
+func CreateServiceYML(serviceRole string) {
+	templateDir, err := createServiceTemplateDir(serviceRole)
+	if err != nil {
+		log.Fatalf("Failed to create the service.yml template: %s", err)
+	}
+	template, err := template.Get(templateDir)
```
cheahuychou: should rename `template` here because `template` refers to the boilr pkg

exo-go/src/template_helpers/index.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		log.Fatalf("Failed to fetch service.yml template: %s", err)
+	}
+	if err = template.Execute(serviceRole); err != nil {
```
cheahuychou: also here

exo-go/src/template_helpers/index.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return templates
+}
+
+// CreateTmpServiceDir makes bolir scaffold the template chosenTemplate
+// and store the scaffoled service folder in a tmp folder, and finally
+// returns the path to the tmp folder
+func CreateTmpServiceDir(chosenTemplate string) string {
+	templateDir := path.Join(".exosphere", chosenTemplate)
+	template, err := template.Get(templateDir)
```
cheahuychou: same here

exo-go/src/template_helpers/index.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		log.Fatalf("Failed to fetch %s template: %s", chosenTemplate, err)
+	}
+	serviceTmpDir, err := ioutil.TempDir("", "service-tmp")
+	if err != nil {
+		log.Fatalf(`Failed to create a tmp folder for the service "%s": %s`, chosenTemplate, err)
+	}
+	if err = template.Execute(serviceTmpDir); err != nil {
```
cheahuychou: and here

exo-go/src/service_helpers/index.go
cheahuychou: I was just following the original implementation and feature tests. So if we don't exit, do we just expect the users to do `Ctr-C` after they see the error?

charlierudolph: I still think we need `os.Exit(1)` somewhere, ideally in the `cmds/add.go`. I'd prefer it not to be in this function.

cheahuychou:

charlierudolph (approved):

PR #307, opened 2017-07-03
wait for expected text in exo-add tests

cheahuychou:
The original implementation never waits for the expected text, it always stops waiting after the first tick. It just happens to work because one tick time turns out to be enough. I had the correct implementation at some point and somehow it got replaced by this.


cheahuychou:
resolved in https://github.com/Originate/exosphere/commit/894de6544402d418ef6fa9196f1251ee31f51451

PR #322, opened 2017-07-07
fetch remote templates

### LINKED ISSUE #318: remote Boilr templates
> When developing a fleet of similar applications
I want to be able to use service templates from a remote location
So that I don't have to copy-and-paste templates into all my application code bases.

- if application.yml contains an entry "templates" that points to a [Boilr template URL](https://github.com/tmrts/boilr#download-a-template), Exosphere imports those templates into Boilr
- running "exo update-templates" re-imports the templates
- "exo create" has to ask for the template URL, if none is given it omits the entry in application.yml

cheahuychou:
resolves #318 

Add `exo fetch-templates` command that fetches git templates declared under `templates` in `application.yml` into `.exosphere` folder , and make those templates submodules of the exosphere application. This command assumes that the application lives in a git repository. If a template already exists, the command will update the submodule git URL and update the submodule 
and template accordingly.


exo-go/features/fetch-templates.feature
```diff
@@ -0,0 +1,53 @@
...
+      services:
+        public:
+          test-service:
+            location: ./test-service
+        private:
+      templates:
+        boilr-spark: https://github.com/tmrts/boilr-spark.git
+      """
+    When starting "exo fetch-templates" in my application directory
+    And waiting until the process ends
```
kevgo: These two steps can be unified to a step `When running ".....`. Running means starting and waiting until the process finishes.

exo-go/features/fetch-templates.feature
```diff
@@ -0,0 +1,53 @@
...
+    Then my application contains the directory ".exosphere/boilr-spark"
+    And my git repository has ".exosphere/boilr-spark" as a submodule
```
kevgo: I would combine these steps into one:

```cucumber
And my git repository has a submodule ".exosphere/boilr-spark" with remote "https://github.com/tmrts/boilr-spark.git"

File: exo-go/features/fetch-templates.feature [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+
+  When developing a fleet of similar applications
+	I want to be able to use service templates from a remote location
+	So that I don't have to copy-and-paste templates into all my application codebases.
+
+  - run "exo fetch-templates" to fetch remote service templates and make those templates submodules
+    of the application codebase
+
+
+  Scenario: fetching and updating remote service templates
```
kevgo: These are two different functionalities. Please create two scenarios here:

Scenario: adding a new template

Scenario: updating an existing template

exo-go/src/cmd/fetch-templates.go
```diff
@@ -0,0 +1,46 @@
...
+	Short: "Fetch service templates",
+	Long: `Fetch service templates
+This command must be called in the root directory of the application`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if util.PrintHelpIfNecessary(cmd, args) {
+			return
+		}
+		fmt.Print("We are about to fetch remote templates\n\n")
+		appConfig := appConfigHelpers.GetAppConfig()
+		if len(appConfig.Templates) > 0 {
```
kevgo: Is this if-clause really necessary? The for loop just does nothing if appConfig.Templates is an empty array.

kevgo (commented):
Nice, thanks for the quick turnaround! I suggest a slightly different API for users:
- `exo add-template <url>` adds a new template to the application
  - add to application.yml
  - create the submodule
- `exo remove-template <name>` removes the template from the application
- `exo fetch` just fetches updates for all existing git submodules in the `.exosphere` folder

cheahuychou:
@kevgo did you mean this? `exo add-template <name> <url>` adds a new template to the application? Also, can we call the last one `exo fetch-templates` so it's clear that we're fetching templates?

kevgo:
Yes on both. Thanks!

cheahuychou:
@kevgo all comments addressed. I wasn't sure how to test `fetch-templates` in the feature tests without doing something complicated with git so only tested that it ran without errors

exo-go/features/add_template/add_template.feature
```diff
@@ -0,0 +1,81 @@
...
+
+  When developing a fleet of similar applications
+	I want to be able to use service templates from a remote location
+	So that I don't have to copy-and-paste templates into all my application codebases.
+
+  - run "exo add-template" to add a remote service template to the application codebase
+
+
+  Scenario: adding a new service template
+  	Given I am in the root directory of an empty git application repository called "test app"
```
kevgo: I think here we can split up the steps, since they aren't really related.

```cucumber
Given I am in the root directory of an empty application called "test app"
And my application is a Git repository
```

exo-go/features/add_template/add_template.feature
```diff
@@ -0,0 +1,81 @@
...
+      templates:
+        boilr-spark: https://github.com/tmrts/boilr-spark.git
+      """
+
+
+  Scenario: adding an existing service template
+    Given I am in the root directory of an empty git application repository called "test app"
+    When running "exo add-template boilr-spark https://github.com/tmrts/boilr-spark.git" in my application directory
+    And starting "exo add-template boilr-spark https://github.com/tmrts/boilr-electron.git" in my application directory
+    And entering into the wizard:
```
kevgo: Let's define the output of the application here as well:

```cucumber
Then I see:
  """
  The template boilr-spark already exists
  """

File: exo-go/features/add_template/add_template.feature [hunk 7]
```diff
@@ -0,0 +1,81 @@
...
+        version: 0.22.1
+      services:
+        public: {}
+        private: {}
+      templates:
+        boilr-spark: https://github.com/tmrts/boilr-spark.git
+      """
+
+
+  Scenario: adding an existing service template
```
kevgo: Scenario: the service template already exists and is not overwritten

exo-go/features/add_template/add_template.feature
```diff
@@ -0,0 +1,81 @@
...
+        version: 0.22.1
+      services:
+        public: {}
+        private: {}
+      templates:
+        boilr-spark: https://github.com/tmrts/boilr-spark.git
+      """
+
+
+  Scenario: updating an existing service template
```
kevgo: Scenario: the service template already exists and is overwritten

exo-go/features/add_template/add_template.feature
```diff
@@ -0,0 +1,81 @@
...
+      services:
+        public: {}
+        private: {}
+      templates:
+        boilr-spark: https://github.com/tmrts/boilr-spark.git
+      """
+
+
+  Scenario: adding an existing service template
+    Given I am in the root directory of an empty git application repository called "test app"
```
kevgo: This step exists in all three scenarios, please extract it into a [background step](https://github.com/cucumber/cucumber/wiki/Background).

exo-go/features/add_template/missing_arguments.feature
```diff
@@ -0,0 +1,20 @@
+Feature: missing arguments
+
+  Scenario: attempting to run exo add-template without specifying template name or URL
+    Given I am in the root directory of an empty git application repository called "test app"
+    When starting "exo add-template" in my application directory
+    Then I eventually see:
```
kevgo: We don't need the "eventually" here. The step `Then I see` should wait until it finds the text in the output stream.

exo-go/features/fetch_templates/fetch_templates.feature
```diff
@@ -0,0 +1,13 @@
...
+
+  When developing an exosphere application
+	I want to be able to fetch the latest updates for the existing service templates in application codebase
+	So that I don't have to manually run the git command to update the submodules
+
+  - run "exo fetch-templates" to add a remote service template to the application codebase
+
+  Scenario: fetching updates for all existing service templates
+    Given I am in the root directory of an empty git application repository called "test app"
+    When running "exo add-template boilr-spark https://github.com/tmrts/boilr-spark.git" in my application directory
```
kevgo: We don't really need to explain using which comand-line call one adds templates here again, we already do this is the other feature file. This step is only here to set up the test environment. Let's do this a more clearly:

```cucumber
Given I am in the root directory of an empty application called "test app"
And my application is a Git repository
And my application has the templates:
  | NAME        | URL                                      |
  | boilr-spark | https://github.com/tmrts/boilr-spark.git |

File: exo-go/features/fetch_templates/fetch_templates.feature [hunk 12]
```diff
@@ -0,0 +1,13 @@
...
+    And running "exo fetch-templates" in my application directory
```
kevgo: ~~And~~When

exo-go/features/fetch_templates/fetch_templates.feature
```diff
@@ -0,0 +1,13 @@
...
+    Then it prints "done" in the terminal
```
kevgo: Let's also check that it exits with code 0 here, which indicates no error more than it printing "done" somewhere.

exo-go/features/help.feature
```diff
@@ -33,6 +36,17 @@ Feature: help command
         exo add
       """
 
+  Scenario: the user enters 'exo add-template help'
+    When running "exo add-template help" in the terminal
+    Then I see:
+      """
+      Adds a remote service template to .exosphere
```
kevgo: This is a proper sentence and should be terminated with a dot. The one on the next line as well.

exo-go/features/help.feature
```diff
@@ -33,6 +36,17 @@ Feature: help command
...
+      This command must be called in the root directory of the application
```
kevgo: I don't think we need to say this. Afaik all Exosphere commands need to be called in the root directory.

kevgo: Later we can add logic to find an `application.yml` file in the nearest parent directory and work there, which will allow to run any Exosphere command in any subdirectory of an application. But that comes later.

exo-go/features/remove_template/remove_template.feature
```diff
@@ -0,0 +1,35 @@
+Feature: removing existing templates
+
+  When developing an exosphere application
+	I want to be able to remove existing service templates from my application codebase
+	So that I don't have to manually run the git command to remove the submodules
+
+  - run "exo remove-template" to remove an existing service template from the application codebase
```
kevgo: the command is `exo remove-template <name>`

exo-go/features/remove_template/remove_template.feature
```diff
@@ -0,0 +1,35 @@
...
+
```
kevgo: Another rule, looking at your example, is that if the last template is removed, it also removes the `templates` key from `application.yml`. Might be worth pointing out here.

exo-go/features/remove_template/remove_template.feature
```diff
@@ -0,0 +1,35 @@
...
+        version: 0.22.1
+      services:
+        public: {}
+        private: {}
+      """
+
+
+  Scenario: removing a non-existing service template
+    Given I am in the root directory of an empty git application repository called "test app"
+    When starting "exo remove-template boilr-spark" in my application directory
```
kevgo: Let's make it more clear that the template doesn't exist here:

```
When starting "exo remove-template non-existing" in my application directory

File: exo-go/src/cmd/add_template.go [hunk 19]
```diff
@@ -0,0 +1,72 @@
...
+			if err := templateHelpers.AddTemplate(gitURL, templateName, templateDir); err != nil {
+				log.Fatalf(`Failed to add template "%s": %s`, templateName, err)
+			}
+		}
+		appConfig := appConfigHelpers.GetAppConfig()
+		if len(appConfig.Templates) == 0 {
+			appConfig.Templates = make(map[string]string)
+		}
+		appConfig.Templates[templateName] = gitURL
+		bytes, err := yaml.Marshal(appConfig)
```
kevgo: Please extract an `appConfigHelpers.SaveAppConfig()` helper here. Even if it doesn't save many lines of code, it still makes the code here more readable. Right now we are suddenly marshaling data structures. After the refactor it will say "the appConfig is getting saved". That's closer to the level at which the rest of the function is at.

exo-go/src/cmd/add_template.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil {
+			log.Fatalf("Failed to marshal application.yml: %s", err)
+		}
+		err = ioutil.WriteFile("application.yml", bytes, 0777)
+		if err != nil {
+			log.Fatalf("Failed to update application.yml: %s", err)
+		}
+		fmt.Println("\ndone")
```
kevgo: The Go way is that if an application doesn't output anything, that means "no error". Let's follow that here and don't print "done" at the end.

exo-go/src/cmd/fetch-templates.go
```diff
@@ -0,0 +1,31 @@
...
+var fetchTemplatesCmd = &cobra.Command{
+	Use:   "fetch-templates",
+	Short: "Fetches updates for all existing templates",
+	Long: `Fetches updates for all existing git submodules in the .exosphere folder
+This command must be called in the root directory of the application`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if util.PrintHelpIfNecessary(cmd, args) {
+			return
+		}
+		fmt.Print("We are about to fetch updates for remote templates\n\n")
```
kevgo: We are about to fetch updates for __the__ remote templates

exo-go/src/cmd/remove_template.go
```diff
@@ -0,0 +1,59 @@
...
+	Short: "Removes an existing service template from .exosphere",
+	Long: `Removes an existing service template from .exosphere
+This command must be called in the root directory of the application`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if util.PrintHelpIfNecessary(cmd, args) {
+			return
+		}
+		if len(args) != 1 {
+			fmt.Println("not enough arguments")
+      os.Exit(1)
```
kevgo: Please configure your editor to run `goimports` every time you save. This line here is indented with spaces, it should be tabs. @charlierudolph can help you set this up.

exo-go/src/process_helpers/index.go
```diff
@@ -0,0 +1,45 @@
+package processHelpers
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os/exec"
+
+	shellwords "github.com/mattn/go-shellwords"
```
kevgo: Great that you use this libary here, totally the right way to do this!

exo-go/src/process_helpers/index.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Start runs the given command in the given dir directory, and returns
+// the pointer to the command, stdout pipe, output buffer and error (if any)
+func Start(command string, dir string) (*exec.Cmd, io.WriteCloser, *bytes.Buffer, error) {
+	commandWords, err := shellwords.Parse(command)
+	if err != nil {
+		return nil, nil, &bytes.Buffer{}, err
+	}
+	cmd := exec.Command(commandWords[0], commandWords[1:]...) // nolint gas
```
kevgo: Why are we disabling the linter here?

exo-go/src/template_helpers/index.go
```diff
@@ -32,6 +33,8 @@ dependencies:
 services:
   public:
   private:
+
+templates:
```
kevgo: I think with the current setup we don't need to store information about the templates here any longer. The template names are the folder names in the `.exosphere` folder, the template URL is the Git remote there. We don't even need to know the URL any longer:
- exo add-template just adds the template if it doesn't exist and aborts with an error if it does
- exo fetch-templates just runs `git pull` in the submodules
- exo remove-template just deletes the Git submodule

exo-go/src/template_helpers/index.go
```diff
@@ -167,3 +170,49 @@ func CreateTmpServiceDir(chosenTemplate string) string {
...
+func FetchTemplates() error {
+	if output, err := processHelpers.Run("git submodule foreach git pull origin master", ""); err != nil {
+		return errors.Wrap(err, fmt.Sprintf("Failed to fetch updates for existing templates: %s\n", output))
+	}
+	return nil
+}
+
+// RemoveTemplate removes the given template from the application
+func RemoveTemplate(templateName, templateDir string) error {
+	if output, err := processHelpers.Run(fmt.Sprintf("git submodule deinit -f %s", templateDir), ""); err != nil {
```
kevgo: The way we call commands right now is cumbersome:
- first we concatenate words into strings using `fmt.Sprintf`
- then in processHelpers we parse the string back into words

Let's provide the command to run directly as as an array of words here. When using varargs, this is intuitive in Go:

```go
... = processHelpers.Run("git", "submodule", "deinit", "-f", templateDir) ...
```

The definition of the Run method:

```go
func Run(command ...string)   // command is an array of strings

File: exo-go/test_helpers/shared_feature_context.go [hunk 27]
```diff
@@ -56,17 +57,20 @@ func SharedFeatureContext(s *godog.Suite) {
 
 	s.Step(`^running "([^"]*)" in the terminal$`, func(command string) error {
 		var err error
-		childOutput, err = run(command)
+		childOutput, err = processHelpers.Run(command, cwd)
```
kevgo: Parse the string into shell words here to feed it as an array into this command.

kevgo (commented):
Very nice, this is getting close! I think if we simplify the feature set as described in the comments, we can simplify the implementation as well, and remove the need to store templates in application.yml altogether.

exo-go/features/fetch_templates/fetch_templates.feature
cheahuychou: The `running` step calls `cmd.CombinedOutput()` which runs the command, waits for the process to finish and returns its combined output. So if that step didn't fail, the process didn't exit with non-zero code. Getting the exit code, as you can see [here](https://github.com/Originate/exosphere/blob/cm-remote-boilr-templates/exo-go/test_helpers/shared_feature_context.go#L152), involves checking the `err` returned by `cmd.Wait()`.  Since we don't call `cmd.Wait()`, we will need to a lot more work to get the exit code. I could try to do that if you want.

exo-go/features/help.feature
cheahuychou: Does it mean that we have to add "." to the help message of every command?

exo-go/features/remove_template/remove_template.feature
cheahuychou: We are no longer storing the template info in `application.yml` so I guess this is irrelevant.

exo-go/src/cmd/add_template.go
cheahuychou: Again, we are no longer storing the template info in application.yml.

exo-go/src/cmd/add_template.go
cheahuychou: If we decide to do this, we need to do this for every command. I think it might not be that user-friendly because say a user runs `exo add-template name url`, all he/she sees is
```
We are about to add a new service template
```

exo-go/src/cmd/remove_template.go
cheahuychou: I already have it setup. I think sometimes it doesn't run, and I'm not sure why.

exo-go/src/process_helpers/index.go
cheahuychou: If we don't disable it, we will get this warning:
```
warning: Subprocess launching with variable.,HIGH,HIGH (gas)
```

exo-go/src/template_helpers/index.go
cheahuychou: For `exo add-template`, I think we should still allow the user to update the git URL for existing template/submodule since it has already been implemented and it is quite cumbersome to do that manually. I agree on the `application.yml` part. The other two commands were already implemented the way you described.

cheahuychou:
@kevgo all comments addressed.

charlierudolph:
@cheahuychou @kevgo thoughts on making the interface
```
exo template add
exo template fetch
exo template remove
```
This is pretty each to do with cobra since its just another level of subcommands. It also puts all the template logic under a single top level subcommand instead of having all three up there. That would this command similar to `git remote`. Just an idea and one that we can switch to later as well.

kevgo:
@cheahuychou good idea, I like it a lot!

cli/features/abbreviations.feature
```diff
@@ -5,13 +5,13 @@ Feature: executing abbreviated Exosphere commands
   I want to be able to use abbreviations for commands
   So that I can use Exosphere without having to type so much.
 
-  Scenario Outline: executing "exo add" with abbreviated commands
-    When executing the abbreviated command "<ADD-COMMAND>" in the terminal
-    Then the full command "exo add" is executed
+  Scenario Outline: executing "exo add-template" with abbreviated commands
```
kevgo: We can omit these abbreviated commands if it helps keep things simpler. The idea didn't really hold up so far and could be added later.

exo-go/features/help.feature
kevgo: Yes, but let's do that in a separate PR

exo-go/features/remove_template/remove_template.feature
kevgo: yes

exo-go/src/cmd/add_template.go
kevgo: It won't print that as well then.

exo-go/src/process_helpers/index.go
kevgo: @charlierudolph the `gas` linter is still in alpha and not ready for production. I suggest we disable it for this project. There are too many places where we disable it without having good reasons.

exo-go/src/template_helpers/index.go
kevgo: The problem is, simply overriding the Git remote of the template causes a ton of edge cases that we rather not deal with. For example, we have to delete the files and clone them. There is Git history stored on the local client (in the `.git` folder). If we start changing the remote of the Git repo to another repository that contains completely different files and history, bad things will happen, and it might not even be possible to pull or push correctly. 

I think it's safer to abort with an error if the template to be added already exists, and force the user to delete the template first and then set it up cleanly with a new URL. 
That's how `git remote` does it too.

exo-go/src/process_helpers/index.go
charlierudolph: This is the only `gas` rule that I know where we disable it. (We do it twice in this file and thats it). And it is for a good reason as thats specifically how we are testing our CLI.

exo-go/src/cmd/add_template.go
```diff
@@ -0,0 +1,54 @@
...
+		if len(args) != 2 {
+			fmt.Println("not enough arguments")
+			os.Exit(1)
+		}
+		fmt.Print("We are about to add a new service template\n")
+		templateName, gitURL := args[0], args[1]
+		templateDir := path.Join(".exosphere", templateName)
+		if osHelpers.DirectoryExists(templateDir) {
+			reader := bufio.NewReader(os.Stdin)
+			query := fmt.Sprintf(`The template "%s" already exists, do you want to update its git URL to %s`, templateName, gitURL)
```
charlierudolph: Thoughts on making this message also include the old url? Ie:
``` 
The template "%s" already exists, do you want to update its git URL from %s to %s
```
If from and to are the same, we could just exit with the message that the template has already been added

exo-go/src/template_helpers/index.go
```diff
@@ -167,3 +168,49 @@ func CreateTmpServiceDir(chosenTemplate string) string {
 	}
 	return serviceTmpDir
 }
+
+// AddTemplate fetches a remote template from GitHub and stores it
+// under templateDir, returns an error if any
+func AddTemplate(gitURL, templateName, templateDir string) error {
+	if output, err := processHelpers.Run("", "git", "submodule", "add", "--name", templateName, gitURL, templateDir); err != nil {
```
charlierudolph, line 175: Why are we leading with an empty string?

exo-go/test_helpers/shared_feature_context.go
```diff
@@ -12,6 +12,8 @@ import (
 
 	"github.com/DATA-DOG/godog"
 	"github.com/DATA-DOG/godog/gherkin"
+	"github.com/Originate/exosphere/exo-go/src/process_helpers"
+	shellwords "github.com/mattn/go-shellwords"
```
charlierudolph: I don't like that this needs to be repeated so much now. Thoughts on adding this into the process helpers and thus we can simplify the code here and in the source as we can pass in the command as a single string

exo-go/test_helpers/create_feature_context.go
```diff
@@ -24,13 +24,8 @@ func CreateFeatureContext(s *godog.Suite) {
...
-		f, err := os.Stat(path.Join(appDir, directory))
-		if err != nil {
-			return errors.Wrap(err, fmt.Sprintf("Failed to get information for the entry %s", directory))
-		}
-		if os.IsNotExist(err) {
-			return fmt.Errorf("%s does not exist", directory)
-		} else if !f.IsDir() {
+		dirPath := path.Join(appDir, directory)
+		if !osHelpers.IsEmpty(dirPath) {
 			return fmt.Errorf("%s is a not a directory", directory)
```
charlierudolph: This error message is wrong now based on what is being tested

exo-go/src/user_input_helpers/index.go
```diff
@@ -47,3 +47,22 @@ func Choose(reader *bufio.Reader, query string, options []string) string {
 	}
 	return options[chosenNumber-1]
 }
+
+// Confirm asks the user to answer "yes" or "no" to the given query
+func Confirm(reader *bufio.Reader, query string) bool {
+	fmt.Printf("%s (y or n): ", query)
+	answer, err := reader.ReadString('\n')
+	if err != nil {
+		log.Fatalf("Failed to reader user input: %s", err)
```
charlierudolph: reader -> read?

exo-go/src/template_helpers/index.go
```diff
@@ -167,3 +168,49 @@ func CreateTmpServiceDir(chosenTemplate string) string {
...
+	if output, err := processHelpers.Run("", "git", "config", "--file=.gitmodules", fmt.Sprintf("submodule.%s.url", templateName), gitURL); err != nil {
+		return errors.Wrap(err, fmt.Sprintf("Failed to update git URL for template %s: %s\n", templateName, output))
+	}
+	if output, err := processHelpers.Run("", "git", "submodule", "sync"); err != nil {
+		return errors.Wrap(err, fmt.Sprintf("Failed to sync the submodule %s: %s\n", templateName, output))
+	}
+	if output, err := processHelpers.Run("", "git", "submodule", "update", "--init", "--recursive", "--remote"); err != nil {
+		return errors.Wrap(err, fmt.Sprintf("Failed to update the submodule %s: %s\n", templateName, output))
+	}
+	return nil
```
charlierudolph: Thoughts on creating a `runSeries` function which takes in an array of commands to run? On error it would wrap the error with
```go
fmt.Sprintf("Command Failed:\nCommand: %s\nOutput:\n%s\n\n'", command, output)
```

exo-go/src/template_helpers/index.go
cheahuychou: The first argument is the path that the command should be run in. Technically, this should be `cwd` but I didn't want to repeat the code below for every function that calls `Run`
```
cwd, err := os.Getwd()
if err != nil {
   log.Fatal("some message")
}
```

exo-go/src/cmd/add_template.go
cheahuychou: Kevin just told me to get rid of this overwriting feature (i.e. just abort when users try to add an existing service). So this is no longer relevant.

exo-go/src/template_helpers/index.go
cheahuychou: I think that's a good idea if we don't want to use customized messages.

cheahuychou:
@kevgo @charlierudolph all comments addressed

cli/src/cli.ls
```diff
@@ -73,16 +74,16 @@ function print-usage
...
+    * add             Add a service to an existing application
+    * clean           Remove dangling Docker images and volumes
+    * clone           Download the source code of an application
+    * create          Create a new application or stand-alone service
+    * deploy          Deploy an application to the cloud
+    * lint            Verify the correctness of an application
+    * run             Run an application locally
+    * setup           Prepare a freshly cloned application for running it
+    * sync            Download updates for an application from its Git repository
+    * test            Run the tests for an application or service
```
charlierudolph: Doesn't `template` needed to be added here?

exo-go/src/template_helpers/index.go
charlierudolph: Thoughts on making the API of Run / Start back to `(command, dir string)`? (thats what it was in the test helpers when it used shellwords internally). Why was it changed?

I don't like that the dir looks as if its part of the command. I'd prefer the command either be a string or an array of strings. (instead of multiple string arguments)

@kevgo thoughts on this? I don't think the variadic arguments works well when we have another string argument to start.

charlierudolph (approved):
After adding update to the CLI doc

Issue #313, opened 2017-07-06
new "exo run" command


kevgo:
> When working on an Exosphere application
I want to be able to boot it up on my machine
So that I can see the code I am working on in action.

#### Rules
- running "exo run" boots up the application
- before running, the code base is prepared to run, for example by installing dependencies of the services
- when the application is running, it displays "application is online" in the terminal
- the terminal shows the output of the different services, each line preceded by the service name
- Ctrl-C stops all services

#### Notes
- in the old implementation, the setup steps are defined in `service.yml` and in the Dockerfile for the service. In the new implementation, they should be only defined in the Dockerfile.
- running "exo setup" is no longer a requirement for running this command. It should do the necessary setup on its own

```cucumber
Scenario: running a freshly cloned application
  Given I am in the root directory of a freshly cloned Exosphere application
  When running "exo run"
  And waiting it displays "application is running"
  Then my application is running

cheahuychou:
closed in #364

PR #327, opened 2017-07-11
port `exo-run` to Go

cheahuychou:
Port the current implementation of `exo-run` to Go, still need to implement the file watcher and app restarter in a separate PR. @charlierudolph @kevgo I have a few questions for you two, please see the comments below.


exo-go/features/run/run.feature
```diff
@@ -0,0 +1,66 @@
...
+
+
+  Scenario: booting a complex Exosphere application
+    Given a running "running" application
+    Then my machine is running ExoCom
+    And my machine is running the services:
+      | NAME  |
+      | web   |
+      | users |
+    # And ExoCom uses this routing:
```
cheahuychou, line 26: This is commented out because the `yaml` library in Go cannot read the following section in `docker-compose.yml` because of `>-`
```
SERVICE_ROUTES: >-
   [{"role":"web","receives":["users.listed","users.created"],"sends":["users.list","users.create"]},{"role":"users","receives":["mongo.list","mongo.create"],"sends":["mongo.listed","mongo.created"],"namespace":"mongo"}]
```
since part of `exo-setup` will be ported Go soon in response to #313 and that strip shouldn't be there anymore, I think we can leave this out for now.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+func (appRunner *AppRunner) Shutdown(closeMessage, errorMessage string) {
+	var exitCode int
+	if len(errorMessage) > 0 {
+		color.Red(errorMessage)
+		exitCode = 1
+	} else {
+		fmt.Printf("\n\n%s", closeMessage)
+		exitCode = 0
+	}
+	cmd, stdoutBuffer, err := dockerCompose.KillAllContainers([]string{}, appRunner.DockerConfigLocation, appRunner.Write)
```
cheahuychou, line 86: For some reason, I don't see the output from `docker-compose down` in the log but it looks like the containers do get killed correctly.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+				for role, onlineText := range appRunner.OnlineTexts {
+					wg.Add(1)
+					go func(role string, stdoutBuffer fmt.Stringer, onlineText string) {
+						defer wg.Done()
+						processHelpers.Wait(stdoutBuffer, onlineText, func() {
+							appRunner.Logger.Log(role, fmt.Sprintf("'%s' is running", role), true)
+						})
+					}(role, stdoutBuffer, onlineText)
+				}
+				wg.Wait()
```
cheahuychou, line 69: I'm not sure if there is anything wrong with the way `WaitGroup` is setup here but based on my observation, it takes longer to start the app in this implementation than in the original livescript implementation. Also, the log appears to be printed out all at once.

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,124 @@
...
+func (logger *Logger) Log(role, text string, trim bool) {
+	if trim {
+		text = strings.TrimSpace(text)
+	}
+	for _, line := range strings.Split(text, `\n`) {
+		left, right := parseLine(role, line)
+		if !util.DoesStringArrayContain(logger.SilencedRoles, left) {
+			output := fmt.Sprintf("%s %s", logger.pad(left), right)
+			if color, exists := logger.getColor(left); exists {
+				color(output)
```
cheahuychou: We used to make `left` bold but I don't think there is a neat way to do this in Go. Is it ok if we don't make it bold?

exo-go/src/process_helpers/index.go
```diff
@@ -0,0 +1,57 @@
...
+	"os/exec"
+	"strings"
+	"time"
+
+	shellwords "github.com/mattn/go-shellwords"
+)
+
+// Run runs the given command, waits for the process to finish and
+// returns the output string and error (if any)
+func Run(command string) (string, error) {
```
cheahuychou: These functions and the relevant code will be updated to [this](https://github.com/Originate/exosphere/blob/1f77a17ca4e1a3fd7bbb1663563734bf52dbf719/exo-go/src/process_helpers/index.go) once #322 is merged.

exo-go/src/app_runner/index.go
hugoesthere: `docker-compose down` usually prints to the same stream that was printing `docker-compose up` output. It might be enough just to check for a non-zero exit code.

exo-go/src/logger/index.go
charlierudolph: > but I don't think there is a neat way to do this in Go
Why? the color library we are using can has a Bold. https://github.com/fatih/color#mix-and-reuse-colors

cheahuychou: Yes, but we have to make a new color. I guess that's fine. I'll change this.

exo-go/src/docker_helpers/index.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+	}
+	return nil
+}
+
+// PullImage pulls the given image from DockerHub
+func PullImage(c *client.Client, image string, done func(error)) {
+	ctx := context.Background()
+	_, err := c.ImagePull(ctx, image, types.ImagePullOptions{})
+	done(err)
```
charlierudolph: Lets avoid callbacks unless absolutely necessary. Go is mostly a synchronous language and uses goroutines when we need to perform things in parallel. It also isn't needed in this case where you could just return an error.

exo-go/features/run/app-dependency.feature
```diff
@@ -0,0 +1,23 @@
+Feature: running Exosphere applications with dependencies
```
kevgo, line 1: Feature: application dependencies

exo-go/features/run/app-dependency.feature
```diff
@@ -0,0 +1,23 @@
...
+
+  As an Exosphere developer
+  I want to have an easy way to run application wide dependencies
+  So that all my services can use these dependencies.
+
+  Rules:
+  - run "exo run" in the directory of your application to run it
```
kevgo, line 8: This is not a rule for the "application-wide dependencies" feature. The rules are:

- list application dependencies in the "dependencies" key of "application.yml"
- these dependencies get booted up before the application services

exo-go/features/run/app-dependency.feature
```diff
@@ -0,0 +1,23 @@
...
+  - this command boots up all the dependencies of the application
+
+
+  Scenario: booting an Exosphere application with Exocom as a listed dependency
```
kevgo, line 12: Scenario: booting an application that uses Exocom

exo-go/features/run/app-dependency.feature
```diff
@@ -0,0 +1,23 @@
...
+    Given a running "simple" application
+    Then my machine is running the services:
+      | NAME         |
+      | exocom0.22.1 |
+
+
+  Scenario: booting an Exosphere application with Nats as a listed dependency
```
kevgo, line 19: Scenario: booting an application that uses NATS

exo-go/features/run/run.feature
```diff
@@ -0,0 +1,66 @@
...
+  I want to have an easy way to run an application on my developer machine
+  So that I can test my app locally.
+
+  Rules:
+  - run "exo run" in the directory of your application to run it
+  - this command boots up all the services of the application
+
+
+  Scenario: booting a functioning Exosphere application
+    Given a running "simple" application
```
kevgo, line 13: We need to describe how running an app works here, since this is the happy path spec for this feature.

```cucumber
Given I am in the root directory of the "running" example application
When starting "exo run"
Then I see "application is running"
And my machine is running the services:
  | NAME  |
  | web   |
  | users |

File: exo-go/features/run/run.feature [hunk 12]
```diff
@@ -0,0 +1,66 @@
...
+    Then my machine is running the services:
+      | NAME |
+      | web  |
+
+
+  Scenario: booting a complex Exosphere application
```
kevgo, line 19: This example can go away, I don't see the point of it.

exo-go/features/run/run.feature
kevgo: Let's make Exocom support a separate feature in its own file.
Also, YAML is just a superset of JSON, so to avoid such crazy YAML constructs, simply use JSON syntax on this line.

exo-go/features/run/run.feature
```diff
@@ -0,0 +1,66 @@
...
+    #   | ROLE  | SENDS                       | RECEIVES                    | NAMESPACE |
+    #   | web   | users.list, users.create    | users.listed, users.created |           |
+    #   | users | mongo.listed, mongo.created | mongo.list, mongo.create    | mongo     |
+    When starting "exo run" in my application directory
+    When the web service broadcasts a "users.list" message
+    Then the "mongo" service receives a "mongo.list" message
+    And the "web" service receives a "users.listed" message
+
+
+  Scenario: booting an Exosphere application with external dependencies
```
kevgo, line 36: This is a separate feature as well, please extract into its own file.

exo-go/features/run/run.feature
```diff
@@ -0,0 +1,66 @@
...
+    Given a running "external-dependency" application
+    Then my machine is running the services:
+      | NAME  |
+      | mongo |
+
+
+  Scenario: a service crashes during startup
```
kevgo, line 43: This is also its own feature

exo-go/features/run/run.feature
```diff
@@ -0,0 +1,66 @@
...
+    When trying to run the "crashing-service" application
+    Then it prints "crasher exited with code 1" in the terminal
+
+
+  Scenario: hiding logs from certain services or dependencies
```
kevgo, line 48: This is a separate feature

exo-go/features/run/run.feature
```diff
@@ -0,0 +1,66 @@
...
+      | web          |
+      | users        |
+      | exocom0.22.1 |
+    And it prints "'web' is running" in the terminal
+    And it prints "all services online" in the terminal
+    And it does not print "'users' is running" in the terminal
+    And it does not print "'exocom' is running" in the terminal
+
+
+  Scenario: booting an Exosphere application with external services
```
kevgo, line 61: The name should be "...with external __docker images__". Also a separate feature.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return appRunner
+}
+
+// Start runs the application
+func (appRunner *AppRunner) Start() {
+	_, stdoutBuffer, err := dockerCompose.RunAllImages(appRunner.Env, appRunner.DockerConfigLocation, appRunner.Write)
+	if err != nil {
+		appRunner.Shutdown("", "Failed to run images")
+	} else {
```
kevgo, line 54: Return here to avoid the `else` block and reduce nesting.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+		appRunner.compileOnlineText(func(err error) {
+			if err != nil {
+				log.Fatal(err)
+			} else {
```
kevgo, line 58: No `else` necessary here, `log.Fatal` ends the app.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+		fmt.Printf("\n\n%s", closeMessage)
+		exitCode = 0
+	}
+	cmd, stdoutBuffer, err := dockerCompose.KillAllContainers([]string{}, appRunner.DockerConfigLocation, appRunner.Write)
+	if err != nil {
+		log.Fatalf("Failed to start killing all containers.\nOutput:%s\nError:%s", stdoutBuffer.String(), err)
+	}
+	if err := cmd.Wait(); err != nil {
+		log.Fatalf("Failed to kill all containers.\nOutput:%s\nError:%s", stdoutBuffer.String(), err)
+	} else {
```
kevgo, line 92: remove the else

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+func (appRunner *AppRunner) Write(text string) {
+	appRunner.Logger.Log("exo-run", text, true)
+}
+
+func (appRunner *AppRunner) compileOnlineText(done func(error)) {
+	appRunner.OnlineTexts = make(map[string]string)
+	for _, dependency := range appRunner.AppConfig.Dependencies {
+		appRunner.OnlineTexts[dependency.Name] = dependency.GetOnlineText()
+	}
+	wg := new(sync.WaitGroup)
```
kevgo, line 107: Why are you waiting here? The function is called `compileOnlineText`, all it should do is return the compiled online texts.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	for service, serviceData := range serviceHelpers.GetServiceConfigs(appRunner.AppConfig.Services) {
+		wg.Add(1)
+		go func(service string, serviceData types.ServiceConfig) {
+			defer wg.Done()
```
kevgo, line 111: Do we really need to defer here? Why can't you simply call `wg.Done()` at the end of this go function? If it returns an error, it didn't see the online text, so shouldn't call `Done`.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+			appRunner.getOnlineText(service, serviceData, func(err error) {
+				if err != nil {
+					done(err)
+				}
+			})
+		}(service, serviceData)
+	}
+	wg.Wait()
```
kevgo, line 119: Why do you wait here and in the `Start` method?

exo-go/src/cmd/clean.go
```diff
@@ -23,56 +21,19 @@ var cleanCmd = &cobra.Command{
 		if err != nil {
 			panic(err)
 		}
-		err = removeImages(c)
+		err = dockerHelpers.RemoveDanglingImages(c)
```
kevgo, line 24: This file contains unrelated changes, please extract them into their own PR.

exo-go/src/cmd/run.go
```diff
@@ -0,0 +1,48 @@
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			c := make(chan os.Signal, 1)
+			signal.Notify(c, os.Interrupt)
+			<-c
+			appRunner.Shutdown(" shutting down ...", "")
+		}()
+		appRunner.Start()
+		wg.Wait()
```
kevgo, line 42: Why can't you read the channel here directly? I don't think you need a go-function here at all.

exo-go/src/docker_compose/index.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"os/exec"
+)
+
+// RunAllImages runs all the docker images
+func RunAllImages(env []string, cwd string, write func(string)) (*exec.Cmd, *bytes.Buffer, error) {
```
kevgo, line 11: Can you use https://github.com/docker/libcompose instead of writing your own Go driver for Docker Compose?

exo-go/src/docker_helpers/index.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// CatFile reads the file fileName inside the docker image image
+func CatFile(c *client.Client, image, fileName string, done func(err error, content []byte)) {
```
kevgo: Better name: `CatFileInDockerImage`

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,124 @@
...
+	SilencedRoles []string
+	Length        int
+	Colors        map[string]func(string, ...interface{})
+}
+
+// NewLogger is Logger's constructor
+func NewLogger(roles, silencedRoles []string) *Logger {
+	logger := &Logger{Roles: roles, SilencedRoles: silencedRoles}
+	logger.Colors = map[string]func(string, ...interface{}){
+		"exocom":     color.Cyan,
```
kevgo: Why do we hard-code colors here? Let's simply assign the first color to the first role, the second color to the second role, etc.

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,124 @@
...
+
+func parseService(text string) string {
+	return stripColor(strip(`(\d+\.)?(\d+\.)?(\*|\d+)$`, text))
+}
+
+func reformatLine(line string) string {
+	return strings.TrimSpace(stripColor(line))
+}
+
+func padLeft(text string, length int, padding rune) string {
```
kevgo, line 117: Everytime you find yourself writing a method that is generic as this one, try to find an existing library first. Here is one that does the same as your code, but is tested: https://github.com/willf/pad

exo-go/src/process_helpers/index.go
```diff
@@ -38,6 +40,17 @@ func Start(dir string, commandWords ...string) (*exec.Cmd, io.WriteCloser, *byte
 	return cmd, in, &out, nil
 }
 
+// Wait waits for the given text for the specified duration
+func Wait(stdout fmt.Stringer, text string, done func()) {
+	ticker := time.NewTicker(100 * time.Millisecond)
+	var output string
+	for !strings.Contains(output, text) {
```
kevgo, line 47: Here might be a reason why the Go version is slower. The Node version evaluates the text right when it is written into the stream, and calls the callback when it finds the text right away. Your method here waits 100ms, then checks, then waits another 100ms. These delays can add up.

kevgo (commented):
Nice job so far! Please try doing way smaller PRs. This PR is way too big, and does too many things at once. It should be split up into 10 PRs or so. Also, please don't do callbacks, as Charlie has pointed out.

kevgo:
I suggest you split up the features into separate files, as described. Then make one PR that only implements the happy path feature. Then another in a child branch that adds application dependencies. Then another that adds external dependencies. This is so big, it is almost impossible to review and make sure it is correctly implemented, and it is hard giving you detailed enough feedback given this vast lump of changes here. Please follow this blog post: http://blog.originate.com/blog/2017/04/28/high-velocity-development-using-nested-feature-branches/

exo-go/src/cmd/clean.go
charlierudolph: Theses are related as she extracted some functions from this file and moved them into a helper function as she needed them somewhere else

cheahuychou:
@kevgo I've split this PR into 5 PRs but I did it based on what the code does instead because I think what you suggest is infeasible without a lot of rewriting and reorganization. #329 and #330 are ready for review. I'm still working on the other three. I'm very sorry that this PR is confusing/hard to review. I'm already not implementing the entire `exo-run` in this PR. Also, it wasn't clear to me how this branch could be split until I was almost done with it.

exo-go/features/run/app-dependency.feature
cheahuychou: I fixed this in #333  but just wanted to point out that I got all of these from the original feature tests.

exo-go/features/run/run.feature
cheahuychou: > YAML is just a superset of JSON, so to avoid such crazy YAML constructs, simply use JSON syntax on this line.

I got rid of this test in #333 but to address this comment, we might need to change how `exo-setup` (or `exo-run` soon) renders `SERVICE_ROUTES`.
@hugobho am I right?

exo-go/src/cmd/run.go
cheahuychou: Without the `Wait`, the process exits after `appRunner.Start()` finishes.

exo-go/src/docker_compose/index.go
cheahuychou: I saw that but I don't know if there is a way to get the command output (for the logger) if we use this. The repo doesn't have very useful examples but I think we'd do something like [this](https://github.com/docker/libcompose/blob/master/example/main.go#L15). Based on the [struct of `options.Up`](https://godoc.org/github.com/docker/libcompose/project/options#Up), I don't know if there is a way to pass in environment variables. Also, we are using `docker-compose` version 3 and it looks like that they don't have support for that yet (see this [issue](https://github.com/docker/libcompose/issues/474))

kevgo:
@cheahuychou no worries, just some comments to help improve this PR  :)

exo-go/features/run/app-dependency.feature
kevgo: Yes, I am aware of that and don't hold it against you. We might have experienced some technical drift here. Since you reimplement the tests in Go, this is a good chance to also fine-tune the wording.

exo-go/src/cmd/clean.go
kevgo: Okay. These changes could have been done separately. No need to change it here now, just pointing out opportunities to trim down this PR.

exo-go/src/cmd/run.go
kevgo: If I read your code correctly, the main thread waits for the goroutine. The goroutine waits for the channel. So why can't the main thread wait for the channel directly?

exo-go/src/docker_compose/index.go
kevgo: Okay, good find! We definitely want to use Docker Compose version 3. That might be the killer argument here.

charlierudolph:
@cheahuychou since you broke this out, can this PR be closed?

exo-go/features/run/run.feature
hugoesthere: @kevgo It looks like the `SERVICE_ROUTES` is already in JSON. Are you talking about `>-`? Because that just indicates that the next line is a string. This isn't something we write but is generated by `docker-compose`.

PR #329, opened 2017-07-12
[exo-run part 1] write logger in Go

cheahuychou:


exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (logger *Logger) logOutput(left, right string) {
+	if printColor, exists := logger.getColor(left); exists {
+		color.Set(printColor, color.Bold)
+		fmt.Printf("%s ", logger.pad(left))
+		color.Unset()
+		color.Set(printColor)
+		fmt.Println(right)
+		color.Unset()
```
charlierudolph: I'd prefer we use https://github.com/fatih/color#insert-into-noncolor-strings-sprintfunc instead of Set / Unset since those have global effects versus just colorizing a particular string

charlierudolph (commented):
We had some unit tests for this class that I'd like us to port over. I'll commit to this branch to give you the basic setup and then you can fill in the tests. Golang is better at protecting private methods so I believe we will need to make the logger more configurable in order to properly tests it. Ideally we could pass in a stream that we write to (where we create a stream in the tests and pass in os.Stdout in the source)

charlierudolph:
Added unit testing framework. See the docs [here](http://onsi.github.io/ginkgo/)

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+	Roles         []string
+	SilencedRoles []string
+	Length        int
+	Colors        map[string]color.Attribute
+}
+
+// NewLogger is Logger's constructor
+func NewLogger(roles, silencedRoles []string) *Logger {
+	logger := &Logger{Roles: roles, SilencedRoles: silencedRoles}
+	commands := []string{"exocom", "exo-run", "exo-clone", "exo-setup", "exo-test", "exo-sync", "exo-lint", "exo-deploy"}
```
kevgo: I'm not sure it's the logger's responsibility to know the Exosphere commands. The logger should just be a generic logger, we should almost make it a separate open-source project. The Exophere binary that uses the logger should provide a list of roles to the logger including the exosphere commands.

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+	logger.Colors = map[string]color.Attribute{}
+	logger.setColors(roles)
+	logger.setLength(append(roles, commands...))
+	return logger
+}
+
+func (logger *Logger) getColor(role string) (color.Attribute, bool) {
+	chosenColor, exists := logger.Colors[role]
```
kevgo: Simplify to `return logger.Colors[role]`

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+	return chosenColor, exists
+}
+
+func (logger *Logger) setColors(roles []string) {
+	defaultColors := []color.Attribute{color.FgMagenta, color.FgBlue, color.FgYellow, color.FgCyan}
```
kevgo: Extract this into a constant that is defined in this file.

```go
const defaultColors = []color.Attribute{color.FgMagenta, color.FgBlue, color.FgYellow, color.FgCyan}

func (logger *Logger) setColors(roles []string) {
  for i, role := range roles {
    ...
 ```

If that doesn't work, then extract a `defaultColors` method that returns the array.

File: exo-go/src/logger/index.go [hunk 5]
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+}
+
+func (logger *Logger) pad(text string) string {
+	return pad.Left(text, logger.Length, " ")
+}
+
+func (logger *Logger) logOutput(left, right string) {
+	if printColor, exists := logger.getColor(left); exists {
```
kevgo: Please use the [guard clauses](https://sourcemaking.com/refactoring/replace-nested-conditional-with-guard-clauses) to avoid unnecessary nesting of code. Nesting increases the cyclomatic complexity and makes code harder to read.

```go
if printColor, exists := logger.getColor(left); !exists {
  log.Fatalf("%s %s\n", logger.pad(left), right)
}
...

File: exo-go/src/logger/index.go (hunk 1 above)
Comment (kevgo) [reply exo-go/src/logger/index.go]: +1, please keep in mind that Go is a concurrent language, so there might be multiple threads writing in parallel and we want to avoid things that use global shared state.

File: exo-go/src/logger/index.go [hunk 6]
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+// Log logs the given text
+func (logger *Logger) Log(role, text string, trim bool) {
+	if trim {
+		text = strings.TrimSpace(text)
+	}
+	for _, line := range strings.Split(text, `\n`) {
+		left, right := parseLine(role, line)
```
kevgo: Better names for these variables than `left` and `right`: `serviceName` and `serviceOutput`

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+		if !util.DoesStringArrayContain(logger.SilencedRoles, left) {
+			logger.logOutput(left, right)
+		}
+	}
+}
+
+// Error logs the given error text
+func (logger *Logger) Error(role, text string, trim bool) {
```
kevgo: Please sort methods and functions alphabetically, so that they are easier to find using binary search!  :)

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+	if trim {
+		text = strings.TrimSpace(text)
+	}
+	for _, line := range strings.Split(text, `\n`) {
+		left, right := parseLine(role, line)
+		if !util.DoesStringArrayContain(logger.SilencedRoles, left) {
+			logger.logOutput(left, right)
```
kevgo: How is this method different from `Log`?

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+}
+
+func parseLine(role, line string) (string, string) {
```
kevgo: Parsing DockerCompose output is a different responsibility than logging text to the console. The logger is a lot more generic and can be used in many apps, parsing DockerCompose only makes sense in this app. Please extract this function into its own file and add unit tests for it.

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+	segments := []string{}
+	for _, segment := range regexp.MustCompile(`\s+\|\s*`).Split(line, -1) {
```
kevgo: This compiles the Regex for every line that is logged. Compiling Regex is slow and memory intensive. Please make the regex a constant that is compiled at the beginning of the application and then reused in the `parseLine` method.

kevgo: Can the regex be simplified to `.+|.*`?

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+		segments = append(segments, strings.TrimSpace(segment))
+	}
+	if len(segments) == 2 {
+		service := parseService(segments[0])
+		return service, reformatLine(segments[1])
```
kevgo: Inline the `service` variable here. Too many temp variables are a code smell.

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	return role, line
```
kevgo: Is there ever legit output that does not contain `|`?

exo-go/src/logger/index.go
cheahuychou: I agree and I did try to avoid that (see the code below) but the entire line was in bold because `color.Bold` actually affects the entire line. That's why I need line 59 to undo that effect. That was also why I said there wasn't a neat way to make the left text bold.
```	
if printColor, exists := logger.getColor(left); exists {
   boldColor := printColor.Add(color.Bold)
  fmt.Printf("%s %s\n", boldColor.Sprintf(logger.pad(left)), printColor.Sprintf(right))
}
```

charlierudolph: This line mutates printColor and doesn't create a new color.
```go
boldColor := printColor.Add(color.Bold)
```
It returns itself to make chaining easier.
The bold version would need to copy the printColor or something first

exo-go/src/logger/index.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func parseService(text string) string {
```
kevgo: This method needs unit tests

exo-go/src/logger/index_test.go
```diff
@@ -0,0 +1,14 @@
+package logger_test
+
+import (
+	"github.com/Originate/exosphere/exo-go/src/logger"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Logger", func() {
```
kevgo: I don't think we can test the logger particularly well, but let's test the logic that parses Docker Compose output.

kevgo (commented):
Great progress! Let's improve the factoring of the code base as commented below.

charlierudolph:
Please don't vendor ginkgo and gomega, they are installed via setup as we need a ginkgo binary. Anytime we need a binary we need to install it with `go get` instead of putting it in the vendor folder

cheahuychou:
@kevgo @charlierudolph all comments addressed

charlierudolph (approved):

cheahuychou:
closed in: https://github.com/Originate/exosphere/commit/1858b0bf8540ad8d01e0f3bd7e3c2397331cc5c0

PR #330, opened 2017-07-12
update config structs and helpers

cheahuychou:


exo-go/src/app_config_helpers/index.go
```diff
@@ -50,3 +50,14 @@ func UpdateAppConfig(reader *bufio.Reader, serviceRole string, appConfig types.A
 		log.Fatalf("Failed to write application.yml: %s", err)
 	}
 }
+
+// GetSilencedDependencies returns a slice of silenced application dependenices
```
charlierudolph, line 54: Maybe mention in here that its the names of the dependencies versus some dependency object

exo-go/src/service_config_helpers/index.go
```diff
@@ -0,0 +1,26 @@
...
+	"path"
+
+	yaml "gopkg.in/yaml.v2"
+
+	"github.com/Originate/exosphere/exo-go/src/types"
+)
+
+// GetServiceConfig reads service.yml of the role service and returns
+// the serviceConfig object
+func GetServiceConfig(location string) types.ServiceConfig {
```
charlierudolph, line 15: Maybe rename `location` -> `serviceFilePath`? Based on the code I believe the location is the root directory of the service and ideally the name is able to reflect that

exo-go/src/service_config_helpers/index.go
```diff
@@ -0,0 +1,26 @@
...
+	yamlFile, err := ioutil.ReadFile(path.Join(location, "service.yml"))
+	if err != nil {
+		log.Fatalf("Failed to read service.yml: %s", err)
```
charlierudolph, line 18: Please return an error instead of using log.Fatal. Same below. 

I also don't think its worth wrapping this error. Since we pass in a file name that includes `service.yml` I expect the error message would give us enough detail. For unmarshalling the yaml, I think the error is useful since that error message would not give us the context about what is being unmarshalled.

exo-go/src/service_config_helpers/index.go
cheahuychou: `location` refers to the name of the service folder. It aligns with the [key](https://github.com/Originate/exosphere/blob/master/exosphere-shared/example-apps/running/application.yml#L12) in `application.yml`. It's not really a service file path since we do `path.Join(location, "service.yml")` within the function.

charlierudolph: Okay. Good as is then

exo-go/src/app_config_helpers/index.go
```diff
@@ -50,3 +50,14 @@ func UpdateAppConfig(reader *bufio.Reader, serviceRole string, appConfig types.A
...
+func GetSilencedDependencies(appConfig types.AppConfig) []string {
```
kevgo, line 55: Better name: `GetSilencedDependencyNames`

exo-go/src/service_config_helpers/index.go
kevgo: What about `ServiceFolderName`?

exo-go/src/service_helpers/index.go
```diff
@@ -27,3 +27,31 @@ func GetExistingServices(services types.Services) []string {
 	}
 	return existingServices
 }
+
+// GetServiceConfigs returns a map that maps each service to its serviceConfig
```
kevgo, line 31: // GetServiceConfigs returns the configurations data for the given services

exo-go/src/service_helpers/index.go
```diff
@@ -27,3 +27,31 @@ func GetExistingServices(services types.Services) []string {
...
+	for service, serviceData := range services.Private {
+		serviceConfigs[service] = serviceData
+	}
+	for service, serviceData := range services.Public {
+		serviceConfigs[service] = serviceData
+	}
+	return serviceConfigs
+}
+
+// GetSilencedServices returns a slice of silenced services
```
kevgo, line 43: The reader knows that this function returns a slice, so no point in repeating this in the comment. 

// GetSilencedServices returns the services that are configured as silent.

exo-go/src/types/dependency.go
```diff
@@ -0,0 +1,57 @@
...
+	case "nats":
+		return map[string]interface{}{"NATS_HOST": dependency.GetContainerName()}
+	default:
+		return dependency.Config.DependencyEnvironment
+	}
+}
+
+// GetContainerName returns the container name for the dependency
+func (dependency *Dependency) GetContainerName() string {
+	return fmt.Sprintf("%s%s", dependency.Name, dependency.Version)
```
kevgo, line 44: Simplify to `return dependency.Name + dependency.Version`

exo-go/src/types/dependency.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// GetOnlineText returns the online text for the dependency
+func (dependency *Dependency) GetOnlineText() string {
```
kevgo, line 48: This is good for now, but let's talk how we can do this more Exosphere-idiomatic. Can Exocom and Nats be be services that have a service.yml file that defines this value? @charlierudolph

exo-go/src/types/index.go
```diff
@@ -1,28 +1,23 @@
...
+	Type        string            `yaml:",omitempty"`
+	Description string            `yaml:",omitempty"`
+	Author      string            `yaml:",omitempty"`
+	Setup       string            `yaml:",omitempty"`
+	Startup     map[string]string `yaml:",omitempty"`
+	Messages    `yaml:",omitempty"`
+}
+
+// Messages represents the messages that the service sends and receives
+type Messages struct {
```
kevgo, line 18: Name this `ServiceMessages`?

PR #331, opened 2017-07-12
[exo-run part 3] add docker compose struct and process helper functions

cheahuychou:


cheahuychou:
closing for now since it's not ready for review yet

cheahuychou:
this PR is ready for review

exo-go/src/docker_compose/index.go
```diff
@@ -0,0 +1,43 @@
...
+func RunAllImages(env []string, cwd string, log func(string)) (*exec.Cmd, *bytes.Buffer, error) {
+	return start(cwd, env, log, "docker-compose", "up")
+}
+
+// KillAllContainers kills all the containers
+func KillAllContainers(env []string, cwd string, log func(string)) (*exec.Cmd, *bytes.Buffer, error) {
+	return start(cwd, env, log, "docker-compose", "down")
+}
+
+func start(dir string, env []string, log func(string), command ...string) (*exec.Cmd, *bytes.Buffer, error) {
```
cheahuychou: I had to write this function although it is kind of similar to `processHelpers.Start` because I want to be able to pass in environment variables and the log function so that I can log the output as the process is run. Also, I had to use both stdout pipe and a std buffer because I want to store the output so we can wait for online texts on top of doing the logging. I'm not very familiar with Go yet so need suggestions for a better way to do this.

exo-go/src/docker_helpers/index.go
```diff
@@ -17,9 +18,60 @@ func CatFileInDockerImage(c *client.Client, image, fileName string) ([]byte, err
 	return []byte(output), err
 }
 
+// ListRunningContainers passes a slice of the names of running containers
+// and error (if any) to the callback function
+func ListRunningContainers(c *client.Client, done func([]string, error)) {
```
cheahuychou: I had to use a callback here because this [step](https://github.com/Originate/exosphere/blob/cm-5-exo-run-cmd/exo-go/test_helpers/run_feature_context.go#L49) kept failing without the callback. I printed out the containers array, and saw the array growing as more tests with this step are run. This is very strange.

charlierudolph: I think we need to cleanup the running containers after each step then or do better filtering in the step. The callback shouldn't change anything aside from preventing that step from being able to return an error

exo-go/src/docker_compose/index.go
charlierudolph: We could make our own struct to wrap `exec.Command` with the useful functions we need. That way we can setup all the needed parameters before starting by calling methods on the structs instead of having to put it into a single function call. It could look something like this
```go
process = processHelpers.NewProcess("arg1", "arg2", "arg3")
process.SetDir(dir)
process.SetEnv(env)
process.SetStdoutLog(func(string) {})
process.Start()
process.GetCmd() // if you need the underlying cmd
```

exo-go/src/process_helpers/process.go
```diff
@@ -0,0 +1,113 @@
...
+
+// NewProcess is Process's constructor
+func NewProcess(commandWords ...string) *Process {
+	process := &Process{Cmd: exec.Command(commandWords[0], commandWords[1:]...)} //nolint gas
+	return process
+}
+
+// Log reads the stream from the given stdPipeReader, logs the
+// output, and update process.Output
+func (process *Process) Log(stdPipeReader io.Reader) {
```
charlierudolph: I think this method should be private as I don't expected anyone to call it directly. Its also only used internally in Start

charlierudolph (approved):

cheahuychou:

cheahuychou:
closed in: https://github.com/Originate/exosphere/commit/b1b7f91d0486971dc0d556c51a4419a4d4f48b57

PR #332, opened 2017-07-12
[exo-run part 4] write app runner in Go

cheahuychou:


exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+	return appRunner
+}
+
+// Start runs the application
+func (appRunner *AppRunner) Start() {
+	_, stdoutBuffer, err := dockerCompose.RunAllImages(appRunner.Env, appRunner.DockerConfigLocation, appRunner.Write)
+	if err != nil {
+		appRunner.Shutdown("", "Failed to run images")
```
charlierudolph: Thoughts on letting the parent perform the shutdown and returning an error in this case? We also lose whatever error message docker is providing.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	} else {
+		appRunner.compileOnlineText(func(err error) {
+			if err != nil {
+				log.Fatal(err)
```
charlierudolph: Since these are helper functions lets return errors instead of using log.Fatal

exo-go/src/process_helpers/index.go
```diff
@@ -38,6 +40,17 @@ func Start(dir string, commandWords ...string) (*exec.Cmd, io.WriteCloser, *byte
 	return cmd, in, &out, nil
 }
 
+// Wait waits for the given text for the specified duration
+func Wait(stdout fmt.Stringer, text string, done func()) {
```
charlierudolph: I think it would be better if we use pipes here. If we have the commands stdout pipe, I believe we can read from it and that will block until data arrives. Thus we shouldn't need to poll and can return right when the data comes in.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+		os.Exit(exitCode)
+	}
+}
+
+// Write logs exo-run output
+func (appRunner *AppRunner) Write(text string) {
+	appRunner.Logger.Log("exo-run", text, true)
+}
+
+func (appRunner *AppRunner) compileOnlineText(done func(error)) {
```
charlierudolph: Please remove the callback and return an error. Also thoughts on naming this compileOnlineTexts. I'd like this method to be updated to returning the onlineTexts instead of saving it on the instance

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	cmd, stdoutBuffer, err := dockerCompose.KillAllContainers([]string{}, appRunner.DockerConfigLocation, appRunner.Write)
+	if err != nil {
+		log.Fatalf("Failed to start killing all containers.\nOutput:%s\nError:%s", stdoutBuffer.String(), err)
+	}
+	if err := cmd.Wait(); err != nil {
+		log.Fatalf("Failed to kill all containers.\nOutput:%s\nError:%s", stdoutBuffer.String(), err)
+	} else {
+		os.Exit(exitCode)
+	}
```
charlierudolph: Lets update this to only kill the docker containers and thus return an error if there is one. The command will deal with the exitCode and printing messages.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+					done(err)
+				}
+			})
+		}(service, serviceData)
+	}
+	wg.Wait()
+	done(nil)
+}
+
+func (appRunner *AppRunner) getOnlineText(role string, serviceData types.ServiceConfig, done func(error)) {
```
charlierudolph: Please remove the callback and return an error. Lets also update this method to return the onlineText intead of saving it to the instance

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	Cwd                  string
+	OnlineTexts          map[string]string
+}
+
+// NewAppRunner is AppRunner's constructor
+func NewAppRunner(appConfig types.AppConfig, logger *logger.Logger) *AppRunner {
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("Failed to get the current path: %s", err)
+	}
```
charlierudolph: Lets pass the cwd in instead of defining grabbing it here.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	appRunner := &AppRunner{AppConfig: appConfig, Logger: logger, DockerConfigLocation: path.Join(cwd, "tmp"), Cwd: cwd}
+	for _, dependency := range appConfig.Dependencies {
+		for variable, value := range dependency.GetEnvVariables() {
+			appRunner.Env = append(appRunner.Env, fmt.Sprintf("%s=%s", variable, value))
+		}
+	}
```
charlierudolph: Getting the `Env` seems like it could move into a helper function that returns the env. It appears we only need it in one place anyway. Thus we wouldn't need to save it on instance

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	return appRunner
+}
+
+// Start runs the application
+func (appRunner *AppRunner) Start() {
```
charlierudolph: We should be following the pattern of having the instance being just a single letter (`a` in this case), because it is typically used so often

cheahuychou:
closing for now since it's not ready for review yet

exo-go/src/app_runner/index.go
cheahuychou: > We also lose whatever error message docker is providing.

you meant in the current implementation?
> letting the parent perform the shutdown and returning an error in this case

so make this function return the error and let `run.go` do the shutdown?

charlierudolph: > you meant in the current implementation?

yes

> so make this function return the error and let run.go do the shutdown?

yes

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	OnlineTexts          map[string]string
+}
+
+// NewAppRunner is AppRunner's constructor
+func NewAppRunner(appConfig types.AppConfig, logger *logger.Logger) *AppRunner {
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("Failed to get the current path: %s", err)
+	}
+	appRunner := &AppRunner{AppConfig: appConfig, Logger: logger, DockerConfigLocation: path.Join(cwd, "tmp"), Cwd: cwd}
```
kevgo: Why is the docker config located in the `tmp` directory?

exo-go/src/app_runner/index.go
kevgo: I personally prefer readable variable names. They make the code more self-describing imo.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	for _, dependency := range appConfig.Dependencies {
+		for variable, value := range dependency.GetEnvVariables() {
+			appRunner.Env = append(appRunner.Env, fmt.Sprintf("%s=%s", variable, value))
```
kevgo: I'm not sure providing the env variables to AppRunning as an array of string variables is the right approach here. Let's provide them as a proper Go structure, and let AppRunner then serialized it to whatever it needs.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return appRunner
+}
+
+// Start runs the application
+func (appRunner *AppRunner) Start() {
+	_, stdoutBuffer, err := dockerCompose.RunAllImages(appRunner.Env, appRunner.DockerConfigLocation, appRunner.Write)
+	if err != nil {
+		appRunner.Shutdown("", "Failed to run images")
+	} else {
```
kevgo: In general please use guard clauses for exceptional situations, rather than `else` blocks.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+		appRunner.OnlineTexts[dependency.Name] = dependency.GetOnlineText()
+	}
+	wg := new(sync.WaitGroup)
+	for service, serviceData := range serviceHelpers.GetServiceConfigs(appRunner.AppConfig.Services) {
+		wg.Add(1)
+		go func(service string, serviceData types.ServiceConfig) {
+			defer wg.Done()
+			appRunner.getOnlineText(service, serviceData, func(err error) {
+				if err != nil {
+					done(err)
```
kevgo: This callback stuff seems non-idiomatic. Don't use callbacks in Go. Provide a channel to functions that return results asynchronously. The caller should listen on the channel to resume when it gets the result.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,145 @@
...
+	if len(serviceData.Location) > 0 {
+		serviceConfig := serviceConfigHelpers.GetServiceConfig(serviceData.Location)
+		appRunner.OnlineTexts[role] = serviceConfig.Startup["online-text"]
+		done(nil)
+	} else if len(serviceData.DockerImage) > 0 {
+		c, err := client.NewEnvClient()
+		if err != nil {
+			panic(err)
+		}
+		dockerHelpers.CatFile(c, serviceData.DockerImage, "service.yml", func(err error, yamlFile []byte) {
```
kevgo: In Go there are no async functions. Functions are synchronous and block until they get their results. If you want to wait for a result that happens later, use channels.

exo-go/src/process_helpers/index.go
cheahuychou: You meant something like this?
```
stdout, err := cmd.StdoutPipe()
if err != nil {
    return cmd, &out, err
}
scanner := bufio.NewScanner(stdout)
for scanner.Scan() {
    output := scanner.Text()
    if strings.Contains(output, text) {
       done()
    }
}
```
If so, we will need to modify `processHelpers.Start` and all the tests. One issue with pipes is that there isn't way (that I know of) to get the entire output of the command like `stdoutBuffer.String()`. Once you read it, the output is gone. We could have a global buffer when we run the test and keep writing to the buffer like [here](https://github.com/Originate/exosphere/blob/f3f6f5a6991a78faa27fb362d4c2a9fd76c92975/exo-go/src/docker_compose/index.go#L20) but that's kind of ugly. Or is there a better way?

exo-go/src/app_runner/index.go
hugoesthere: This is where `docker-compose.yml` is located. We're keeping it in a `tmp` dir because it's an auto-generated file and should not be checked into git, as we discussed when first writing this. Maybe this could be renamed to `DockerComposeLocation` or something else that makes this more clear?

cheahuychou:
Reopening this since I've addressed most of the comments and have a few questions about the rest.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,100 @@
...
+	for role, onlineText := range onlineTexts {
+		wg.Add(1)
+		go func(role string, stdoutBuffer fmt.Stringer, onlineText string) {
+			processHelpers.Wait(stdoutBuffer, onlineText, func() {
+				appRunner.Logger.Log(role, fmt.Sprintf("'%s' is running", role), true)
+			})
+			wg.Done()
+		}(role, stdoutBuffer, onlineText)
+	}
+	wg.Wait()
```
cheahuychou: > @kevgo Why do you wait here and in the Start method?

because I need to wait for all online texts to show up before printing "all services online"

exo-go/src/process_helpers/index.go
```diff
@@ -38,6 +40,17 @@ func Start(dir string, commandWords ...string) (*exec.Cmd, io.WriteCloser, *byte
 	return cmd, in, &out, nil
 }
 
+// Wait waits for the given text
+func Wait(stdout fmt.Stringer, text string, done func()) {
+	ticker := time.NewTicker(20 * time.Millisecond)
```
cheahuychou: reduced the tick time, still waiting for @charlierudolph to respond to my questions about how to use pipes.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,100 @@
...
+	if len(errorMessage) > 0 {
+		color.Red(errorMessage)
+	} else {
+		fmt.Printf("\n\n%s", closeMessage)
+	}
+	cmd, _, err := dockerCompose.KillAllContainers([]string{}, appRunner.DockerConfigLocation, appRunner.Write)
+	if err != nil {
+		return err
+	}
+	if err := cmd.Wait(); err != nil {
```
cheahuychou: So far the function always exits with exit code 1 right here. All it does it running `docker-compose down`. So I'm not sure why this is. The output I got is:
```
 shutting down ...exo-run Gracefully stopping... (press Ctrl+C again to force)
[time] Failed to shutdown the app
Output: 
Error: exit status 1
```

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,100 @@
...
+	AppConfig            types.AppConfig
+	Logger               *logger.Logger
+	Env                  map[string]string
+	DockerConfigLocation string
+	OnlineTexts          map[string]string
+}
+
+// NewAppRunner is AppRunner's constructor
+func NewAppRunner(appConfig types.AppConfig, logger *logger.Logger, cwd string) *AppRunner {
+	appRunner := &AppRunner{AppConfig: appConfig, Logger: logger, Env: appConfigHelpers.GetEnvironmentVariables(appConfig), DockerConfigLocation: path.Join(cwd, "tmp")}
```
charlierudolph: Thoughts on styling thins in the following way to make it a little easier to read:
```go
appRunner := &AppRunner{
  AppConfig: appConfig, 
  Logger: logger, 
  Env: appConfigHelpers.GetEnvironmentVariables(appConfig), 
  DockerConfigLocation: path.Join(cwd, "tmp"),
}
```

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	process, err := dockerCompose.KillAllContainers([]string{}, appRunner.DockerConfigLocation, appRunner.Write)
+	if err != nil {
+		return process.Output, err
+	}
+	err = process.Wait()
+	return process.Output, err
+}
+
+// Start runs the application and returns the process output and an error if any
```
charlierudolph: Do you need to return the output? Based on its usage in the next PR, I think we could get away with returning just an error given that we wrap any errors to include the process output. Same for the shutdown method

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	wg.Wait()
+	if err == nil {
+		appRunner.Write("all services online")
+	}
+	return process.Output, err
+}
+
+// Write logs exo-run output
+func (appRunner *AppRunner) Write(text string) {
```
charlierudolph: Does this need to be a public function? It appears to only be used internally.

exo-go/src/app_runner/index.go
```diff
@@ -0,0 +1,100 @@
...
+func (appRunner *AppRunner) getEnv() []string {
+	formattedEnvVars := []string{}
+	for variable, value := range appRunner.Env {
+		formattedEnvVars = append(formattedEnvVars, fmt.Sprintf("%s=%s", variable, value))
+	}
+	return formattedEnvVars
+}
+
+// Shutdown shuts down the application and returns the process output and an error if any
+func (appRunner *AppRunner) Shutdown(closeMessage, errorMessage string) (string, error) {
```
charlierudolph: Thoughts on making the input to this be a struct (maybe named `ShutdownConfig`) which has two properties closeMessage and errorMessage. The current interface is a bit awkward to me as you always need to specify an empty errorMessage or closeMessage which wouldn't be necessary with an object since it would default to an empty string

cheahuychou:

charlierudolph (approved):

cheahuychou:
closed in: https://github.com/Originate/exosphere/commit/dc6ebc44830341b23d4176aab86c8c5f1b9cf3cf

PR #334, opened 2017-07-13
[exo-run part 2] update config structs and helpers

cheahuychou:
This PR contains the changes in #330 and the helper functions that @hugobho also needs. That one got closed because I made a mistake when pushing new commits to its parent branch. All comments addressed.


charlierudolph:
Thoughts on adding some unit tests for this? (After part 1, goes in since it sets up the framework). I'd like to move us away from entirely feature tests anyway and it would be much easier to see what each part is trying to accomplish with some unit tests. Most should be pretty simple. Probably not worth it to have unit tests for the things involving docket but just some of the config parsing would be nice

exo-go/src/app_config_helpers/index.go
```diff
@@ -26,20 +27,82 @@ func GetAppConfig() types.AppConfig {
...
+		for variable, value := range dependency.GetEnvVariables() {
+			envVars[variable] = value
+		}
+	}
+	return envVars
+}
+
+// GetDependencyNames returns the names of all dependencies listed in appConfig
+func GetDependencyNames(appConfig types.AppConfig) []string {
+	dependencyNames := []string{}
```
kevgo: Here is a trick that I like doing: call the variable that contains the result of the function `result`. There are several advantages to this:
- the method name is called `GetDependencyNames`, so it is pretty clear that it returns dependency names. No point in saying that again via the variable name here. 
- if you name this variable `result`, it is abundantly clear that it contains what the function returns. With the current name all it says that it contains some sort of dependency names, but its unclear whether these are the ones that ultimately get returned or other dependency names.
- less thinking about what the proper variable name is
- `result` is shorter

exo-go/src/types/dependency.go
```diff
@@ -0,0 +1,53 @@
...
+// Dependency represents a dependency of an application
+type Dependency struct {
+	Name    string
+	Version string
+	Silent  bool             `yaml:",omitempty"`
+	Config  DependencyConfig `yaml:",omitempty"`
+}
+
+// DependencyConfig represents the configuration of a dependency
+type DependencyConfig struct {
```
kevgo: I would like to suggest putting each struct into its own file, and matching the file name with the struct name. That's a best practice from Ruby that is incredibly useful.

exo-go/src/types/index.go
```diff
@@ -1,42 +1,42 @@
...
+// Services represents the mapping of protection level to services
+type Services struct {
+	Public  map[string]ServiceData
+	Private map[string]ServiceData
 }
 
-// ServiceConfig represents the configuration of a service
+// ServiceConfig represents the configuration of a service as provided in
+// service.yml
 type ServiceConfig struct {
```
kevgo: Please extract all structs into their own files.

kevgo (commented):
We are getting close here!

charlierudolph (commented):
This still needs some unit tests for some of the app config helpers and dependency methods. Otherwise looks good

exo-go/src/app_config_helpers/index_test.go
```diff
@@ -0,0 +1,137 @@
...
+var _ = Describe("GetAppConfig", func() {
+	var appConfig types.AppConfig
+
+	It("should read application.yml successfully", func() {
+		appDir := path.Join("..", "..", "..", "exosphere-shared", "example-apps", "complex-setup-app")
+		if err := os.Chdir(appDir); err != nil {
+			panic(err)
+		}
+		var err error
+		appConfig, err = appConfigHelpers.GetAppConfig()
```
charlierudolph: I'd suggest updating this function to accept a `cwd` so that the tests don't need to change the current working directory.

exo-go/src/app_config_helpers/index_test.go
```diff
@@ -0,0 +1,137 @@
...
+		exocom := types.Dependency{Name: "exocom", Version: "0.22.1"}
+		mongoConfig := types.DependencyConfig{
+			Ports:                 []string{"4000:4000"},
+			Volumes:               []string{"{{EXO_DATA_PATH}}:/data/db"},
+			OnlineText:            "waiting for connections",
+			DependencyEnvironment: map[string]string{"DB_NAME": "test-db"},
+			ServiceEnvironment:    map[string]string{"COLLECTION_NAME": "test-collection"},
+		}
+		mongo := types.Dependency{Name: "mongo", Version: "3.4.0", Config: mongoConfig}
+		expected := []types.Dependency{exocom, mongo}
```
charlierudolph: Can we remove the exocom and mongoConfig variables? I think we can even remove the expected variable.
```go
Expect(appConfig.Dependencies).To(Equal([]types.Dependency{
  types.Dependency{
    Name: "exocom", 
    Version: "0.22.1",
  },
  types.Dependency{
    Name: "mongo", 
    Version: "3.4.0", 
    Config: types.DependencyConfig{
      Ports:                 []string{"4000:4000"},
      Volumes:               []string{"{{EXO_DATA_PATH}}:/data/db"},
      OnlineText:            "waiting for connections",
      DependencyEnvironment: map[string]string{"DB_NAME": "test-db"},
      ServiceEnvironment:    map[string]string{"COLLECTION_NAME": "test-collection"},
    },
  },
}))
```

A similar idea can be applied to other tests

charlierudolph (approved):
Tests look great!

cheahuychou:
closed in: https://github.com/Originate/exosphere/commit/baecdf69b9f7ed67dfd458db833f0733a3873d96

PR #335, opened 2017-07-13
[exo-deploy - 1] Generate aws module

hugoesthere:
First iteration of exo-deploy, it just generates a couple modules containing application config. Working on generating service modules next.


exo-go/bin/setup
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 
 go get -u gopkg.in/alecthomas/gometalinter.v1
+go get -u github.com/jteeuwen/go-bindata
```
charlierudolph: this can be combined with the previous line
```
go get -u gopkg.in/alecthomas/gometalinter.v1 github.com/jteeuwen/go-bindata
```

exo-go/lint-config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Exclude": ["src/terraform_helpers/bindata.go"]
+}
```
charlierudolph: 👍

exo-go/src/terraform_file_helpers/template_builder.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/hoisie/mustache"
+)
+
+// RenderTemplates renders a Terraform template
+func RenderTemplates(templateName string, varsMap map[string]string) string {
+	template := getTemplate(templateName)
+	return mustache.Render(template, varsMap)
+}
+
+// getTemplate returns a stringified template
```
charlierudolph: Comments are only required on public functions.

exo-go/src/terraform_file_helpers/template_builder.go
```diff
@@ -0,0 +1,45 @@
...
+
+	var filePerm os.FileMode = 0744 //standard Unix file permission: rwxrw-rw-
+	err = os.MkdirAll(filepath.Join(cwd, "terraform"), filePerm)
+	if err != nil {
+		log.Fatalf("Failed to get create directory: %s", err)
+	}
+
+	err = ioutil.WriteFile(filepath.Join(cwd, "terraform", "main.tf"), []byte(data), filePerm)
+	if err != nil {
+		log.Fatalf("Failed writing Terraform files: %s", err)
```
charlierudolph: Please no log.Fatal / panic in helper functions. Please only call this from the top level command. Instead this function should return an error. We can wrap the errors via https://github.com/pkg/errors to give extra context.

.travis.yml
```diff
@@ -29,7 +29,8 @@ before_install:
 
 
 install:
-  - yarn install
+  - mkdir node_modules/
+  - mkdir go-binaries/
```
hugoesthere, line 33: This is an attempt to work around Morula. `yarn install` is no longer needed on the root level, and is causing issues because it's running the pre-publish script `bin/cross-compile` before `./morula all bin/setup` is run.  `bin/cross-compile` uses a command `go-bindata`, which is installed in the `exo-go/bin/setup` script. We can simply run `./morula all bin/setup`, but it throws an error when ignored directories haven't been created yet (`node_modules/` and `go-binaries/` in this case).

We're working on a PR for Morula so it doesn't throw when trying to ignore non-existent directories.

hugoesthere:
@charlierudolph ready for another look.

bin/cross-compile
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -ex
 
+cd exo-go
```
kevgo, line 4: Why are you moving test infrastructure around in this PR that is to create exo-deploy? Can you please extract this separate change into its own PR?

exo-go/lint-config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Exclude": ["src/terraform_file_helpers/bindata.go"]
```
kevgo, line 2: Cool!

exo-go/features/help.feature
```diff
@@ -54,6 +55,15 @@ Feature: help command
         exo create
       """
 
+  Scenario: the user enters 'exo deploy help'
+    When running "exo deploy help" in the terminal
```
kevgo: The command to see help is `exo help deploy`

exo-go/src/cmd/deploy.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/Originate/exosphere/exo-go/src/terraform_file_helpers"
+	"github.com/spf13/cobra"
+)
+
+// deployCmd represents the deploy command
+var deployCmd = &cobra.Command{
+	Use:   "deploy",
+	Short: "Deploys Exosphere application to the cloud",
+	Run: func(cmd *cobra.Command, args []string) {
+		if printHelpIfNecessary(cmd, args) {
```
kevgo: What is this method for? Cobra has built-in functionality for showing help. You have to define a `Long` entry in the `cobra.Command` struct for that.

exo-go/src/cmd/deploy.go
```diff
@@ -0,0 +1,36 @@
...
+			return
+		}
+		fmt.Println("We are about to deploy an application!")
+
+		appConfig, err := appConfigHelpers.GetAppConfig()
+		if err != nil {
+			log.Fatalf("Failed to read application configuration: %s", err)
```
kevgo: Shorter and more active error message: `Cannot read application configuration: ...`

exo-go/src/terraform_file_helpers/index.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/pkg/errors"
+)
+
+// GenerateTerraform generates the main terraform file given application and service configuration
+func GenerateTerraform(appConfig types.AppConfig) error {
+	fileData := []string{}
+
+	moduleData, err := generateAwsModule(appConfig)
+	if err != nil {
+		return errors.Wrap(err, "Failed to generate Terraform")
```
kevgo: Failed to generate __AWS__ Terraform module

kevgo (commented):
This is looking promising! Good idea to bundle the Terraform templates into the binary using go-bindata!

exo-go/src/cmd/deploy.go
hugoesthere: Yes, it calls a shared method that every other command has been using. Looks like that method leverages the cobra `Help` function.

bin/cross-compile
hugoesthere: This goes a couple layers deep. `bin/cross-compile` needs to be run at the root level so the `go-binaries/` directory exists on the root level as well. This used to be done in the `pre-publish` script defined in `package.json`, but since we're no longer running `yarn install` at the root during setup (per my comment above), we have to move `cross-compile` to the root level and run it there ourselves.

exo-go/src/terraform_file_helpers/template_builder.go
```diff
@@ -0,0 +1,48 @@
...
+	template, err := getTemplate(templateName)
+	if err != nil {
+		return "", err
+	}
+	return mustache.Render(template, varsMap), nil
+}
+
+// WriteTerraformFile writes the main Terraform file to the path: cwd/terraform/main.tf
+func WriteTerraformFile(data string) error {
+	cwd, err := os.Getwd()
```
charlierudolph: Please pass in the cwd / appDir in as a parameter. Lets just get the working directory once in the command and pass it to the helper functions as needed. This also makes things easier to test.

charlierudolph (approved):

PR #347, opened 2017-07-18
Terraform refactor

hugoesthere:
Big terraform refactor, all changes contained within `exo-deploy/terraform`. `exo-deploy/terraform/vars.tf` and `exo-deploy/terraform/test.tf` would be checked in with the user's application normally but they just serve as examples here.


exo-deploy/terraform/aws/alb-security-groups/security-groups.tf
```diff
@@ -0,0 +1,64 @@
+resource "aws_security_group" "internal_alb" {
+  name        = "${var.name}-internal-alb"
+  vpc_id      = "${var.vpc_id}"
+  description = "Allows internal ALB traffic"
+
+  ingress {
```
Bowbaq: Use separate `aws_security_group_rule` resources to allow further modification elsewhere

exo-deploy/terraform/aws/alb-security-groups/security-groups.tf
```diff
@@ -0,0 +1,64 @@
+resource "aws_security_group" "internal_alb" {
```
Bowbaq: If this is strictly for the ALB, then I think it should either be part of the `alb` module, or at least scoped under it at `alb/security-groups`

exo-deploy/terraform/aws/alb-security-groups/security-groups.tf
```diff
@@ -0,0 +1,64 @@
...
+  name        = "${var.name}-internal-alb"
+  vpc_id      = "${var.vpc_id}"
+  description = "Allows internal ALB traffic"
```
Bowbaq: Perhaps a clearer way to put this would be `Allows internal VPC traffic through the ALB`?

exo-deploy/terraform/aws/alb-security-groups/security-groups.tf
```diff
@@ -0,0 +1,64 @@
...
+  tags {
+    Name        = "${var.name}-internal-alb"
+    Environment = "${var.env}"
+  }
+}
+
+resource "aws_security_group" "external_alb" {
+  name        = "${var.name}-external-alb"
+  vpc_id      = "${var.vpc_id}"
+  description = "Allows external ALB traffic"
```
Bowbaq: Perhaps a clearer way to put this would be `Allows all external traffic through the ALB`?

exo-deploy/terraform/aws/alb-security-groups/vars.tf
```diff
@@ -0,0 +1,29 @@
...
+  description = "The CIDR block of the VPC"
+}
+
+variable "vpc_id" {
+  description = "ID of the VPC"
+}
+
+/* Output */
+
+output "internal_alb_id" {
```
Bowbaq: This is a misleading name (this is not the id of the ALB itself), `internal_id` would be clearer. The importing module can further prefix the output like so:
```
# alb module
output "alb_internal_security_group_id" {
  value = "${module.security_groups.internal_id}"
}
```

exo-deploy/terraform/aws/alb-security-groups/vars.tf
```diff
@@ -0,0 +1,29 @@
...
+  description = "ID of internal ALB security group"
+  value       = "${aws_security_group.internal_alb.id}"
+}
+
+output "external_alb_id" {
```
Bowbaq: This is a misleading name, `external_id` would be clearer

exo-deploy/terraform/aws/alb-security-groups/vars.tf
```diff
@@ -0,0 +1,29 @@
...
+variable "name" {
+  description = "Name of application"
+}
+
+variable "env" {
+  description = "Environment tag, e.g prod"
+}
+
+variable "vpc_cidr" {
+  description = "The CIDR block of the VPC"
```
Bowbaq: Let's try to be consistent in the descriptions, we prefix with `The` here but not below?

exo-deploy/terraform/aws/alb/alb.tf
```diff
@@ -0,0 +1,60 @@
...
+    Service     = "${var.name}"
+    Environment = "${var.env}"
+  }
+
+  access_logs {
+    bucket = "${var.log_bucket}"
+  }
+}
+
+resource "aws_alb_target_group" "target_group" {
```
Bowbaq: This supports `tags`, we should use them

exo-deploy/terraform/aws/alb/route53.tf
```diff
@@ -0,0 +1,25 @@
+resource "aws_route53_record" "external" {
```
Bowbaq: It's strange that route53 is just included in the ALB module but security groups aren't

exo-deploy/terraform/aws/alb/vars.tf
```diff
@@ -0,0 +1,77 @@
...
+  default     = ""
+}
+
+variable "health_check_endpoint" {
+  description = "Endpoint for the alb to hit when performing health checks"
+  default     = "/"
+}
+
+variable "internal" {
+  description = "Set this to false for public ALB's"
```
Bowbaq: `ALBs`

exo-deploy/terraform/aws/alb/vars.tf
```diff
@@ -0,0 +1,77 @@
...
+variable "log_bucket" {
+  description = "S3 bucket id to write ELB logs into"
+}
+
+variable "name" {
+  description = "Name tag, e.g stack"
+}
+
+variable "security_groups" {
+  description = "ID's of the ALB's security groups"
```
Bowbaq: `IDs`. Also we're not using the possessive form [elsewhere](https://github.com/Originate/exosphere/pull/347/files#diff-fd4680f5f5073eb4b3a47883f20a8821R27) we should keep consistent

exo-deploy/terraform/aws/alb/vars.tf
```diff
@@ -0,0 +1,77 @@
...
+  type        = "list"
+}
+
+variable "ssl_certificate_arn" {
+  description = "The ARN of the SSL server certificate. This is not used for internal ALBs"
+  default     = ""
+}
+
+variable "subnet_ids" {
+  description = "List of subnet ID's the ALB should live in"
```
Bowbaq: `IDs`

exo-deploy/terraform/aws/aws.tf
```diff
@@ -0,0 +1,60 @@
+data "aws_region" "current" {
+  current = true
+}
+
+data "aws_availability_zones" "available" {}
+
+module "internal_dns" {
```
Bowbaq: Since this depends on the `network` module, would it make more sense to declare it after in source order?

exo-deploy/terraform/aws/aws.tf
```diff
@@ -0,0 +1,60 @@
...
+  source = "./internal-dns"
+
+  name    = "${var.name}.local"
```
Bowbaq: Seems that the convention is to pass `"${var.env}-${var.name}"` as the name, so this should be `"${var.env}.${var.name}.local"`

exo-deploy/terraform/aws/aws.tf
```diff
@@ -0,0 +1,60 @@
...
+  servers = ["${cidrhost(module.network.vpc_cidr, 2)}"]
+}
+
+module "network" {
+  source = "./network"
+
+  name               = "${var.env}-${var.name}"
+  env                = "${var.env}"
+  availability_zones = "${data.aws_availability_zones.available.names}"
+  region             = "${data.aws_region.current.name}"
```
Bowbaq: Why not use the `aws_region` data source within the module?

exo-deploy/terraform/aws/aws.tf
```diff
@@ -0,0 +1,60 @@
...
+  source = "./ecs-cluster"
+
+  name          = "${var.env}-${var.name}"
+  env           = "${var.env}"
+  region        = "${data.aws_region.current.name}"
+  instance_type = "${var.ecs_instance_type}"
+  ebs_optimized = "${var.ecs_ebs_optimized}"
+  key_name      = "${var.key_name}"
+
+  alb_security_groups = ["${module.alb_security_groups.internal_alb_id}",
```
Bowbaq: Formatting is weird here

exo-deploy/terraform/aws/custom/exocom/exocom-cluster/ami.tf
```diff
@@ -0,0 +1,13 @@
...
+  most_recent = true
+
+  filter {
+    name   = "owner-alias"
+    values = ["amazon"]
+  }
+
+  filter {
+    name   = "name"
+    values = ["amzn-ami-2016.09.g-amazon-ecs-optimized"]
```
Bowbaq: There is a newer AMI (2017.03 if memory serves)

Bowbaq (commented):
*Note*: Clicked the submit button by mistake, expect the second half shortly

exo-deploy/terraform/aws/ecs-cluster/ami.tf
```diff
@@ -0,0 +1,13 @@
...
+  most_recent = true
+
+  filter {
+    name   = "owner-alias"
+    values = ["amazon"]
+  }
+
+  filter {
+    name   = "name"
+    values = ["amzn-ami-2016.09.g-amazon-ecs-optimized"]
```
Bowbaq: There is a more recent AMI (`amzn-ami-2017.03.d-amazon-ecs-optimized`)

exo-deploy/terraform/aws/vars.tf
```diff
@@ -0,0 +1,94 @@
...
+variable "ecs_instance_type" {
+  description = "Instance type to use for ECS instances"
+  default     = "t2.micro"
+}
+
+variable "env" {
+  description = "Name of the environment, used for naming and prefixing"
+}
+
+variable "key_name" {
```
Bowbaq: `ssh_key_name`? Also, using a single SSH key is a bit of an anti pattern we've been trying to move away from. At least for the bastion, it would be preferable to use the public keys of team members

exo-deploy/terraform/aws/network/vpc/vars.tf
```diff
@@ -0,0 +1,26 @@
...
+variable "env" {
+  description = "Environment tag, e.g prod"
+}
+
+variable "cidr" {
+  description = "The CIDR block of the VPC"
+  default     = "10.0.0.0/16"
+}
+
+variable "name" {
```
Bowbaq: move this up one slot, seems like we have `var env` followed by `var name` everywhere else

exo-deploy/terraform/aws/network/subnets/private-subnets.tf
```diff
@@ -0,0 +1,76 @@
+resource "aws_subnet" "private" {
+  vpc_id            = "${var.vpc_id}"
+  cidr_block        = "${cidrsubnet(var.vpc_cidr, 8, count.index + length(var.availability_zones) + 1)}"
```
Bowbaq: `/24` is a rather small subnet, and they're hard to change once in place, perhaps something a bit bigger would be in order

exo-deploy/terraform/aws/network/network.tf
```diff
@@ -0,0 +1,28 @@
...
+}
+
+module "subnets" {
+  source = "./subnets"
+
+  name               = "${var.name}"
+  env                = "${var.env}"
+  vpc_id             = "${module.vpc.id}"
+  vpc_cidr           = "${module.vpc.cidr}"
+  availability_zones = "${var.availability_zones}"
```
Bowbaq: The length of this list should be restricted to a reasonable number (say 3 zones?). `us-east-1` will give you 5 if you use the data source

exo-deploy/terraform/aws/network/subnets/vars.tf
```diff
@@ -0,0 +1,34 @@
+/* Variables */
+
+variable "availability_zones" {
+  description = "Availability zones to use for subnets. Two subnets will be created per availability zone"
+  type        = "list"
+}
+
+variable "env" {
```
Bowbaq: Put env & name at the top?

exo-deploy/terraform/aws/network/vars.tf
```diff
@@ -0,0 +1,54 @@
...
+
+variable "key_name" {
+  description = "Name of the key pair stored in AWS used to SSH into bastion instances"
+}
+
+variable "name" {
+  description = "Name tag, e.g stack"
+}
+
+variable "region" {
```
Bowbaq: Just use the data source in this module?

exo-deploy/terraform/aws/network/bastion/bastion.tf
```diff
@@ -0,0 +1,32 @@
+module "ami" {
```
Bowbaq: We may want to just use amazon linux here, since all the other boxes use it?

exo-deploy/terraform/aws/network/bastion/vars.tf
```diff
@@ -0,0 +1,48 @@
...
+  type        = "list"
+}
+
+variable "vpc_id" {
+  description = "ID of the VPC"
+}
+
+/* Output */
+
+output "ips" {
```
Bowbaq: I've found that remembering bastion IPs is tedious and adding a DNS record is pretty useful (although it does come with a small downgrade in security). Alternatively, perhaps an output of this stack should be the correct SSH configuration to SSH through?

exo-deploy/terraform/aws/ecs-cluster/vars.tf
```diff
@@ -0,0 +1,126 @@
...
+  description = "If CPU usage is above this threshold for 5min, scale up"
+  default     = 90
+}
+
+variable "high_memory_threshold" {
+  description = "If memory usage is above this threshold for 5min, scale up"
+  default     = 90
+}
+
+variable "instance_type" {
```
Bowbaq: Let's try and keep related variables together here, don't put instance_type & key_name in the middle of the cloudwatch alarm settings

exo-deploy/terraform/aws/ecs-cluster/vars.tf
```diff
@@ -0,0 +1,126 @@
...
+  description = "If CPU usage is below this threshold for 5min, scale down"
+  default     = 10
+}
+
+variable "low_memory_threshold" {
+  description = "If memory usage is below this threshold for 5min, scale down"
+  default     = 10
+}
+
+variable "max_size" {
```
Bowbaq: le'ts keep this & min_size together with desired_capacity

exo-deploy/terraform/aws/private-service/private-service.tf
```diff
@@ -0,0 +1,45 @@
...
+  memory                = "${var.memory}"
+  name                  = "${var.env}-${var.name}"
+  region                = "${var.region}"
+}
+
+resource "aws_ecs_service" "service" {
+  name                               = "${var.name}"
+  cluster                            = "${var.cluster_id}"
+  deployment_minimum_healthy_percent = 100
+  desired_count                      = 1
```
Bowbaq: This should be configurable, and the default should not be 1 (we don't want the service to be down if 1 instance dies for some reason)

exo-deploy/terraform/aws/public-service/public-service.tf
```diff
@@ -0,0 +1,48 @@
...
+  memory                = "${var.memory}"
+  name                  = "${var.env}-${var.name}"
+  region                = "${var.region}"
+}
+
+resource "aws_ecs_service" "service" {
+  name                               = "${var.name}"
+  cluster                            = "${var.cluster_id}"
+  deployment_minimum_healthy_percent = 100
+  desired_count                      = 1
```
Bowbaq: Same comment as for the private service, should be configurable & default to 2 or more

exo-deploy/terraform/aws/public-service/public-service.tf
```diff
@@ -0,0 +1,48 @@
...
+  task_definition                    = "${module.task_definition.arn}"
+  iam_role                           = "${var.ecs_role_arn}"
+
+  depends_on = ["module.external_alb"]
```
Bowbaq: This is unnecessary

exo-deploy/terraform/aws/private-service/private-service.tf
```diff
@@ -0,0 +1,45 @@
...
+  task_definition                    = "${module.task_definition.arn}"
+  iam_role                           = "${var.ecs_role_arn}"
+
+  depends_on = ["module.internal_alb"]
```
Bowbaq: This is unnecessary

exo-deploy/terraform/test.tf
```diff
@@ -0,0 +1,148 @@
+terraform {
+  required_version = "= 0.9.9"
```
Bowbaq: might as well require the latest here, 0.9.11

exo-deploy/terraform/test.tf
```diff
@@ -0,0 +1,148 @@
...
+
+  backend "s3" {
```
Bowbaq: Where do the credentials come from for this?

Bowbaq (changes requested):
Comments about this code:
- The fact that `vars.tf` files also include outputs is a bit weird
- When there is a choice of using an inline block or a separate resource (security group rules for example), use a separate resource. This lets users of the module further modify the resource if needed
- Why doesn't `exocom-cluster` make use of `ecs-cluster`?
- We should try to keep the ordering of variables between modules consistent (ie. always put `env` & `name` at the top, followed by required variables, followed by optional variables. Let's also group related variables together, it makes it much easier to find them
- This should include a module to setup ECR repositories. Another useful module would be an RDS database.
- A README explaining how to use these modules is in order I think. I think this should at least include the setup required before this can be used (create a bucket for the terraform state, an SSH key, register a domain name, setup a route53 zone, setup a certificate, ...)
- A CI job that verifies that the code is properly formatted (`terraform fmt`) and that it at least plans cleanly (run `terraform plan` with no errors) would be nice

General comments:
- How does this relate to the code under `exo-deploy/templates`?
- This is really big & hard to review. If there's any chance to break this into smaller chunks, that would be good
- Why would you check in any of the module code under `exo-deploy/terraform` with the user application?

exo-deploy/terraform/aws/alb-security-groups/security-groups.tf
chrizkim: @Bowbaq what's the benefit of the `aws_security_group_rule` over inline blocks?

exo-deploy/terraform/aws/alb/route53.tf
chrizkim: this was because there was a 1-1 mapping between a record and an ALB, while security groups were 1-many (1 security group would be reused for many ALBs). @Bowbaq - do you think it would be preferable to have each ALB create its own security group?

exo-deploy/terraform/aws/alb-security-groups/security-groups.tf
Bowbaq: If you don't, then you can't add rule to the security group after the fact (both methods are incompatible). Adding rules outside the defining module is sometimes necessary. Since it's functionally equivalent, the more flexible option is preferable. See [docs](https://www.terraform.io/docs/providers/aws/r/security_group_rule.html)

exo-deploy/terraform/aws/alb/route53.tf
Bowbaq: That would allow for finer grained access control. 

At the very least, we should assign the private security group to private services & workers and the public one to public services, otherwise there isn't much of a point in having two of them. If we go that route then I'd say just roll this in the `ecs-cluster` security group file

hugoesthere:
> The fact that vars.tf files also include outputs is a bit weird

Yeah, I thought it might be easier if you could just look at one file and see what needs to be passed in and what's outputted, but if that's not idiomatic we can move those output blocks

> Why doesn't exocom-cluster make use of ecs-cluster

`exocom-cluster` uses its own `aws_instance`. We need some way to grab the IP of the instance Exocom is running on so we can create a route 53 record for it. We can't place an alb in front of that service b/c it needs to support TCP. I forget why we didn't use and elb or try beanstalk. @chrizkim?

> This should include a module to setup ECR repositories. Another useful module would be an RDS database. 
I think this should at least include the setup required before this can be used (create a bucket for the terraform state, an SSH key, register a domain name, setup a route53 zone, setup a certificate, ...)

Yes, we have plans to implement these.

> How does this relate to the code under exo-deploy/templates

It replaces everything in there.

> Why would you check in any of the module code under exo-deploy/terraform with the user application?

Only the application specific `.tf` file will be checked in, `test.tf` in this case. Everything else will be sourced.

Bowbaq:
> exocom-cluster uses its own aws_instance. We need some way to grab the IP of the instance Exocom is running on so we can create a route 53 record for it. We can't place an alb in front of that service b/c it needs to support TCP. I forget why we didn't use and elb or try beanstalk.

Having a single container on a single instance is a double whammy as far a single points of failure go. This is basically guaranteed to fail. 

If the only reason is that the IP needs to be known, why not just make it a service in the regular cluster & use the instance metadata API on startup to discover the host IP and update a DNS record in Route53? That only solves half the problem of course, for the second half Exocom will have to support some sort of high-availability mode

exo-deploy/terraform/aws/aws.tf
chrizkim: I think I would want the local name to be consistent regardless of the environment. For example `users.myapp.local` should just go to the users service within the correct environment. I wouldn't want to have to do `users.prod-myapp.local` for one environment and `users.dev-myapp.local` for another when the two can't even conflict

exo-deploy/terraform/aws/aws.tf
chrizkim: i felt like it made the lower level modules more flexible

exo-deploy/terraform/aws/network/subnets/private-subnets.tf
chrizkim: what about `/20` (16 subnets, 4,064 hosts per subnet):
```
cidrsubnet(var.vpc_cidr, 4, (count.index + length(var.availability_zones)) * 16)
```

exo-deploy/terraform/aws/network/network.tf
chrizkim: +1 Good call, i don't think we would want to use more than 3. should probably set this in `aws.tf`

exo-deploy/terraform/aws/network/bastion/bastion.tf
chrizkim: Amazon Linux is pretty bloated with crap so I prefer to avoid it as a bastion

exo-deploy/terraform/aws/network/bastion/vars.tf
chrizkim: don't like the idea of an easily guessable hostname for bastions, especially since we would want to reuse this setup for multiple partners/projects. outputting ssh config sounds good though

exo-deploy/terraform/aws/aws.tf
Bowbaq: I think namespacing with the environment makes sense because for external DNS you have to do it anyways, so it makes sense to keep things consistent.

At least we should set the comment to something more meaningful than `""` so that they can be differentiated in the UI (currently if you have 5 env, you have 5 identically named zones)

exo-deploy/terraform/aws/aws.tf
Bowbaq: How so?

exo-deploy/terraform/aws/network/subnets/private-subnets.tf
Bowbaq: `/20` sounds good, I haven't checked the math

exo-deploy/terraform/aws/network/bastion/bastion.tf
Bowbaq: I don't feel strongly about this, but consistency would be nice. The segmentio stack uses ubuntu-based AMIs instead of Amazon's ECS AMIs

Bowbaq:
I've just learned that `terraform { ... }` blocks can be used in all modules (not just the root module). We should figure out the minimum version of terraform required for this set of modules to work and loosely pin them (maybe something like `>= 0.9, < 0.10`)

exo-deploy/terraform/aws/alb/route53.tf
hugoesthere: We are assigning the public sg to public services and private sg to private ones. Worker services don't need a load balancer sg since they have no load balancer. I had moved those modules to `alb/security_groups/` but we can move it into `ecs-cluster/security-groups.tf` if that's what we agree on.

exo-deploy/terraform/aws/vars.tf
hugoesthere: How does that work? Do you just pass in the public key instead of the key name?

exo-deploy/terraform/aws/alb/route53.tf
Bowbaq: I was thinking of [this line](https://github.com/Originate/exosphere/pull/347/files#diff-72f68ee95759d5b6b31ca9523e06c535R45) but you're right, it doesn't apply.

I guess my only remaining concern is that you could easily write something like:
```hcl
module "web" {
  source = "./aws/public-service"

  name               = "web"
  alb_security_group = "${module.aws.internal_alb_security_group}"

  ...
}
```

This is incorrect but pretty easy to miss IMO

exo-deploy/terraform/aws/vars.tf
Bowbaq: You can do it a few different ways:
- Bake an AMI with the correct keys
- Create a temporary key for setup then have something like ansible come in & setup the keys
- Put the keys in an S3 bucket and have a script on the bastion poll periodically

exo-deploy/terraform/aws/alb/route53.tf
hugoesthere: Is it possible to assign the correct security group to the proper ALB at a lower level (without creating a security group for each ALB)? The only reason we bubble up the security groups is because the services and hence ALBs are created at the root level.

Or is it not a big deal to be creating a security group for each ALB?

Bowbaq: Security groups are free, so creating one per ALB shouldn't be a problem. That said, you'd still need to authorize traffic from those security groups to the ECS instances, so that would really just move the problem.

One option could be to use the [data source](https://www.terraform.io/docs/providers/aws/d/security_group.html) to find the proper security group within the VPC from inside the service modules. Using something like `coalesce()`, this could be the default, while still leaving the possibility for the user of the module to override if they know better.

hugoesthere:
@Bowbaq we're merging this for now. This will be a first iteration and we'll refine/add to it later but in the mean time we just need something that works.

PR #346, opened 2017-07-18
[exo-deploy 3] Generate Exocom module

hugoesthere:

PR #349, opened 2017-07-19
[new exo-run - 1] get docker configs

cheahuychou:
- Use `AppDependency` interface as defined in #346 and move `Dependency` tests to `app_dependency_helpers/`
- Update `ServiceConfig` struct and add `GetDockerConfig` function which will be used in `dockerSetup` in #350 and #351 .

### DIFF
--- a/exo-go/src/app_config_helpers/index.go
+++ b/exo-go/src/app_config_helpers/index.go
@@ -28,10 +28,10 @@ func GetAppConfig(appDir string) (result types.AppConfig, err error) {
 
 // GetEnvironmentVariables returns the environment variables of
 // all dependencies listed in appConfig
-func GetEnvironmentVariables(appConfig types.AppConfig) map[string]string {
+func GetEnvironmentVariables(appConfig types.AppConfig, appDir string) map[string]string {
 	result := map[string]string{}
 	for _, dependency := range appConfig.Dependencies {
-		for variable, value := range appDependencyHelpers.Build(dependency, appConfig).GetEnvVariables() {
+		for variable, value := range appDependencyHelpers.Build(dependency, appConfig, appDir).GetEnvVariables() {
 			result[variable] = value
 		}
 	}
--- a/exo-go/src/app_config_helpers/index.go
+++ b/exo-go/src/app_config_helpers/index.go
@@ -9,7 +9,7 @@ import (
 	"github.com/Originate/exosphere/exo-go/src/types"
 	"github.com/Originate/exosphere/exo-go/src/util"
 	"github.com/pkg/errors"
-	prompt "github.com/segmentio/go-prompt"
+	"github.com/segmentio/go-prompt"
 	"gopkg.in/yaml.v2"
 )
 
--- a/exo-go/src/app_config_helpers/index.go
+++ b/exo-go/src/app_config_helpers/index.go
@@ -5,10 +5,11 @@ import (
 	"io/ioutil"
 	"path"
 
+	"github.com/Originate/exosphere/exo-go/src/app_dependency_helpers"
 	"github.com/Originate/exosphere/exo-go/src/types"
 	"github.com/Originate/exosphere/exo-go/src/util"
 	"github.com/pkg/errors"
-	"github.com/segmentio/go-prompt"
+	prompt "github.com/segmentio/go-prompt"
 	"gopkg.in/yaml.v2"
 )
 
@@ -30,7 +31,7 @@ func GetAppConfig(appDir string) (result types.AppConfig, err error) {
 func GetEnvironmentVariables(appConfig types.AppConfig) map[string]string {
 	result := map[string]string{}
 	for _, dependency := range appConfig.Dependencies {
-		for variable, value := range dependency.GetEnvVariables() {
+		for variable, value := range appDependencyHelpers.Build(dependency, appConfig).GetEnvVariables() {
 			result[variable] = value
 		}
 	}
--- a/exo-go/src/app_config_helpers/index.go
+++ b/exo-go/src/app_config_helpers/index.go
@@ -28,10 +28,10 @@ func GetAppConfig(appDir string) (result types.AppConfig, err error) {
 
 // GetEnvironmentVariables returns the environment variables of
 // all dependencies listed in appConfig
-func GetEnvironmentVariables(appConfig types.AppConfig, appDir string) map[string]string {
+func GetEnvironmentVariables(appConfig types.AppConfig, appDir, homeDir string) map[string]string {
 	result := map[string]string{}
 	for _, dependency := range appConfig.Dependencies {
-		for variable, value := range appDependencyHelpers.Build(dependency, appConfig, appDir).GetEnvVariables() {
+		for variable, value := range appDependencyHelpers.Build(dependency, appConfig, appDir, homeDir).GetEnvVariables() {
 			result[variable] = value
 		}
 	}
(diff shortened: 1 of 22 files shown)


exo-go/src/app_dependency_helpers/index.go
```diff
@@ -0,0 +1,25 @@
+package appDependencyHelpers
+
+import (
+	"github.com/Originate/exosphere/exo-go/src/types"
+)
+
+// AppDependency contains methods that return config information about a dependency
+type AppDependency interface {
+	GetContainerName() string
+	GetDockerConfig() (types.DockerConfig, error)
```
cheahuychou: @charlierudolph  This function is not unit-tested yet but I will be writing [tests](https://github.com/Originate/exosphere/blob/master/exo-setup/src/app-setup.spec.ls) for `docker-compose` which will indirectly test this function. I can write unit tests here too if you want.

cheahuychou:

cheahuychou:

exo-go/src/app_dependency_helpers/index_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	var _ = Describe("GetContainerName", func() {
+		It("should be the concatenation of dependency name and version", func() {
+			Expect(exocom.GetContainerName()).To(Equal("exocom0.22.1"))
+			Expect(mongo.GetContainerName()).To(Equal("mongo3.4.0"))
+			Expect(nats.GetContainerName()).To(Equal("nats0.9.6"))
+		})
+	})
+
+	var _ = Describe("GetDokerConfig", func() {
```
charlierudolph: I believe there is a typo here as the c is missing from Docker

exo-go/src/app_config_helpers/index.go
```diff
@@ -5,10 +5,11 @@ import (
 	"io/ioutil"
 	"path"
 
+	"github.com/Originate/exosphere/exo-go/src/app_dependency_helpers"
 	"github.com/Originate/exosphere/exo-go/src/types"
 	"github.com/Originate/exosphere/exo-go/src/util"
 	"github.com/pkg/errors"
-	"github.com/segmentio/go-prompt"
+	prompt "github.com/segmentio/go-prompt"
```
charlierudolph: This seems like an unrelated change. Did you change this, goimports change it or something else?

exo-go/src/app_dependency_helpers/exocom.go
```diff
@@ -19,3 +52,47 @@ func (exocom exocomDependency) GetDeploymentConfig() map[string]string {
...
+	}
+	return map[string]string{"EXOCOM_PORT": port}
+}
+
+// GetOnlineText returns the online text for the exocom
+func (exocom exocomDependency) GetOnlineText() string {
+	return "ExoCom WebSocket listener online"
+}
+
+// GetServiceEnvVariables returns the environment variables for the depedency
```
charlierudolph: This comment matches `GetEnvVariables`. These are the env variables that need to be passed to the services.

Same update needed for the other app dependencies

exo-go/src/app_dependency_helpers/exocom.go
```diff
@@ -1,12 +1,45 @@
...
+	}
+	serviceData := serviceConfigHelpers.GetServiceData(exocom.appConfig.Services)
+	for serviceName, serviceConfig := range serviceConfigs {
+		route := map[string]interface{}{
+			"role":     serviceName,
+			"receives": serviceConfig.ServiceMessages.Receives,
+			"sends":    serviceConfig.ServiceMessages.Sends,
+		}
+		namespace := serviceData[serviceName].NameSpace
+		if len(namespace) > 0 {
```
charlierudolph: Would result in the same thing but thoughts on using `if namespace != "" {`. I prefer that since `""` is the empty value for strings.

exo-go/src/app_dependency_helpers/generic.go
```diff
@@ -16,3 +29,48 @@ func (dependency genericDependency) GetDeploymentConfig() map[string]string {
...
+// GetOnlineText returns the online text for the dependency
+func (dependency genericDependency) GetOnlineText() string {
+	return dependency.config.Config.OnlineText
+}
+
+func (dependency genericDependency) getRenderedVolumes() ([]string, error) {
+	homeDir, err := osHelpers.GetUserHomeDir()
+	if err != nil {
+		return []string{}, err
+	}
```
charlierudolph: Thoughts on making the homeDir a property of genericDependency? This would be passed in when its created and would thus need to be an argument to `appDependencyHelper.Build`. That would make this easier to test.

exo-go/src/app_dependency_helpers/index_test.go
```diff
@@ -0,0 +1,167 @@
...
+				case "exocom":
+					exocom = builtDependency
+				case "mongo":
+					mongo = builtDependency
+				}
+			}
+			nats = appDependencyHelpers.Build(types.Dependency{
+				Name:    "nats",
+				Version: "0.9.6",
+			}, appConfig, appDir)
```
charlierudolph: This appears to be setting `nats` for all future tests in this file. The initialization of nats should be done in a beforeEach block. We want all the `it` to not effect each other and can use beforeEach for common setup needed between tests. Each test should start with a clean slate. Thoughts on restructuring these tests to be the following:
```go
var _ = Describe("AppDependency", func() {
  var _ = Describe("Build", func() {})
  
  var _ = Describe("nats dependency", func() { 
    // Tests for GetContainerName / GetDokerConfig / etc 
    // a beforeEach creates the nats variable used in each test
  })
  
  var _ = Describe("exocom dependency", func() {
    // similar to nats dependency
  })
  
  var _ = Describe("generic dependency", func() {
    // similar to nats dependency
  })
})
```

exo-go/src/terraform_file_helpers/index.go
```diff
@@ -94,9 +95,13 @@ func generateServiceModule(serviceName string, serviceConfig types.ServiceConfig
 }
 
 func generateDependencyModules(appConfig types.AppConfig) (string, error) {
+	appDir, err := os.Getwd()
+	if err != nil {
+		return "", err
+	}
```
charlierudolph: Please pass in the cwd instead of grabbing it inside this function. That will make this easier to test and help us limit the calls to `os.Getwd()`.

exo-go/src/app_dependency_helpers/exocom.go
```diff
@@ -1,12 +1,45 @@
...
+			route["namespace"] = namespace
+		}
+		routes = append(routes, route)
+	}
+	return routes, nil
+}
+
+// GetContainerName returns the container name for the dependency
```
charlierudolph: I think `for the dependency` can be removed from all the comments. I don;'t think it adds anything as you are calling this method on the dependency so I think that is a safe assumption that that is what you are talking about.

Same update needed for the other app dependencies.

exo-go/src/service_config_helpers/index_test.go
```diff
@@ -21,6 +21,32 @@ var _ = BeforeSuite(func() {
 	Expect(err).ToNot(HaveOccurred())
 })
 
+var _ = Describe("GetServiceData", func() {
+
+	It("should not return an error when application.yml is valid", func() {
```
charlierudolph: GetServiceData never returns an error so I think this test description needs to be updated to actually describe what this function is doing. Maybe `joins the public and private services into a single map`

exo-go/src/app_config_helpers/index.go
cheahuychou: I didn't change it, goimports changed it.

charlierudolph: k. fine as is. Odd that it changed it now

cheahuychou:

charlierudolph (approved):

cheahuychou:
closed in: https://github.com/Originate/exosphere/commit/fb29477ac4607ce1e93a0bac2bec7c50a6b9cc2d

PR #350, opened 2017-07-19
[new exo-run - 2] write docker_setup in Go

cheahuychou:


cheahuychou:

cheahuychou:

exo-go/src/docker_setup/index.go
```diff
@@ -0,0 +1,138 @@
...
+func NewDockerSetup(appConfig types.AppConfig, serviceConfig types.ServiceConfig, serviceData types.ServiceData, role string, logger *logger.Logger, appDir string) *DockerSetup {
+	return &DockerSetup{
+		AppConfig:     appConfig,
+		ServiceConfig: serviceConfig,
+		ServiceData:   serviceData,
+		Role:          role,
+		Logger:        logger,
+		AppDir:        appDir,
+	}
+}
```
charlierudolph: I don't think this function is needed since all you do is assign things. In this case, the caller can just create a `DockerSetup` struct directly.

exo-go/src/docker_setup/index.go
```diff
@@ -0,0 +1,138 @@
...
+	for _, dependency := range dockerSetup.ServiceConfig.Dependencies {
+		result[strings.ToUpper(dependency.Name)] = dependency.Name
+	}
+	return result
+}
+
+func (dockerSetup *DockerSetup) getDockerLinks() []string {
+	result := []string{}
+	for _, dependency := range dockerSetup.ServiceConfig.Dependencies {
+		result = append(result, fmt.Sprintf("%s%s:%s", dependency.Name, dependency.Version, dependency.Name))
```
charlierudolph, line 44: Is the first part of this the docker container name? could you use dependeny.GetContainerName()?

exo-go/src/docker_setup/index.go
```diff
@@ -0,0 +1,138 @@
...
+		return result, err
+	}
+	return joinDockerConfigMaps(result, dependencyDockerConfigs), nil
+}
+
+func (dockerSetup *DockerSetup) getRenderedVolumes(volumes []string) ([]string, error) {
+	homeDir, err := osHelpers.GetUserHomeDir()
+	if err != nil {
+		return []string{}, err
+	}
```
charlierudolph: Please pass in the home dir

exo-go/src/docker_setup/index.go
```diff
@@ -0,0 +1,138 @@
...
+	dataPath := path.Join(homeDir, ".exosphere", dockerSetup.AppConfig.Name, dockerSetup.Role, "data")
+	renderedVolumes := []string{}
+	if err := os.MkdirAll(dataPath, 0777); err != nil { //nolint gas
+		return renderedVolumes, errors.Wrap(err, "Failed to create the necessary directories for the volumes")
+	}
+	for _, volume := range volumes {
+		renderedVolumes = append(renderedVolumes, strings.Replace(volume, "{{EXO_DATA_PATH}}", dataPath, -1))
+	}
+	return renderedVolumes, nil
```
charlierudolph: This is very similar to the getRenderedVolums in the first PR. I think we can create a helper to dry this up

exo-go/src/docker_setup/index.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (dockerSetup *DockerSetup) getServiceDependenciesDockerConfigs() (map[string]types.DockerConfig, error) {
+	result := map[string]types.DockerConfig{}
+	for _, dependency := range dockerSetup.ServiceConfig.Dependencies {
+		if !dependency.Config.IsEmpty() {
+			// only add dependency docker config if the Config field exists,
+			// otherwise the dependency has already been listed as an application dependency
```
charlierudolph: Can we have tests for this instead of a comment?

exo-go/src/docker_setup/index_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	var _ = BeforeSuite(func() {
+		appDir = path.Join("..", "..", "..", "exosphere-shared", "example-apps", "external-dependency")
+		appConfig, err := appConfigHelpers.GetAppConfig(appDir)
+		Expect(err).NotTo(HaveOccurred())
+		serviceConfigs, err := serviceConfigHelpers.GetServiceConfigs(appDir, appConfig)
+		Expect(err).NotTo(HaveOccurred())
+		serviceData := serviceConfigHelpers.GetServiceData(appConfig.Services)
+		serviceName := "mongo"
+		setup = dockerSetup.NewDockerSetup(appConfig, serviceConfigs[serviceName], serviceData[serviceName], serviceName, logger.NewLogger([]string{}, []string{}), appDir)
```
charlierudolph: Please create a new `dockerSetup` in a beforeEach so we have a new instance in each test. The rest should be fine to keep in `BeforeSuite `

exo-go/src/docker_setup/index_test.go
```diff
@@ -0,0 +1,74 @@
...
+	})
+
+	var _ = Describe("GetServiceDockerConfigs", func() {
+		var dockerConfigs map[string]types.DockerConfig
```
charlierudolph: If this is needed in multiple tests, please create it in a `beforeEach` so the tests do not rely on each other

exo-go/src/docker_setup/index.go
cheahuychou: no, we need to call `appDependencyHelpers.Build(...)` first. I could do that if you want.

exo-go/src/docker_setup/index.go
cheahuychou: We do have tests for this but in #351 where we test the generated `docker-compose.yml` for a complex app. Do you want me to test it here too? I think it makes sense.

exo-go/src/docker_setup/index.go
charlierudolph: Hmm, If thats whats being used I think that would be a good thing in case we want to change the container names down the road.

cheahuychou: I actually don't know if the first part of each link is supposed to be the container name or it just happens to be. I think we should leave it as it is. What do you think?

exo-go/src/docker_setup/index.go
cheahuychou: I have added tests for this.

cheahuychou:
@charlierudolph all comments addressed.

exo-go/src/docker_setup/index.go
charlierudolph: @hugobho do you know? What is the format of docker links?

exo-go/src/docker_setup/index.go
charlierudolph: Since we have tests for this can we remove the comment?

charlierudolph (approved):
We can follow up with @hugobho if that is suppose to be the container name. And can fix it separately

cheahuychou:
closed in: https://github.com/Originate/exosphere/commit/7a0b544cd3e45a27b29f47d2f515bb977f43d1eb

exo-go/src/docker_setup/index.go
hugoesthere: The mapping is `#{dependency_container_name}:#{dependency_name}`. `dependency_container_name` is generated by us and defined in `docker-compose.yml`, and `dependency_name` is the name a service gives its dependency in its source code. Let's say you have a `users` service that connects to mongo via `http://mongo...` in its source. The mongo container is given a name `mongo:3.4.0`, but the `users` service expects to connect to a container called `mongo`. With this link mapping, `http://mongo...` will be redirected to hit the `mongo:3.4.0` container.

PR #351, opened 2017-07-19
[new exo-run - 3] write app_setup in Go

cheahuychou:


cheahuychou:

cheahuychou:

exo-go/src/app_setup/helpers.go
```diff
@@ -0,0 +1,30 @@
...
+// CheckoutApp copies the example app appName to cwd
+func CheckoutApp(cwd, appName string) error {
+	src := path.Join("..", "..", "..", "exosphere-shared", "example-apps", appName)
+	dest := path.Join(cwd, appName)
+	err := os.RemoveAll(dest)
+	if err != nil {
+		return err
+	}
+	return testHelpers.CopyDir(src, dest)
+}
```
charlierudolph: I think this should be in test helpers since its only used in tests

exo-go/src/app_setup/index.go
```diff
@@ -0,0 +1,129 @@
...
+		return err
+	}
+	if err := process.Wait(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// StartSetup sets up the entire app and returns an error if any
+func (appSetup *AppSetup) StartSetup() error {
```
charlierudolph: This name makes it seem like the setup is not finished after it returns. Is that true?

exo-go/src/docker_compose/index.go
```diff
@@ -18,3 +18,19 @@ func KillAllContainers(cwd string, log func(string)) (*processHelpers.Process, e
 	process.SetStdoutLog(log)
 	return process, process.Start()
 }
+
+// PullAllImages pulls all the docker images defined in docker-compose.yml
+func PullAllImages(dockerComposeLocation string, log func(string)) (*processHelpers.Process, error) {
+	process := processHelpers.NewProcess("docker-compose", "pull")
+	process.SetDir(dockerComposeLocation)
+	process.SetStdoutLog(log)
+	return process, process.Start()
```
charlierudolph: The use of each of these calls Wait right after. Thoughts on updating these functions to wait for the process to finish and thus just return an error. Same for `BuildAllImages`

exo-go/src/app_setup/index_test.go
```diff
@@ -0,0 +1,155 @@
...
+		externalServices = []string{"external-service"}
+		internalDependencies = []string{"exocom0.22.1"}
+		externalDependencies = []string{"mongo3.4.0"}
+		allServices = util.JoinSlices(internalServices, externalServices, internalDependencies, externalDependencies)
+	})
+
+	var _ = Describe("StartSetup", func() {
+
+		It("should starts up the app sucessfully when application.yml and service.yml are valid", func() {
+			err := setup.StartSetup()
```
charlierudolph: Please update the tests to not rely on each other. Put anything you need in each test in a beforeEach / beforeSuite. If there are some things that only needed in a few of the tests wrap them in a describe block so it can have its own beforeEach

exo-go/src/util/slice.go
```diff
@@ -0,0 +1,10 @@
+package util
+
+// JoinSlices joins the given slices
+func JoinSlices(slices ...[]string) []string {
+	result := []string{}
+	for _, slice := range slices {
+		result = append(result, slice...)
+	}
+	return result
+}
```
charlierudolph, line 10: Thoughts on naming this `JoinStringSlices`?

exo-go/src/app_setup/index_test.go
cheahuychou: but then this test suite will call`StartSetup()` (`exo-setup`) 10+ times?

charlierudolph: If you don't want it in beforeEach (because its too slow) lets move it to beforeSuite so its more apparent we need it for all the tests.

exo-go/src/app_setup/helpers.go
cheahuychou: As I mentioned, we can't use the existing function in `test_helpers/index.go` because of the path difference. I think we might need to pass in relative path.

cheahuychou:

exo-go/src/app_setup/helpers.go
charlierudolph: Lets try following this https://andrewbrookins.com/tech/golang-get-directory-of-the-current-file/ in order to the get the current file.

exo-go/src/app_setup/index.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+// AppSetup sets up the app
+type AppSetup struct {
+	AppConfig             types.AppConfig
+	Logger                *logger.Logger
+	DockerComposeConfig   types.DockerCompose
+	ServiceData           map[string]types.ServiceData
+	ServiceConfigs        map[string]types.ServiceConfig
+	DockerComposeLocation string
```
charlierudolph: I don't think this is good name. Its actually the folder we are storing `docker-compose.yml`. It also appears that its related to appDir. Thoughts on creating a function which computes this to avoid us from having to save it as a variable?

Also it appears `dockerComposeLocation` is used mostly to refer to the folder that contains `docker-compose.yml` and but is also used in dockerHelpers.GetDockerCompose as the path to the `docker-compose.yml`. Thoughts on renaming to `dockerComposeDirPath` and `dockerComposeFilePath`?

cheahuychou:

exo-go/src/app_setup/index.go
cheahuychou: comment addressed

exo-go/src/app_setup/index_test.go
```diff
@@ -0,0 +1,146 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Setup", func() {
+
+	var dockerCompose types.DockerCompose
+	var internalServices, externalServices, internalDependencies, externalDependencies, allServices []string
+
+	var _ = BeforeSuite(func() {
+		err := testHelpers.CheckoutApp("", "complex-setup-app")
```
charlierudolph: I don't like passing in "". Can we pass in the actual cwd?

charlierudolph (approved):

cheahuychou:
closed in: https://github.com/Originate/exosphere/commit/7458f863fb2655d939423df547cd0881b0732839

PR #355, opened 2017-07-20
simplify struct methods

charlierudolph:
* Use pointer receivers everywhere so we can modify the struct if needed. Otherwise you get a copy of the object and thus can't edit it.
* Make the variables a single letter as they are essentially the `this` we were using in javascript and we are likely to use it everywhere.

### DIFF
--- a/exo-go/src/app_dependency_helpers/exocom.go
+++ b/exo-go/src/app_dependency_helpers/exocom.go
@@ -15,13 +15,13 @@ type exocomDependency struct {
 	appDir    string
 }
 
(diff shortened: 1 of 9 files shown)


kevgo (approved):
+1 on using pointers as receivers.

For single-letter characters, I personally consider them an anti-pattern that is an obstacle to self-describing code. Unless using `i` for loop indexes. See https://docs.quantifiedcode.com/python-anti-patterns/maintainability/using_single_letter_as_variable_name.html for more background.

But I recognize that this is a Go pattern here. I wish the Go community would at least use `this` here.

cheahuychou (approved):

PR #358, opened 2017-07-21
process_helpers updates

charlierudolph:
Removes the pipes solution in favor of using a single pipe and allowing multiple functions to be added as callbacks anytime something is output.


exo-go/src/app_runner/index.go
```diff
@@ -90,26 +91,29 @@ func (a *AppRunner) Start() error {
...
+		onlineTextRegex, err = regexp.Compile(fmt.Sprintf("%s.*%s", role, onlineText))
+		if err != nil {
 			return err
 		}
+		go func(role string, onlineTextRegex *regexp.Regexp) {
+			a.waitForOnlineText(process, role, onlineTextRegex)
+			wg.Done()
+		}(role, onlineTextRegex)
 	}
+	wg.Wait()
```
charlierudolph, line 107: We can move back to doing this all at the same time since we no longer have multiple goroutines reading from the same pipe and instead each is adding a new func to be called anytime there is new output

exo-go/src/app_runner/index.go
```diff
@@ -90,26 +91,29 @@ func (a *AppRunner) Start() error {
 	if err != nil {
 		return err
 	}
+	var wg sync.WaitGroup
+	var onlineTextRegex *regexp.Regexp
 	for role, onlineText := range onlineTexts {
-		if err = a.waitForOnlineText(process, role, onlineText); err != nil {
+		wg.Add(1)
+		onlineTextRegex, err = regexp.Compile(fmt.Sprintf("%s.*%s", role, onlineText))
```
charlierudolph, line 98: Moved this regular expression compilation up here so we can deal with the error if there is one and now the goroutine doesn't need to somehow pass an error

exo-go/src/process_helpers/process.go
```diff
@@ -119,47 +129,41 @@ func (p *Process) Wait() error {
...
+	} else {
+		id := uuid.NewV4().String()
+		p.AddOutputFunc(id, func(output string) {
+			if condition(output) {
+				err <- nil
+				p.RemoveOutputFunc(id)
+			}
+		})
 	}
+	p.outputMutex.Unlock()
```
charlierudolph, line 144: Lock/unlock the output to prevent it from being updated while this function is checking against the pre-existing output and adding a new callback if needed

exo-go/test_helpers/add_feature_context.go
```diff
@@ -28,7 +28,7 @@ func createEmptyApp(appName, cwd string) error {
 	fields := []string{"AppName", "AppDescription", "AppVersion", "ExocomVersion"}
 	inputs := []string{appName, "Empty test application", "1.0.0", "0.22.1"}
 	for i, field := range fields {
-		if err := process.WaitForTextWithTimeout(field, 1000); err != nil {
+		if err := process.WaitForTextWithTimeout(field, 5000); err != nil {
```
charlierudolph, line 32: The 1000 millisecond timeout wasn't enough on travis. It shouldn't ever take 5 seconds but this is just to give it a nice buffer

exo-go/src/process_helpers/index.go
```diff
@@ -26,24 +25,13 @@ func Run(dir string, commandWords ...string) (string, error) {
...
-	}
-	cmd := exec.Command(commandWords[0], commandWords[1:]...) // nolint gas
-	cmd.Dir = dir
-	stdoutPipe, err := cmd.StdoutPipe()
-	if err != nil {
-		return err
-	}
-	go readPipe(stdoutPipe, log)
-	if err := cmd.Start(); err != nil {
+	process := NewProcess("docker-compose", "up")
```
cheahuychou: Why are we making a process for `"docker-compose up"` inside `RunAndLog`?

charlierudolph: no wonder this is broken. Thanks!

PR #361, opened 2017-07-24
[new exo-run 4] boot up dependencies separately from services

### LINKED ISSUE #315: boot up dependencies separately from services
> When using dependencies that take a while to boot up
I want "exo run" to wait before it boots up services that depend on them
So that my services find fully running dependencies when they boot up and don't have to implement complicated wait or retry logic to connect to their dependencies.

- "exo run" first boots up all dependencies, and only when they all run then all the services
- the dependencies include application and service dependencies
- after the dependencies are up, it prints "dependencies are online"
- after the services are up, it prints "services are online"

cheahuychou:
resolves #315


exo-go/src/app_runner/index.go
```diff
@@ -82,23 +89,49 @@ func (a *AppRunner) Shutdown(shutdownConfig types.ShutdownConfig) error {
...
+		return errors.Wrap(err, fmt.Sprintf("Failed to run services\nOutput: %s\nError: %s\n", process.Output, err))
+	}
+	for role, onlineText := range serviceOnlineTexts {
+		if err := a.waitForOnlineText(process, role, onlineText); err != nil {
+			return err
+		}
 	}
-	return err
+	a.write("all services online")
+	return nil
```
charlierudolph: `startDependencies` and `startServices`are very repetitive. I think the two could be dried up with something like:
```go
func startImages(imageNames []string, imageOnlineTexts map[string]string, identifier string) error {
  // identifier used when wrapping errors and for the final log message 
}
```

cheahuychou:
@charlierudolph comment addressed

exo-go/src/docker_compose/index.go
```diff
@@ -20,9 +20,9 @@ func PullAllImages(dockerComposeLocation string, log func(string)) error {
 	return processHelpers.RunAndLog(dockerComposeLocation, log, "docker-compose", "pull")
 }
 
-// RunAllImages runs all the docker images
-func RunAllImages(env []string, cwd string, log func(string)) (*processHelpers.Process, error) {
-	process := processHelpers.NewProcess("docker-compose", "up")
+// RunImages runs the given docker images
+func RunImages(images []string, env []string, cwd string, log func(string)) (*processHelpers.Process, error) {
+	process := processHelpers.NewProcess(append([]string{"docker-compose", "up"}, images...)...)
```
charlierudolph, line 25: I think this line can be simplified to `processHelpers.NewProcess("docker-compose", "up", images...)`

charlierudolph (commented):
Please merge with master as the waiting for online text changed with my process helpers PR. I want to take one more quick look after that is merged

exo-go/src/docker_compose/index.go
cheahuychou: I had already tried that. This was the error:
```
src/docker_compose/index.go:25: too many arguments in call to processHelpers.NewProcess
	have (string, string, []string...)
	want (...string)
```

charlierudolph: Danget. Come on golang. Thats just weird

exo-go/src/app_runner/index.go
```diff
@@ -63,6 +70,31 @@ func (a *AppRunner) getEnv() []string {
...
+		if err != nil {
+			return err
+		}
+		go func(role string, onlineTextRegex *regexp.Regexp) {
+			a.waitForOnlineText(process, role, onlineTextRegex)
+			wg.Done()
+		}(role, onlineTextRegex)
+	}
+	wg.Wait()
+	if err == nil {
```
charlierudolph: I know you were just copying code, but I believe this check can go away as err will never be non-nil at this point.

charlierudolph (approved):

cheahuychou:
closed in: https://github.com/Originate/exosphere/commit/114650a41c58c210805ee4f2d9b3ae621ebff121

PR #362, opened 2017-07-25
Exo-deploy cleanup

hugoesthere:
Extracted an `AppDeployer` class to prepare it for another PR coming soon, and did some small cleanup.


exo-go/src/app_deployer/index.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// NewAppDeployer is the constructor for the appDeployer class
+func NewAppDeployer(appConfig types.AppConfig, serviceConfigs map[string]types.ServiceConfig, appDir, homeDir string) *AppDeployer {
+	return &AppDeployer{
+		AppConfig:      appConfig,
+		ServiceConfigs: serviceConfigs,
+		AppDir:         appDir,
+		HomeDir:        homeDir,
+		TerraformDir:   filepath.Join(appDir, "terraform"),
```
charlierudolph: Since this is based on appDir, how about having it be a computed property `getTerraformDir`? (can make it public if needed)

charlierudolph (approved):

exo-go/src/app_deployer/index.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// NewAppDeployer is the constructor for the appDeployer class
+func NewAppDeployer(appConfig types.AppConfig, serviceConfigs map[string]types.ServiceConfig, appDir, homeDir string) *AppDeployer {
+	return &AppDeployer{
+		AppConfig:      appConfig,
+		ServiceConfigs: serviceConfigs,
+		AppDir:         appDir,
+		HomeDir:        homeDir,
+		TerraformDir:   getTerraformDir(appDir),
```
charlierudolph: I was trying to suggest removing `TerraformDir` as a property of the `AppDeployer` in favor of a function.

exo-go/src/terraform_file_helpers/index_test.go
```diff
@@ -132,12 +137,13 @@ var _ = Describe("Given an application with public and private services", func()
...
-  cluster_id            = "${module.aws.cluster_id}"
-  command               = ["exo-js"]
-  cpu_units             = "128"
-  env                   = "production"
-  memory_reservation    = "128"
-  region                = "${var.region}"
+  cluster_id    = "${module.aws.cluster_id}"
+  command       = ["exo-js"]
+  cpu           = "128"
+	desired_count = 1
```
charlierudolph: This appears to have odd whitespace

exo-go/src/app_deployer/index.go
hugoesthere: I can envision this field being needed in other functions in this package. Like in a `Destroy()` function or something. In that case does it make more sense to keep it as a struct field?

charlierudolph: I think its best to keep minimal state (less the think about). This can be computed very quickly so I don't see any reason we need to cache it. The function can be public if needed elsewhere.

charlierudolph (approved):

PR #363, opened 2017-07-25
[new exo-run 5] rebooting changed services

### LINKED ISSUE #314: rebooting changed services
> When developing code on a running Exosphere application
I want that services that I change get automatically rebooted
So that I see the effects of my edits in the running application.

- anytime a file is created, updated, or deleted, the respective service is restarted

cheahuychou:
resolves #314


exo-go/src/service_restarter/index.go
```diff
@@ -0,0 +1,79 @@
...
+	ServiceName      string
+	ServiceDir       string
+	DockerComposeDir string
+	Env              []string
+	Log              func(string)
+	watcher          *fsnotify.Watcher
+}
+
+// Watch watches the service directory for changes
+func (s *ServiceRestarter) Watch(watcherErr chan<- error) error {
```
charlierudolph: Thoughts on renaming `watcherErr` to `watchErrChannel`

exo-go/src/app_runner/index.go
```diff
@@ -12,30 +12,31 @@ import (
...
-	AppDir               string
-	homeDir              string
-	Env                  map[string]string
-	DockerConfigLocation string
+	AppConfig        types.AppConfig
+	Logger           *logger.Logger
+	AppDir           string
+	homeDir          string
+	Env              map[string]string
+	DockerComposeDir string
```
charlierudolph, line 28: In a separate PR please make this a computed property instead of saving it since it is related to appDir

exo-go/src/app_runner/index.go
```diff
@@ -113,6 +114,19 @@ func (a *AppRunner) Shutdown(shutdownConfig types.ShutdownConfig) error {
...
 // Start runs the application and returns the process and returns an error if any
 func (a *AppRunner) Start() error {
+	watcherErr := make(chan error)
+	if err := a.watchServices(watcherErr); err != nil {
+		return err
+	}
+	go func() {
+		err := <-watcherErr
+		if err != nil {
+			if err := a.Shutdown(types.ShutdownConfig{CloseMessage: "Failed to restart"}); err != nil {
```
charlierudolph: Thoughts on changing "Failed to restart" to "Error watching services for changes". Its not guaranteed the error happened while trying to restart

exo-go/src/app_runner/index.go
```diff
@@ -113,6 +114,19 @@ func (a *AppRunner) Shutdown(shutdownConfig types.ShutdownConfig) error {
...
+				a.write("Failed to shutdown")
+			}
+		}
+	}()
```
charlierudolph: I think this goroutine and the watcherErr variable should live in the `watchServices` function.

exo-go/src/docker_compose/index.go
```diff
@@ -3,28 +3,43 @@ package dockerCompose
...
 	process := processHelpers.NewProcess(append([]string{"docker-compose", "up"}, images...)...)
-	process.SetDir(cwd)
+	process.SetDir(dockerComposeDir)
 	process.SetEnv(env)
 	process.AddOutputFunc("log", log)
 	return process, process.Start()
 }
+
+// StartContainer starts the docker container of the given service
+func StartContainer(serviceName string, env []string, dockerComposeDir string, log func(string)) error {
```
charlierudolph: This is called startContainer but it calls `docker-compose restart ...`. Should it be renamed to RestartContainer?

exo-go/src/service_restarter/index.go
```diff
@@ -0,0 +1,79 @@
...
+			select {
+			case event := <-watcher.Events:
+				if isCreate(event) {
+					s.Log(fmt.Sprintf("Restarting service '%s' because %s was created", s.ServiceName, event.Name))
+				} else if isRemove(event) {
+					s.Log(fmt.Sprintf("Restarting service '%s' because %s was deleted", s.ServiceName, event.Name))
+				} else if isChange(event) {
+					s.Log(fmt.Sprintf("Restarting service '%s' because %s was changed", s.ServiceName, event.Name))
+				}
+				watcherErr <- s.restart(watcherErr)
```
charlierudolph: I don't think we should push nil into the error channel. I think we should only push errors.

exo-go/src/service_restarter/index.go
```diff
@@ -0,0 +1,79 @@
...
+
+func isCreate(event fsnotify.Event) bool {
+	return event.Op&fsnotify.Create == fsnotify.Create
+}
+
+func isRemove(event fsnotify.Event) bool {
+	return event.Op&fsnotify.Remove == fsnotify.Remove
+}
+
+func (s *ServiceRestarter) restart(watcherErr chan<- error) error {
```
charlierudolph: This function doesn't need to return an error and could just push the errors it runs into the error channel directly.

exo-go/src/service_restarter/index.go
```diff
@@ -0,0 +1,79 @@
...
+	if err := s.watcher.Close(); err != nil {
+		return err
+	}
+	if err := dockerCompose.KillContainer(s.ServiceName, s.DockerComposeDir, s.Log); err != nil {
+		s.Log(fmt.Sprintf("Docker failed to kill container %s", s.ServiceName))
```
charlierudolph: Thoughts on wrapping the error instead of logging and leaving it the listener of the error channel to log? I believe currently there are some errors pushed into the channel that aren't wrapped and also not logged because the listener on the channel does not log the error passed to it.

exo-go/src/service_restarter/index.go
charlierudolph: Also since this is passed an error channel I don't think we need to return an error and can instead pass any error into the error channel.

exo-go/features/run/restart-service.feature
```diff
@@ -0,0 +1,18 @@
...
+
+  Rules:
+  - "exo run" will restart a service when a change occurs inside the service folder
+
+  Scenario: automatically restart a service when a change occurs
+    Given I am in the root directory of the "running" example application
+    And my application has been set up correctly
+    When starting "exo run" in my application directory
+    Then it prints "all dependencies online" in the terminal
+    And it prints "all services online" in the terminal
```
charlierudolph: I think these three lines should all start with And instead of "When" / "Then" since they are preconditions. We should have only one "When" per scenario and that should be the action we are testing

exo-go/features/run/restart-service.feature
```diff
@@ -0,0 +1,18 @@
...
+    When adding a file to "mongo-service" service folder
+    Then the "users" service restarts
+    Then it prints "'users' restarted successfully" in the terminal
```
charlierudolph: Thoughts on combining these two steps? `Then the "users" service restarts` could wait for both the restarting and restarted successfully message

cheahuychou:

charlierudolph:
Please look back through my comments as a few were not addressed. In the future, it may be helpful to comment on each thread as it is addressed in order to keep better track. Also now that the logging was removed the service restarter the app runner should log any error it receives in the `watchErrorChannel`.

exo-go/src/service_restarter/index.go
```diff
@@ -61,22 +59,21 @@ func isRemove(event fsnotify.Event) bool {
 	return event.Op&fsnotify.Remove == fsnotify.Remove
 }
 
-func (s *ServiceRestarter) restart(watcherErrChannel chan<- error) error {
+func (s *ServiceRestarter) restart(watcherErrChannel chan<- error) {
 	if err := s.watcher.Close(); err != nil {
-		return err
+		watcherErrChannel <- err
```
charlierudolph: You will still need to return after pushing an error into the channel as we don't want the function to continue executing.

cheahuychou: fixed

exo-go/src/service_restarter/index.go
```diff
@@ -0,0 +1,85 @@
...
+			select {
+			case event := <-watcher.Events:
+				if isCreate(event) {
+					s.Log(fmt.Sprintf("Restarting service '%s' because %s was created", s.ServiceName, event.Name))
+				} else if isRemove(event) {
+					s.Log(fmt.Sprintf("Restarting service '%s' because %s was deleted", s.ServiceName, event.Name))
+				} else if isChange(event) {
+					s.Log(fmt.Sprintf("Restarting service '%s' because %s was changed", s.ServiceName, event.Name))
+				}
+				s.restart(watcherErrChannel)
```
charlierudolph, line 40: I think we should return here after calling restart. Since the watcher will be closed in the restart function I don't think we need to keep this goroutine running

charlierudolph (approved):

cheahuychou:
closed in: https://github.com/Originate/exosphere/commit/4c74b22e7ba6fb901a1aa594f4238215c8d44de6

PR #364, opened 2017-07-25
[new exo-run 6] make exo-run set up the app

### LINKED ISSUE #313: new "exo run" command

cheahuychou:
resolves #313


cli/features/tutorial.feature
```diff
@@ -175,11 +176,10 @@ Feature: Following the tutorial
               - '27017:27017'
             online-text: 'waiting for connections'
       """
-    When running "exo setup" in this application's directory
     And running "exo test" in this application's directory
-    Then it prints "todo-service works" in the terminal
+    # Then it prints "todo-service works" in the terminal
```
cheahuychou: I think I need help with fixing this test. So the new `exo-run` also launches the app and I think `exo-test` also does that. So I got this error if I keep the original steps:
```
exo-test  Error: Error: (HTTP code 500) server error - driver failed programming external connectivity on endpoint todo-service-test-mongo (a7bae7fbc65152f46db7f28968da29f37335add98a7586cc95313e58b1ccb6c4): Bind for 0.0.0.0:27017 failed: port is already allocated 
```
I tried not doing `exo-run`, and got this:
```
exo-test  Testing application 'todo-app'
    exo-test  html-server has no tests, skipping
    exo-run Attaching to exocom0.22.1
     exocom Exocom 0.22.1
     exocom 
     exocom 'exocom' is running
    exo-run all dependencies online
     exocom ExoCom WebSocket listener online at port 80
     exocom ExoCom HTTP service online at port 80
    exo-run exocom0.22.1 is up-to-date
    exo-run Creating html-server
todo-service  bash: node_modules/cucumber/bin/cucumber.js: No such file or directory
todo-service  
    exo-test  todo-service is broken
    exo-run Attaching to html-server
    exo-run html-server exited with code 137
    exo-run 
    exo-run exocom0.22.1 exited with code 137
    exo-test  Tests failed
```
I might be wrong but I think this is because `exo-test` is run locally instead of inside the docker image, and we don't run startup or dockerfile commands locally.

charlierudolph: As we talked about on standup, lets just remove this test.

exo-go/features/run/exit-codes.feature
```diff
@@ -0,0 +1,15 @@
+@verbose
```
charlierudolph: I believe this tag can be removed

exo-go/src/cmd/run.go
```diff
@@ -34,32 +35,43 @@ var runCmd = &cobra.Command{
...
-		appRunner := appRunner.NewAppRunner(appConfig, logger, appDir, homeDir)
+		fmt.Printf("Setting up %s %s\n\n", appConfig.Name, appConfig.Version)
+		setup, err := appSetup.NewAppSetup(appConfig, logger, appDir, homeDir)
+		if err != nil {
+			panic(err)
+		}
+		err = setup.Setup()
+		if err != nil {
+			fmt.Println("setup failed")
+			panic(err)
```
charlierudolph: Instead of printing "setup failed" and then panic-ing with the error lets wrap the error and pass the wrapped error to panic

charlierudolph (approved):

cheahuychou:

PR #365, opened 2017-07-25
exo-deploy: Run basic terraform commands

hugoesthere:
Runs basic terraform commands that initializes state and plans deployment.


exo-go/src/app_deployer/index.go
```diff
@@ -12,27 +14,51 @@ import (
...
+}
+
+func (d *AppDeployer) write(text string) {
+	err := d.Logger.Log("exo-deploy", text, true)
+	if err != nil {
+		fmt.Printf("Error logging exo-deploy output: %v\n", err)
+	}
 }
 
 func getTerraformDir(appDir string) string {
```
charlierudolph: Sorry I missed this but I expected this to be function on the `AppDeployer` struct and thus appDir wouldn't need to be passed in. `getSecretsFile` should also use this function since the secrets is in the terraform dir

exo-go/src/app_deployer/index.go
```diff
@@ -12,27 +14,51 @@ import (
...
 // NewAppDeployer is the constructor for the appDeployer class
-func NewAppDeployer(appConfig types.AppConfig, serviceConfigs map[string]types.ServiceConfig, appDir, homeDir string) *AppDeployer {
+func NewAppDeployer(appConfig types.AppConfig, serviceConfigs map[string]types.ServiceConfig, appDir, homeDir string, logger *logger.Logger) *AppDeployer {
 	return &AppDeployer{
 		AppConfig:      appConfig,
 		ServiceConfigs: serviceConfigs,
 		AppDir:         appDir,
 		HomeDir:        homeDir,
+		Logger:         logger,
 	}
```
charlierudolph: There isn't much point for having a constructor unless you are actually performing some initialization logic. If you are just assigning public properties then you don't need one as that can just be done directly

hugoesthere:

charlierudolph (approved):

exo-go/src/process_helpers/index.go
```diff
@@ -34,7 +34,9 @@ func RunAndLog(dir string, env []string, log func(string), commandWords ...strin
 	}
 	process := NewProcess(commandWords...)
 	process.SetDir(dir)
-	process.SetEnv(env)
+	if len(env) > 0 {
+		process.SetEnv(env)
+	}
```
hugoesthere: hey @charlierudolph what do you think about this? I was running into a problem where `terraform` commands couldn't find the proper aws credentials. It seems like if I pass an empty `env` array into this function then the process will be run with a literal empty mapping of env vars. Whereas if we only set the env vars when actual values are present then we avoid overwriting existing env vars?

I think Terraform may be setting my AWS credentials as env vars automatically somehow by reading my ~/.aws/credentials file, and if we pass in an empty env vars map then that will overwrite these credentials? Although I'd expect the order to be check env vars and then check credentials file. Not sure if that makes any sense, but the terraform commands worked after I made this change.

charlierudolph: It appears you can get the current env with `os. Environ()`. Thoughts on appending that with the supplied env?

hugoesthere: I like that, although that should probably be done in a separate PR after your execplus changes have been merged in?

charlierudolph: I suggest doing it in this PR. I don't think the current change is a good one. Should be a quick change to apply this and I will deal with merge conflicts in my PR since mine isn't passing yet anyway

exo-go/src/terraform_command_helpers/index.go
```diff
@@ -0,0 +1,26 @@
+package terraformCommandHelpers
```
charlierudolph: Please move this into the terraform directory

charlierudolph (approved):

hugoesthere:
Changes have been merged in: https://github.com/Originate/exosphere/commit/e83351944f7a23c1bb9c2d74dc1cfe68ae39e804

PR #366, opened 2017-07-26
[exo-deploy 1]: configure blank AWS account

hugoesthere:


exo-go/glide.lock
```diff
@@ -0,0 +1,190 @@
...
+  - internal/assertion
+  - internal/asyncassertion
+  - internal/oraclematcher
+  - internal/testingtsupport
+  - matchers
+  - matchers/support/goraph/bipartitegraph
+  - matchers/support/goraph/edge
+  - matchers/support/goraph/node
+  - matchers/support/goraph/util
+  - types
```
charlierudolph: This file should be removed as we are using the glide.lock file at the top level and not inside `exo-go`

exo-go/src/aws_helpers/terraform_helpers.go
```diff
@@ -0,0 +1,82 @@
...
+// creates s3 bucket to store terraform remote state
+func createRemoteState(currSession *session.Session, config *aws.Config, bucketName string) error {
+	s3client := s3.New(currSession, config)
+
+	hasBucket, err := hasBucket(s3client, bucketName)
+	if err != nil {
+		return err
+	}
+
+	if !hasBucket {
```
charlierudolph: Thoughts on changing this to 
```
if hasBucket {
  return nil 
}
```
Then the rest of the function can be unindented.

Same for hasTable in the createLockTable function

exo-go/src/aws_helpers/terraform_helpers.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+// InitAccount prepares a blank AWS account to be used with Terraform
+func InitAccount(bucketName, tableName, region string) error {
+	config := aws.NewConfig().WithRegion(region)
+	session := session.Must(session.NewSession())
```
charlierudolph: Lets not use Must since that panics on error. Instead we should be bubbling the error up.

exo-go/src/aws_helpers/index.go
```diff
@@ -0,0 +1,35 @@
...
+
+func hasTable(dynamodbClient *dynamodb.DynamoDB, tableName string) (bool, error) {
+	tables, err := dynamodbClient.ListTables(&dynamodb.ListTablesInput{})
+	if err != nil {
+		return false, err
+	}
+
+	for _, table := range tables.TableNames {
+		if tableName == *table {
+			return true, err
```
charlierudolph: `return true, nil` to make it more apparent we are not returning an error. Same for returning false a couple lines below

exo-go/src/application/deployer.go
```diff
@@ -19,12 +21,30 @@ type Deployer struct {
...
+	}
 
-	err := terraform.GenerateFile(d.AppConfig, d.ServiceConfigs, d.AppDir, d.HomeDir, terraformDir)
+	// Generate Terraform Scripts
+	err = terraform.GenerateFile(terraformConfig)
 	if err != nil {
 		return err
 	}
 
+	// Run Terraform commands
```
charlierudolph: I don't think these comments are really adding anything. Also similarly to the previous PR, I'd suggest removing the empty newlines in the functions

exo-go/src/aws/terraform_helpers.go
```diff
@@ -0,0 +1,82 @@
...
+		versioningStatus := "Enabled"
+		_, err = s3client.PutBucketVersioning(&s3.PutBucketVersioningInput{
+			Bucket: &bucketName,
+			VersioningConfiguration: &s3.VersioningConfiguration{
+				Status: &versioningStatus,
+			},
+		})
+	}
+
+	return err
```
charlierudolph: I think this function can be simplified by returning nil if hasBucket is true and thus we can unindent everything currently inside the if block.

Please remove the empty lines in this function as well.

A similar idea can be applied to `createLockTable` as well.

exo-go/src/aws/terraform_helpers.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	if !hasBucket {
+		_, err = s3client.CreateBucket(&s3.CreateBucketInput{Bucket: &bucketName})
+		if err != nil {
+			return err
+		}
+
+		// enable versioning
+		versioningStatus := "Enabled"
```
charlierudolph: I saw you using `aws.String` elsewhere can that be used here instead of declaring the variable?

exo-go/src/aws/terraform_helpers.go
```diff
@@ -0,0 +1,82 @@
...
+		return err
+	}
+
+	if !hasBucket {
+		_, err = s3client.CreateBucket(&s3.CreateBucketInput{Bucket: &bucketName})
+		if err != nil {
+			return err
+		}
+
+		// enable versioning
```
charlierudolph: I think this comment can be removed

hugoesthere:

exo-go/src/aws/terraform_helpers.go
```diff
@@ -0,0 +1,75 @@
...
+
+// creates dynamodb table to store terraform lock file
+func createLockTable(currSession *session.Session, config *aws.Config, tableName string) error {
+	dynamodbClient := dynamodb.New(currSession, config)
+	hasTable, err := hasTable(dynamodbClient, tableName)
+	if err != nil {
+		return err
+	}
+	if hasTable {
+		return err
```
charlierudolph: I think we should `return nil` here to make more apparent we are not returning an error

exo-go/src/terraform/index.go
```diff
@@ -11,33 +11,33 @@ import (
...
 	return err
 }
 
 // Generate generates the contents of the main terraform file given application and service configuration
-func Generate(appConfig types.AppConfig, serviceConfigs map[string]types.ServiceConfig, appDir, homeDir string) (string, error) {
+func Generate(config types.TerraformConfig) (string, error) {
 	fileData := []string{}
 
-	moduleData, err := generateAwsModule(appConfig)
+	moduleData, err := generateAwsModule(config.AppConfig, config.RemoteBucket, config.LockTable, config.Region)
```
charlierudolph: Thoughts on passing the entires config object into this method? Same for `generateDependencyModules`

charlierudolph (approved):
Good to merge after comments are addressed

PR #378, opened 2017-07-28
[exo-deploy 2]: first half of secrets management

hugoesthere:
C & R of CRUD. Will upload vendor files shortly (updating to Sierra)


exo-go/src/aws_helpers/secrets_helpers.go
```diff
@@ -0,0 +1,165 @@
...
+
+	results, err := s3client.GetObject(&s3.GetObjectInput{
+		Bucket: aws.String(secretsBucket),
+		Key:    aws.String(secretsFile),
+	})
+	if err != nil {
+		return "", err
+	}
+
+	defer results.Body.Close()
```
hugoesthere: I'm getting a linter warning that the error value here is not checked. How important is it to check the error for a reader I'm done with anyways?

exo-go/src/aws_helpers/secrets_helpers.go
```diff
@@ -0,0 +1,165 @@
...
+		return err
+	}
+
+	// create bucket if it doesn't exist
+	if !hasBucket {
+		_, err = s3client.CreateBucket(&s3.CreateBucketInput{Bucket: &secretsBucket})
+		if err != nil {
+			return err
+		}
+	}
```
charlierudolph: Can the logic for creating a bucket if it doesn't exist be extracted into its own function?

exo-go/src/aws_helpers/secrets_helpers.go
```diff
@@ -0,0 +1,165 @@
...
+			switch awsErr.Code() {
+			case s3.ErrCodeNoSuchKey:
+				return putSecretsFile(map[string]string{}, secretsBucket, region)
+			default:
+				return awsErr
+			}
+		} else {
+			return err
+		}
+	}
```
charlierudolph: Same here. Can creating the secret file if it doesn't exist also be extracted into its own function?

exo-go/src/cmd/configure.go
```diff
@@ -0,0 +1,121 @@
...
+				secretValue := prompt.StringRequired("Secret value")
+				secrets[secretName] = secretValue
+			}
+		}
+
+		secretsPretty, err := json.MarshalIndent(secrets, "", "  ")
+		if err != nil {
+			log.Fatalf("Could not marshal secrets map: %s", err)
+		}
+		fmt.Println("You are creating these secrets:\n") //nolint vet
```
charlierudolph: Can we use `Print` here with two newlines in order to avoid having to disable the linter? Same elsewhere in this file

exo-go/src/aws_helpers/secrets_helpers.go
```diff
@@ -0,0 +1,165 @@
...
+	buffer := bytes.NewBuffer(nil)
+	if _, err = io.Copy(buffer, results.Body); err != nil {
+		return "", err
+	}
+
+	return string(buffer.Bytes()), err
+}
+
+// CreateSecrets creates new secret key value pair
+func CreateSecrets(newSecrets map[string]string, secretsBucket, region string) error {
```
charlierudolph: Thoughts on defining a Secrets type which is an alias for map[string]string. We can then define methods on the custom type such as merge and hasConflictingKey.

exo-go/src/cmd/configure.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return appConfig
+}
+
+func getBucketConfig() (string, string) {
+	appConfig := getAppConfig()
+	secretsBucket := fmt.Sprintf("%s-terraform-secrets", appConfig.Name)
+	awsRegion := "us-west-2" //TODO: read from app.yml
+	return secretsBucket, awsRegion
+}
```
charlierudolph: Can getAppConfig and getBucketConfig be extracted into helper files? I'd prefer to not have much logic in the cmd files.

exo-go/src/aws_helpers/secrets_helpers.go
```diff
@@ -0,0 +1,165 @@
...
+func tfstringToMap(tfvars string) map[string]string {
+	secretsMap := map[string]string{}
+	secretPairs := strings.Split(tfvars, "\n")
+	secretPairs = secretPairs[:len(secretPairs)-1] //remove trailing empty elem
+	for _, secret := range secretPairs {
+		s := strings.Split(secret, "=")
+		secretsMap[s[0]] = s[1]
+	}
+	return secretsMap
+}
```
charlierudolph: Thoughts on creating a TFString struct which holds a string value and has FromMap to create a TFString and then has `String()` to get the underlying value and `ToMap()` to convert it to a map. It could a

exo-go/src/aws_helpers/secrets_helpers.go
```diff
@@ -0,0 +1,165 @@
...
+
+// checks if bucket exists and returns a proper error message otherwise
+// different from hasBucket which only checks if a bucket exists
+func verifyBucket(s3client *s3.S3, secretsBucket string) error {
+	hasBucket, err := hasBucket(s3client, secretsBucket)
+	if err != nil {
+		return err
+	}
+	if !hasBucket {
+		return errors.New("secrets bucket not found. Run 'exo configure' to initialize bucket")
```
charlierudolph: I don't like this. We should just create it here instead of telling them to run a command to do it.

charlierudolph (commented):
Please try to make as much of this as testable as possible. We should be able to test a lot of the methods around the tfstring and secrets map

exo-go/src/aws_helpers/secrets_helpers.go
charlierudolph: I don't like ignoring any errors. Can we run this at the end of the function instead of deferring it?

charlierudolph: Also I think the copy into a buffer may be able to replaced with `ioutil.ReadAll`

exo-go/src/aws_helpers/secrets_helpers.go
hugoesthere: I like that. I think you meant to finish your last sentence though?

exo-go/src/cmd/configure.go
hugoesthere: Extracting getAppConfig makes sense because that happens in multiple commands even outside of `exo configure`, but getBucketConfig really on returns two strings specific to `exo configure`. Is that worth extracting a file for? It really shouldn't be public to anything outside of this command as well.

exo-go/src/aws_helpers/secrets_helpers.go
charlierudolph: TFString can actually be a type string and you still define methods on it
```golang
type TFString string

func (s TFString) ToMap() {}
```

The `fromMap` could exist on the `Secrets` type and be name `ToTFString()`

exo-go/src/cmd/configure.go
charlierudolph: A lot of the secrets config is probably specific to `exo configure` as well. There will be even more commands that need to be added to this file so I'd like to extract all that we can

exo-go/src/types/tf_string.go
```diff
@@ -0,0 +1,22 @@
...
+	secretsMap := Secrets{}
+	secretPairs := strings.Split(string(str), "\n")
+	if len(secretPairs) > 1 {
+		for _, secret := range secretPairs {
+			s := strings.Split(secret, "=")
+			secretsMap[s[0]] = strings.Trim(s[1], "\"")
+		}
+	}
+	return secretsMap
+}
```
charlierudolph: Thinking about this more I don't think we need the TFString type. We could get by with just the Secrets struct and add a NewSecrets methods which takes in a string and parses it into a map (same logic as this function)

exo-go/src/types/secrets.go
```diff
@@ -0,0 +1,67 @@
...
+	if len(secretPairs) > 1 {
+		for _, secret := range secretPairs {
+			s := strings.Split(secret, "=")
+			secretsMap[s[0]] = strings.Trim(s[1], "\"")
+		}
+	}
+	return secretsMap
+}
+
+// ToTfString converts map to .tfvars string:
```
charlierudolph: Lets rename this to `TFString()` or `String()` as that is more common in the golang world.

exo-go/src/aws_helpers/index.go
```diff
@@ -20,6 +26,23 @@ func hasBucket(s3client *s3.S3, bucketName string) (bool, error) {
...
+	}
+
+	if !hasBucket {
+		_, err = s3client.CreateBucket(&s3.CreateBucketInput{Bucket: &bucketName})
+		if err != nil {
+			return err
+		}
+	}
+
+	return err
```
charlierudolph: This could be simplied with
```golang
if hasBucket {
  return nil
}
_, err = s3client.CreateBucket(&s3.CreateBucketInput{Bucket: &bucketName})
return err
```

exo-go/src/aws_helpers/index.go
```diff
@@ -33,3 +56,42 @@ func hasTable(dynamodbClient *dynamodb.DynamoDB, tableName string) (bool, error)
...
+
+	if err != nil {
+		awsErr, ok := err.(awserr.Error)
+		if ok && awsErr.Code() == s3.ErrCodeNoSuchKey {
+			return putS3Object(s3client, fileContents, bucketName, fileName)
+		}
+		return err
+	}
+
+	return err
```
charlierudolph: Please use `return nil` here. It is equivalent, but I think easier to realize that when that line is reached, the function did not encounter an error

exo-go/src/aws_helpers/index_test.go
```diff
@@ -0,0 +1,69 @@
+package awsHelper_test
+
+import (
+	"github.com/Originate/exosphere/exo-go/src/types"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("AwsHelpers", func() {
```
charlierudolph: This describe block is unneeded as you in the `AwsHelpers` suite

exo-go/src/aws_helpers/index.go
charlierudolph: Also thoughts on removing all the empty line in functions? I think its better to keep them with fewer lines and that should hopefully make us focus on extracting out helper methods as needed as you don't get the empty lines to visually separate the different parts. Same for all the files in this PR

exo-go/src/cmd/shared.go
```diff
@@ -11,3 +19,22 @@ func printHelpIfNecessary(cmd *cobra.Command, args []string) bool {
...
 }
+
+func getAppConfig() types.AppConfig {
+	appDir, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
+	appConfig, err := appConfigHelpers.GetAppConfig(appDir)
+	if err != nil {
+		log.Fatalf("Cannot read application configuration: %s", err)
```
charlierudolph: This function should return an error instead of log.Fatalf or panic

exo-go/src/aws_helpers/index.go
hugoesthere: Ok, will do.

hugoesthere:

exo-go/src/aws/terraform_helpers.go
```diff
@@ -24,26 +24,19 @@ func InitAccount(bucketName, tableName, region string) error {
...
-		// enable versioning
-		versioningStatus := "Enabled"
-		_, err = s3client.PutBucketVersioning(&s3.PutBucketVersioningInput{
-			Bucket: &bucketName,
-			VersioningConfiguration: &s3.VersioningConfiguration{
-				Status: &versioningStatus,
-			},
-		})
-	}
+	// enable versioning
```
charlierudolph: I don't think this comment is needed.

exo-go/src/aws/terraform_helpers.go
```diff
@@ -24,26 +24,19 @@ func InitAccount(bucketName, tableName, region string) error {
...
+	versioningStatus := "Enabled"
```
charlierudolph: Can you use aws.String() instead of the `versioningStatus` variable?

exo-go/src/cmd/configure.go
```diff
@@ -0,0 +1,124 @@
...
+		newSecrets := map[string]string{}
+		secretName := prompt.String("Secret name")
+		if secretName != "" {
+			if _, hasKey := existingSecrets[secretName]; hasKey {
+				fmt.Printf("Secret for '%s' already exists. Use 'exo configure update' to update.\n\n", secretName)
+			} else {
+				secretValue := prompt.StringRequired("Secret value")
+				newSecrets[secretName] = secretValue
+			}
+		}
```
charlierudolph, line 91: There appears to be the same code block here and below. Can we remove this by making the for loop below a forever loop and just have a break statement when secretName is empty?

exo-go/src/types/secrets.go
```diff
@@ -0,0 +1,49 @@
...
+	"strings"
+)
+
+// Secrets map contains maps from secret keys to values
+type Secrets map[string]string
+
+// NewSecrets creats a Secrets map from a .tfvars string
+// a="b"
+// c="d" ->
+// {a:b, c:d}
```
charlierudolph: Is this comment needed now that we have the tests for it? I also don't think its very clear. I would say the example needs to be written as:

```
Input: `a="b"\nc="d"`
Output: Secrets{"a": "b", "c": "d"}
```

hugoesthere: I'll just remove the comment now that we have tests

hugoesthere: Well, since the golinter requires we have comments I'll leave it but change it to be more clear.

charlierudolph:
Code looks great. Sorry missed this before, but we should have some feature tests for this.

charlierudolph:
Nevermind about the feature tests, forgot this touches aws. Good to merge

charlierudolph (approved):

Issue #374, opened 2017-07-28
go: update package names


charlierudolph:
Update package names to be have no dash / underscore.

See https://blog.golang.org/package-names

Some possible updates:
* test_helpers -> test/helpers, test/featurectx
* app_*, service_config_helpers, service_restarter -> application
* docker_* -> docker
* os_helpers (move into util)
* template_helpers -> template
* terraform_file_helpers -> terraform

Ideally do one package consolidation / update at a time to keep reviews easy and get it merged quickly to prevent conflicts.

kevgo:
👍 have to refactor the helpers a bit in my opinion as well. I think we have to understand better how to do those idiomatically in Go.

PR #385, opened 2017-07-28
rename package: logger -> application

charlierudolph:
part of #374


exo-go/src/docker_setup/index.go
```diff
@@ -19,7 +18,6 @@ type DockerSetup struct {
 	ServiceConfig types.ServiceConfig
 	ServiceData   types.ServiceData
 	Role          string
-	Logger        *logger.Logger
```
charlierudolph, line 22: This was unused

cheahuychou (approved):

PR #393, opened 2017-07-31
extract templates into their own repos

charlierudolph:
* Extracted the templates into their own repos
* Update the tutorial to use a single scenario instead of multiple since they share state
* Update tests to use a os temp directory
  * got an error when using a local `tmp` directory (adding a submodule into a gitignored folder)

https://github.com/Originate/exosphere-htmlserver-express
https://github.com/Originate/exoservice-js-mongodb


cheahuychou (approved):

exo-go/features/tutorial.feature
```diff
@@ -15,14 +15,18 @@ Feature: Following the tutorial
...
 
 
-  Scenario: verify that exo commands can be run by running "exo version"
-    When starting "exo version" in the terminal
-    Then I see "Exosphere-Go" in the terminal
-
-
-  Scenario: setting up the application
+  Scenario: tutorial
+    ########################################
```
kevgo, line 19: Cool ideas to use comments to structure this a bit!

PR #394, opened 2017-07-31
rename packages - docker

charlierudolph:


exo-go/src/application/docker_compose_builder.go
```diff
@@ -1,18 +1,18 @@
...
 	"github.com/Originate/exosphere/exo-go/src/config"
-	"github.com/Originate/exosphere/exo-go/src/docker_helpers"
+	"github.com/Originate/exosphere/exo-go/src/docker"
 	"github.com/Originate/exosphere/exo-go/src/types"
 	"github.com/Originate/exosphere/exo-go/src/util"
 )
 
-// DockerSetup renders docker-compose.yml file with service configuration
-type DockerSetup struct {
+// DockerComposeBuilder renders docker-compose.yml file with service configuration
```
cheahuychou: I think this comment is misleading since it doesn't render `docker-compose.yml`. It just gets the docker configs for the given service. I think I copied this from the livescript code. Sorry about that.

exo-go/src/application/initializer_test.go
```diff
@@ -50,56 +46,48 @@ var _ = Describe("Setup", func() {
...
 
-	It("should generate an image name for each dependency and external service", func() {
+		By("generate an image name for each dependency and external service")
 		for _, serviceName := range util.JoinStringSlices(internalDependencies, externalDependencies, externalServices) {
 			Expect(len(dockerCompose.Services[serviceName].Image)).ToNot(Equal(0))
 		}
-	})
 
-	It("should generate a container name for each service and dependency", func() {
+		By("should generate a container name for each service and dependency")
```
kevgo, line 66: What is the reasoning behind using `By` here? I think it reads strange:
> Describe Initializer by should generate a container name for ...

Why not stick to the good old BDD style:
> Describe Initializer: it generates a container name for ...

charlierudolph: We are using http://onsi.github.io/ginkgo/#documenting-complex-its-by because the original beforeSuite takes 10 seconds. We moved from beforeSuite because ginkgo only allows a single beforeSuite for each package and this now lives in a package with other files. We could move this test into their own package in order to use the beforeSuite in the way we were before.

kevgo: I see. Cool solution! This saves us from running the same 10s beforeSuite multiple times. Nice!

PR #397, opened 2017-08-01
[exo-deploy 3]: update and delete secrets

hugoesthere:
UD of CRUD.


exo-go/src/types/secrets.go
```diff
@@ -45,3 +45,11 @@ func (s Secrets) Merge(newSecrets Secrets) Secrets {
 	}
 	return s
 }
+
+// DeleteSecrets deletes secrets from s. Ignores them if they do not exist
+func (s Secrets) DeleteSecrets(toDelete []string) Secrets {
```
charlierudolph: I think this method should be called `Delete` instead of `DeleteSecrets`

exo-go/src/cmd/configure.go
```diff
@@ -110,8 +111,106 @@ var configureCreateCmd = &cobra.Command{
...
+		}
+
+		secretKeys := []string{}
+		for true {
+			secretName := prompt.String("Secret name (leave blank to finish prompting)")
+			if secretName == "" {
+				break
+			} else {
+				secretKeys = append(secretKeys, secretName)
+			}
```
charlierudolph: Thoughts on having the user continually choose what secret they want to delete instead of entering secret names (there would be an option for done)? This may be unwieldy if the user has a lot of secrets. Do you have an estimate for the number of secrets a user will have? Otherwise we should probably validate that they are deleting a secret that exists.

hugoesthere: I don't have a strong feeling either way. There shouldn't be too many secrets, only things like db credentials and api keys. The AWS client already ignores any secrets you try to delete that don't exist. Even if we write our own validator that seems like a lot of extra overhead just to essentially ignore the secret as well right (which AWS is already doing)?

charlierudolph: > just to essentially ignore the secret as well right

Thats not what I think we would be doing. I'm expecting a similar error message that we have with update when the secret does not exist

exo-go/src/cmd/configure.go
```diff
@@ -110,8 +107,117 @@ var configureCreateCmd = &cobra.Command{
...
+		}
+
+		secretsString, err := aws.ReadSecrets(secretsBucket, awsRegion)
+		if err != nil {
+			log.Fatalf("Cannot read secrets: %s", err)
+		}
+		existingSecrets := types.NewSecrets(secretsString)
+		existingSecretKeys := existingSecrets.Keys()
+		fmt.Print("Existing secrets:\n\n")
+		prettyPrint(existingSecrets)
```
charlierudolph: This chunk of code (definition of ReadSecrets to prettyPrint) appears duplicated between update and delete. Getting the bucket could even be included in that as the error message should probably say `Cannot get AWS secrets config:` instead of `Cannot update/delete secrets:`

charlierudolph (approved):
Good to merged after drying up some duplication. In a follow up PR, I think it would be nice to update the secrets_helpers to a struct to reduce the amount of env variables we are passing around in the commands.

PR #399, opened 2017-08-02
[exo-test - 2] update config structs and helpers

cheahuychou:
- Add `Test` to `types.ServiceConfig` and update the type of `DockerConfig.Build` so that we can provide `context` to the `docker-compose` command.
- Add a few helper functions so we don't have to keep calling `config.NewAppDependency` and passing a lot of variables around.


exo-go/src/application/docker_compose_builder.go
```diff
@@ -14,18 +14,31 @@ import (
...
-	AppConfig     types.AppConfig
-	ServiceConfig types.ServiceConfig
-	ServiceData   types.ServiceData
-	Role          string
-	AppDir        string
-	HomeDir       string
+	AppConfig                types.AppConfig
+	ServiceConfig            types.ServiceConfig
+	ServiceData              types.ServiceData
+	BuiltAppDependencies     map[string]config.AppDependency
```
charlierudolph, line 20: I don't think we need `Built` in this property name. Same for `BuiltServiceDependencies`. Applies throughout this PR.

exo-go/src/types/docker_config.go
```diff
@@ -4,7 +4,7 @@ package types
 // docker-compose.yml
 type DockerConfig struct {
 	Image         string            `yaml:",omitempty"`
-	Build         string            `yaml:",omitempty"`
+	Build         map[string]string `yaml:",omitempty"`
```
charlierudolph, line 7: Was this defined wrong originally? This also feels like something that should be updated in its own PR

exo-go/src/application/docker_compose_builder.go
cheahuychou: I agree that is not a good variable name but this is tricky because we have `types.Dependencies` and `types.Dependency` which refer to the fields in `types.AppConfig`. `config.AppDepency` refers the type that has all functions we need for compiling docker config. There are [cases](https://github.com/Originate/exosphere/blob/e4696af7e6ce49b7c9176f2a952594dd22ccff4a/exo-go/src/application/docker_compose_builder.go#L39) where I need to iterate through both `appConfig.Dependencies` or `serviceConfig.Dependencies` and `builtDependencies` so having different names avoids confusion.

exo-go/src/types/docker_config.go
cheahuychou: It wasn't wrong, this `docker-compose` field could either be a string or a map, see [here](https://docs.docker.com/compose/compose-file/#context). Since  `exo-test` uses different `Dockerfile`s from `exo-run`. I need those fields to provide the right context for the commands in the `Dockerfile`s because I ran into some permission errors.

> This also feels like something that should be updated in its own PR

You're right but I thought it also made sense to have one PR for updating the `types` and config helpers functions so that we can handle running tests. I didn't expect the latter to be this much code. Sorry about that.

exo-go/src/application/docker_compose_builder.go
charlierudolph: Could we rename `types.Dependencies` and `types.Dependency` to make it more clear that they are configuration structs?

exo-go/src/types/docker_config.go
hugoesthere: Are we calling the Dockerfile inside the test folder something else? Can't we just point `build` to `./service` vs `./service/tests`?

exo-go/src/application/docker_compose_builder.go
cheahuychou: What would the names be?

charlierudolph: `DependenciesConfig` and `DependencyConfig`. I believe we already have `DependencyConfig` which could be renamed to `DependencyConfigOptions`? We could do this rename in a separate PR and get it merged before this one.

exo-go/src/types/docker_config.go
cheahuychou: I'm not sure if I understood you correctly. Did you mean setting `build` to `../..` (since it needs to be relative to `docker-compose.yml`)? I don't think that works since `docker-compose` will try to look for a `service/Dockerfile`. If we set `build` to `../tests`, it will find the `Dockerfile` but it doesn't have access to `package.json` which is in `../..`.  I believe the user will also have to change the paths in `Dockerfile` (e.g. `npm install ../package.json`).

hugoesthere: Yes that's what I meant. Set `build` to `../tests` for `exo tests` and `./service` for `exo run`. The user should have to copy `package.json` from its parent directory and then run install. (It shouldn't be running `npm install` like that, see: https://github.com/Originate/space-tweet/blob/master/users-service/Dockerfile#L4). But regardless, the test Dockerfile will live under `service/tests` and reference files in `service/`, so its build commands should reflect that.

hugoesthere: Nevermind. In order to run build a Dockerfile that lives in a different directory than its context, we must specify both `context` (where the Dockerfile is run) and the `dockerfile` path (where it lives). This is the proper implementation.

cheahuychou:
@charlierudolph has agreed to do the renaming in a separate PR. This PR is ready for review.

exo-go/test_helpers/shared_feature_context.go
```diff
@@ -32,6 +32,7 @@ func SharedFeatureContext(s *godog.Suite) {
 		if err != nil {
 			panic(err)
 		}
+		appDir = ""
```
charlierudolph: since this is in a beforeScenario I don't think its needed here any longer

exo-go/src/config/app_config_helpers.go
```diff
@@ -6,42 +6,40 @@ import (
...
+	for _, dependency := range appConfig.Dependencies {
+		builtDependency := NewAppDependency(dependency, appConfig, appDir, homeDir)
+		result[dependency.Name] = builtDependency
+	}
+	return result
 }
 
 // GetEnvironmentVariables returns the environment variables of
 // all dependencies listed in appConfig
 func GetEnvironmentVariables(appConfig types.AppConfig, appDir, homeDir string) map[string]string {
```
charlierudolph: Based on its usage I think GetEnvironmentVariables should take in the appBuiltDependencies instead of building them again.

charlierudolph (approved):
after comment is addressed

exo-go/src/application/docker_compose_builder.go
charlierudolph: I'll extract this into another issues

cheahuychou:
closed in https://github.com/Originate/exosphere/commit/4e5d3bdd7feb9795fc01eb208438fc8fd23562ba

PR #400, opened 2017-08-02
[exo-test - 3] write service_tester in Go

cheahuychou:


cheahuychou:
This PR is ready for review. I'm not sure why the unit tests [don't produce any output ](https://travis-ci.org/Originate/exosphere) when I use `io.Pipe()` for the `mockLogger`and [work fine](https://travis-ci.org/Originate/exosphere/builds/260782688?utm_source=github_status&utm_medium=notification) when I set `Writer` to `os.Stdout`.

cheahuychou:
Talked to @charlierudolph and decided to remove the unit tests since the feature tests in #408 are testing the same thing.

exo-go/test_helpers/shared_feature_context.go
```diff
@@ -32,6 +32,7 @@ func SharedFeatureContext(s *godog.Suite) {
 		if err != nil {
 			panic(err)
 		}
+		appDir = ""
```
charlierudolph: I believe this can be removed

exo-go/src/application/service_tester.go
```diff
@@ -0,0 +1,129 @@
...
+		return err
+	}
+	if err := s.Shutdown(types.ShutdownConfig{CloseMessage: "killing test containers\n"}); err != nil {
+		return err
+	}
+	if exitCode > 0 {
+		if err := s.Runner.Logger.Log("exo-test", fmt.Sprintf("'%s' tests failed", s.Role), true); err != nil {
+			return err
+		}
+		return fmt.Errorf("'%s' tests failed", s.Role)
```
charlierudolph: Why does this return an error? If we need to know success or no we should return two variables. I don;'t like that we are mixing general errors with whether or not the tests passed.

exo-go/bin/spec
```diff
@@ -3,5 +3,5 @@ set -e
 
 bin/lint
 bin/build
-bin/unit-test
 bin/feature-test
+bin/unit-test
```
charlierudolph: Does this still need to be switched or can this be reverted?

exo-go/src/application/runner.go
```diff
@@ -77,18 +77,22 @@ func (r *Runner) getEnv() []string {
 	return formattedEnvVars
 }
 
-func (r *Runner) runImages(imageNames []string, imageOnlineTexts map[string]string, identifier string) error {
+func (r *Runner) runImages(imageNames []string, imageOnlineTexts map[string]string, identifier string) (string, error) {
+	if len(imageNames) == 0 {
+		r.logChannel <- fmt.Sprintf("all %s online", identifier)
+		return "", nil
+	}
```
charlierudolph: I'd prefer we avoid calling runImages with no imageNames instead of having the logic in this function. Nothing came online, so seems wrong to print out "all ... online"

exo-go/src/application/runner.go
```diff
@@ -137,9 +141,11 @@ func (r *Runner) waitForOnlineText(cmdPlus *execplus.CmdPlus, role string, onlin
 	if err != nil {
 		fmt.Printf("'%s' failed to come online after an hour", role)
 	}
-	err = r.Logger.Log(role, fmt.Sprintf("'%s' is running", role), true)
-	if err != nil {
-		fmt.Printf("Error logging '%s' as online: %v\n", role, err)
+	if role != "" {
```
charlierudolph: When is role an empty string?

exo-go/src/application/service_tester.go
cheahuychou: I also didn't like that. How about just logging the logger error here and still returning an error if the tests failed and nil if the tests passed?

cheahuychou: Actually, we would have to print instead of doing `Logger.Log` because that also returns an error.

cheahuychou: Ok, came up with a different way

charlierudolph: We appear to be talking about different things. I'm not talking about the logger error but the fact that this returns `fmt.Errorf("'%s' tests failed", s.Role)`. I would expect we would be returning nil here.

exo-go/src/application/runner.go
cheahuychou: I don't know if you remember coming to help me with waiting for text in `docker-compose` log. So we want to wait for `"[service-name] exited with code"` which is output by `docker-compose` itself. So in the log channel, we don't have the prefix `exo-test` yet, therefore the role [here](https://github.com/Originate/exosphere/blob/d3288cbc8662d6c12fef67bf730417703c91ab8f/exo-go/src/application/service_tester.go#L83) needs to be an empty string.

charlierudolph: Okay. Thoughts on updating this to have 
```
if role == "" { 
  return
}
```
That prevents the logging code from being indented an extra level

cheahuychou:

exo-go/src/application/service_tester.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if err := s.Shutdown(types.ShutdownConfig{CloseMessage: "killing test containers\n"}); err != nil {
+		return testPassed, err
+	}
+	var message string
+	if testPassed {
+		message = fmt.Sprintf("'%s' tests passed", s.Role)
+	} else {
+		message = fmt.Sprintf("'%s' tests failed", s.Role)
+	}
```
charlierudolph: I think this can be further dried up with something like
```go
resultString := 'failed'
if testPassed {
  resultString = 'passed'
}
```

charlierudolph (approved):

exo-go/src/application/service_tester.go
cheahuychou: Are you referring to not using `else` and/or changing the message to just be "passed" or "failed"? Or you meant do that and do `fmt.Sprintf` later?

charlierudolph: Both. The current branching still has a good amount of duplication. You should be able to get by with a single `fmt.Spintf`

cheahuychou:
closed in https://github.com/Originate/exosphere/commit/c16ee503a6b9e2f4e13615fe1d1941bfd7b3e3c3

PR #404, opened 2017-08-02
remove service.yml setup command

charlierudolph:
This is no longer used. The tutorial test will fail until I update the templates (which would cause master to fail until this is merged). Once I get a review here, I will do that.

Related: Thoughts on updating template support in order to specify a tag / commit sha? That way you don't always install the latest on the default branch but can install a particular version of the repo.

### DIFF
--- a/exo-go/features/tutorial.feature
+++ b/exo-go/features/tutorial.feature
@@ -87,10 +87,6 @@ Feature: Following the tutorial
     description: serves HTML UI for the test app
     author: test-author
 
-    # defines the commands to make the service runnable:
-    # install its dependencies, compile it, etc.
-    setup: yarn install
-
     # defines how to boot up the service
     startup:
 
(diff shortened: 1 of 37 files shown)


kevgo:
You mean install a certain older (stable) version of templates? Why can't we always install the latest version from master?

charlierudolph:
Yes. Maybe in the case that master has a bug. It would be useful just for our testing purposes so we are using a stable version of the template. Thus when the template gets updated, it can't break the exosphere tests.

kevgo:
Hmm, good idea. Maybe make it an optional argument?

charlierudolph:
Yeah thats what I was thinking

hugoesthere (approved):
👍 

I'm for attaching commit shahs to templates as well.

charlierudolph:
Going to close since this is blocked by #407

PR #405, opened 2017-08-03
[exo-test - 1] clean up containers started by exo-run feature tests

cheahuychou:
The `exo-run` feature tests start `exo run` command but never exit it properly (because we only wait until `"all services online"`), therefore we end up with a lot of containers every time we run the tests. I added a `AfterScenario` that kills all the containers created by the app that gets run in the scenario. 

@charlierudolph released a new version of `exocom0.22.1` that turns out to be not working properly. So I had to change everything to `0.21.8` for now.


exo-go/bin/spec
```diff
@@ -3,5 +3,5 @@ set -e
 
 bin/lint
 bin/build
-bin/unit-test
 bin/feature-test
+bin/unit-test
```
charlierudolph: Is this change still needed?

charlierudolph (commented):
What prompted the change to all the timeouts?

cheahuychou:
I ran into timeout errors because I believe we were sharing containers (e.g. exocom and mongo) between feature tests. Now that we are removing containers, we need to give each test more time to start the containers.

exo-go/bin/spec
cheahuychou: Still investigating. Apparently, there is a bug in the implementation of `exo-clean` feature tests.

charlierudolph:
@hugobho can you speak to this? Are containers reused?

hugoesthere:
I have no knowledge of whether containers were used or reused between tests. As long as we are stopping and removing containers and then starting fresh ones between each test then I think that's the best implementation.

As for this PR, I think this is definitely getting rid of the orphaned containers issue we were running into right @cheahuychou? The only problem now is that it's exposing some bugs we had in our `exo clean` tests. For instance we are running `exo run` twice in order to create dangling images, but not doing any cleanup after either process. We also aren't creating dangling volumes quite right. I'm not sure if that's related to the errors we're seeing on Travis though.

exo-go/test_helpers/clean_feature_context.go
```diff
@@ -31,7 +31,10 @@ func startAndRemoveContainer(dockerClient *client.Client, imageName string) erro
 	if err != nil {
 		return err
 	}
-	return dockerClient.ContainerRemove(ctx, createdBody.ID, types.ContainerRemoveOptions{})
+	if err := dockerClient.ContainerStart(ctx, createdBody.ID, types.ContainerStartOptions{}); err != nil {
```
hugoesthere: We should also mount a volume here.

hugoesthere: Nevermind, let's go with the steps we discussed over Slack for creating dangling images and volumes.

cheahuychou:
So `AfterScenario` was returning this error for the `exo-clean` scenario:
```
Error:exit status 1
Output:The EXOCOM_PORT variable is not set. Defaulting to a blank string.
Stopping mongo3.4.0 ... 
Stopping exocom0.21.8 ... 
oving mongo3.4.0 ... 
Removing exocom0.21.8 ... 
oving network tmp_default
network tmp_default has active endpoints
```
I also found a bug in `s.Step(^my machine has both dangling and non-dangling Docker images and volumes$`) and `startAndRemoveContainers`. We were running `exo-run` twice which wasn't necessary and led to that strange error. So @hugobho came up with a different way of  creating dangling images and volumes that doesn't leave us with that active-endpoints error. 

I've removed the wait time changes.

exo-go/test_helpers/shared_feature_context.go
```diff
@@ -33,6 +33,7 @@ func SharedFeatureContext(s *godog.Suite) {
 		if err != nil {
 			panic(err)
 		}
+		appDir = ""
```
charlierudolph: Lets do this in a `BeforeScenario` instead of `BeforeSuite`

charlierudolph (approved):

hugoesthere (approved):

cheahuychou:
closed in https://github.com/Originate/exosphere/commit/4ea377eda02e2beefc47a9d581110dfc7fd98deb

PR #408, opened 2017-08-03
[exo-test - 4] write exo-test in Go

### LINKED ISSUE #368: port exo-test to go
* Tests should be run in the same docker image as used for `exo run` (will most likely share the same setup code)
* Add support for multiple test commands (this is in order to support extra setup needed since the docker image may only install the production dependencies and not the test dependencies)
  * Note: it may be worthwhile down the line to have them define a separate dockerfile for tests

@hugobho @kevgo Please review criteria and then we can start work

cheahuychou:
resolves #368 

`exo-test` now runs the test command defined `service.yml` inside a docker container. Users are expected to specify the necessary setup commands (e.g. for installing dev developments and binaries and setting paths etc.) inside a separate test `Dockerfile` inside `service-name/tests`. Currently, the command only supports running tests for the entire application. The livescript implementation supports running tests for a specific service. That feature could easily be added in a separate PR.


example-apps/tests-cucumber/exoservice/README.md
```diff
@@ -0,0 +1,2 @@
+# 
+>
```
charlierudolph: Lets make the example app have the minimum needed for testing. I believe we can remove this file

example-apps/tests-cucumber/exoservice/dependency-lint.yml
```diff
@@ -0,0 +1,31 @@
...
+requiredModules:
+  files:
+    dev:
+      - '{features,spec,test}/**/*.js'
+      - '**/*{.,_,-}{spec,test}.js'
+    ignore:
+      - 'node_modules/**/*'
+    root: '**/*.js'
+  stripLoaders: false
+  transpilers: []
```
charlierudolph: I believe this file is unneeded

example-apps/tests-cucumber/exoservice/features/support/world.js
```diff
@@ -0,0 +1,4 @@
+function World() {
+}
+
+module.exports = World;
```
charlierudolph: I believe we can delete this file and update features/support/env.js to not require the file

example-apps/tests-cucumber/exoservice/src/server.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+
+  beforeAll: (done) => {
+    // TODO: add asynchronous init code here, or delete the whole block
```
charlierudolph: please remove this comment and the beforeAll as a whole since we aren't using it

exo-go/test_helpers/shared_feature_context.go
```diff
@@ -131,7 +131,7 @@ func SharedFeatureContext(s *godog.Suite) {
 
 	s.Step(`^it prints "([^"]*)" in the terminal$`, func(text string) error {
 		if childCmdPlus != nil {
-			return childCmdPlus.WaitForText(text, time.Minute)
+			return childCmdPlus.WaitForText(text, time.Minute*2)
```
charlierudolph: What caused the need for this change?

exo-go/src/application/tester.go
```diff
@@ -0,0 +1,79 @@
...
+	initializer, err := NewInitializer(a.AppConfig, a.Logger, a.AppDir, a.homeDir)
+	initializer.logChannel = a.logChannel
+	if err != nil {
+		return err
+	}
+	runner, err := NewRunner(a.AppConfig, a.Logger, a.AppDir, a.homeDir)
+	if err != nil {
+		return err
+	}
+	runner.logChannel = a.logChannel
```
charlierudolph: I'd prefer we didn't do this. I think we can remove the need by passing in an extra argument to runner. Same with Initializer

exo-go/src/application/tester.go
```diff
@@ -0,0 +1,79 @@
...
+		return &Tester{}, err
+	}
+	return &Tester{
+		AppConfig:              appConfig,
+		InternalServiceConfigs: internalServiceConfigs,
+		ServiceData:            config.GetServiceData(appConfig.Services),
+		AppDir:                 appDir,
+		homeDir:                homeDir,
+		Logger:                 logger,
+		logChannel:             logger.GetLogChannel("exo-test"),
```
charlierudolph: I believe if we stop overriding the log channels of created instances of runner and initializer we won't need this anymore and can just use the logger directly.

exo-go/test_helpers/shared_feature_context.go
cheahuychou: We don't know the order in which the tests are run (it is random I believe). So when we wait for text [here](https://github.com/Originate/exosphere/blob/3cc763de668a7096a772303cc3cf0f301d862609/exo-go/features/test/test.feature#L11), we might end up giving up on waiting because `tweets-service` gets run before `users-service`.

charlierudolph: Auhh okay. Lets make a new step for this then so that we don't inflate the timeout for other steps. We could probably join those 5 steps into one with is something like
```
Then I eventually see the following snippets:
  | Testing service 'users-service'  |
  | 'users-service' tests passed     |
  | Testing service 'tweets-service' |
  | 'tweets-service' tests passed    |
  | All tests passed                 |
```

cheahuychou:

exo-go/test_helpers/test_feature_context.go
```diff
@@ -0,0 +1,23 @@
+package testHelpers
+
+// nolint gocyclo
```
charlierudolph, line 3: This should not be needed here

exo-go/test_helpers/test_feature_context.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"time"
+
+	"github.com/DATA-DOG/godog"
+	"github.com/DATA-DOG/godog/gherkin"
+)
+
+// nolint gocyclo
```
charlierudolph: This should not be needed

charlierudolph (approved):

PR #412, opened 2017-08-06
Support multiple GOPATHs

kevgo:
It is possible to define multiple GOPATHs (see https://github.com/Originate/guide/blob/master/go/install.md#linux). This PR makes the `build` command use the last GOPATH it finds.

### DIFF
--- a/exo-go/bin/build
+++ b/exo-go/bin/build
@@ -2,4 +2,5 @@
 set -e
 
 bin/build_bin_data
-go build -o $GOPATH/bin/exo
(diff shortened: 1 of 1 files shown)


charlierudolph:
Thoughts on renaming the `exo-go` folder to `exo` and then we can run `go install` instead of `go build` with the output path? That way we don't need to parse the go path.

kevgo:
That's fine with me as well. We can also move all content from the `exo-go` folder to the main directory if you ask me.

charlierudolph:
closing in favor of #434

PR #413, opened 2017-08-06
Refactor Exosphere architecture

kevgo:
This PR addresses an action item from the last retro: improve the code architecture and factoring.

### addressed problems
- too many "trash can" packages called "util", "helpers", "tools" etc. Such names say nothing. Code should be organized by function, i.e. one package for Docker, one for AWS, one for deployment, etc. These packages then contain the structs, methods, and helpers for that type of functionality. 
- Having such large packages aggravates the circular dependency problem: since they do too much, they are imported by too many other packages
- files are too large (too many loc)

### changes made

- removes the `types` package, packages now contain structs and methods together
- splits up the `util` package into more focussed helper libraries that are easier extracted
- extract all standalone functions into their own file

### changes not done
- refactoring of testing and deployment logic, since they are still being worked on
- no actual code changes in this PR, only moving code around

### new helper packages (to be extracted into dedicated micro-libraries outside of this code base)
- `docker`: running Docker
- `dockercompose`: running DockerCompose
- `ostools`: helpers for OS things
- `runtools`: helpers for running processes
- `stringtools`: helpers for massaging strings

### application packages (these belong to the app)

- applicationdeployer: all logic for deploying apps
- applicationrunner: all logic for running an app
- applicationtester: all logic for testing an app
- aws: logic to set up an AWS cell for Terraform
- config: logic for handling config files (YML files)
- dockercomposebuilder: builds a DockerCompose file
- logger: the logger
- template: helpers for managing templates
- terraform: Terraform logic

### TODO (later, this PR is big enough)
- extract applicationcreator, applicationadder packages 
- refactor AWS into a proper HostingInitializer object inside the `applicationdeployer` package
- refactor Terraform into a proper TerraformBuilder object inside the `applicationdeployer` package
- extract some of the standalone helper packages into dedicated micro-libraries
- refactor up the `template` package into an `applicationdeployer` package


exo-go/src/applicationdeployer/deployer.go
```diff
@@ -1,18 +1,18 @@
-package application
+package applicationdeployer
```
cheahuychou: Thoughts on shortening this to `"appdeployer"`? Same for `"applicationrunner"` and `"applicationtester"`.

exo-go/src/runtools/run_and_log.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"os"
+
+	execplus "github.com/Originate/go-execplus"
+	shellwords "github.com/mattn/go-shellwords"
+)
+
+// AndLog runs the given command, logs the process to the given
+// channel, waits for the process to finish and returns an error (if any)
+func AndLog(dir string, env []string, logChannel chan string, commandWords ...string) error {
```
cheahuychou: Thoughts on keeping the name `runtools.RunAndLog` since there is `runtools.Run`?

exo-go/src/runtools/run_series.go
```diff
@@ -0,0 +1,17 @@
+package runtools
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+)
+
+// Series runs each command in commands and returns an error if any
+func Series(dir string, commands [][]string) error {
```
cheahuychou: and keeping `runtools.RunSeries` here?

cheahuychou (commented):
This PR contains changes that "revert" the refactoring that @charlierudolph did. Originally, in my implementation we had `appRunner`, `appTester`, `logger` etc. in separated packages but Charlie merged those packages into `application`. We also had `osHelpers` and `processHelpers` etc. (I agree that word "helpers" didn't say much) which also got merged by Charlie into `util`. So I'm interested in hearing @charlierudolph's thoughts on these changes.

kevgo:
@charlierudolph as TL definitely has the final word here. This is just a proposal for the direction how I would structure this. The more important takeaways are:

- I think separating most of the business logic from the Cobra structure via a lightweight hexagonal architecture is valuable. That's the `applicationXXX` packages. This improves testability, provides the ability to swap out Cobra with something else if needed, and provides other developers a clean Go API in case they want to import and call Exosphere functionality programmatically. 
- more focused packages allow extracting functionality into more files. This makes it easier to navigate the code, and also see the test coverage better.
- many of the helper packages I have extracted are only there to be extracted further into external packages.

exo-go/src/applicationdeployer/deployer.go
kevgo: Good idea! Implemented

exo-go/src/runtools/run_and_log.go
kevgo: Good catch. Originally the package was named `run`, and then golint complained that the name `run.RunAndLog` stutters. Now that the package is named `runtools`, this is no longer an issue.

exo-go/src/runtools/run_series.go
kevgo: Fixed as well, thanks!

charlierudolph:
@kevgo can we please break this into small independent PRs which each does a piece of this? I think that are a lot of parts that are better discussed one at a time. 

Things I'm good with are: `ostools`, `runtools`, `stringtools`.

I think all the app* packages should be under `/app`. We could have `/runner`, `/deployer`, `/tester` packages inside of it. I think that follows more what I've seen in other repos. I didn't like having the large amount of top level folders, especially with the same prefix. 

I'm uncertain about `docker` as we may want to make that `dockertools` since its built on top of `docker/docker` (recently renamed to `moby/moby`). We could also have `docker/tools`, `docker/compose`, `docker/composebuilder` packages.

I like removing the `types` package and putting the structs in the package they belong to.

I liked having the logger in the `/application` as that is the only place that used it.

kevgo:
Okay thanks! Closing this PR. I have created separate issues for each refactoring.

PR #414, opened 2017-08-07
add race detector

### LINKED ISSUE #336: Add the Go race detector to our test suites
https://blog.golang.org/race-detector

Please addition to the testsuites of all Go code bases in Exosphere and Exocom.

charlierudolph:
resolves #336 

Updated go-execplus (and thus all dependencies)


kevgo:
Hmm, we still have Exocom in the vendor folder here. Why?

charlierudolph:
We were originally using it in the template that was extracted to its own repo. Removed

hugoesthere (approved):

PR #415, opened 2017-08-07
better `exo clean`

charlierudolph:
more consistent and passes locally for me versus it never use to

### DIFF
--- a/exo-go/src/cmd/clean.go
+++ b/exo-go/src/cmd/clean.go
@@ -1,9 +1,10 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


kevgo:
Nice clean-up! What do you think about keeping business logic out of the `cmd` directory? This directory should only be there to implement the Cobra framework. The logic should exist independent of which CLI framework we use. This makes it easier to update to newer versions of Cobra (which might have breaking changes), as well as switching to a different CLI framework altogether.

charlierudolph:
I think that can be dealt with separately. The amount of business logic in the cmd didn't really change here. Can you create a separate issue with a more detailed proposal of what you want the `cmd` package to look like?

hugoesthere (approved):

PR #431, opened 2017-08-08
add ability to specify template tag/commit

### LINKED ISSUE #407: templates: add commit option to hash
This now conflicts with [fetch](https://github.com/Originate/exosphere/blob/556cfe9290454021907609bc3efac492b84e64a4/exo-go/src/template/index.go#L45) which pulls the latest master. (which isn't correct to me anyway as master isn't necessarily the default branch). We could update fetch to run `git pull` on each submodule and ignore the error when the user has supplied a specific commit / tag.  

Running git pull while on a tag outputs:
```
You are not currently on a branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>

```
We could recognize that error and ignore it.

@kevgo @cheahuychou @hugobho thoughts?

charlierudolph:
resolves #407 

* Updated the fetch command to only fetch updates for a single template since those updated via a tag shouldn't/can't be updated.
* Removed some error wrapping as util.Run now gives good error messages that include the command being run and the output

### DIFF
--- a/exo-go/features/template/add_template/add_template.feature
+++ b/exo-go/features/template/add_template/add_template.feature
@@ -16,7 +16,6 @@ Feature: adding remote templates
     When running "exo template add boilr-license https://github.com/tmrts/boilr-license.git" in my application directory
     Then my application contains the directory ".exosphere/boilr-license"
     And my git repository has a submodule ".exosphere/boilr-license" with remote "https://github.com/tmrts/boilr-license.git"
-    And my git repository has a submodule ".exosphere/boilr-license" at commit "afb2fa6"
 
   Scenario: adding a new service template with a tag
     When running "exo template add boilr-license https://github.com/tmrts/boilr-license.git 0.0.1" in my application directory
--- a/exo-go/features/template/add_template/add_template.feature
+++ b/exo-go/features/template/add_template/add_template.feature
@@ -13,9 +13,15 @@ Feature: adding remote templates
 
 
   Scenario: adding a new service template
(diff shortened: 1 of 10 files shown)


exo-go/features/template/add_template/add_template.feature
```diff
@@ -13,9 +13,15 @@ Feature: adding remote templates
 
 
   Scenario: adding a new service template
-    When running "exo template add boilr-spark https://github.com/tmrts/boilr-spark.git" in my application directory
-    Then my application contains the directory ".exosphere/boilr-spark"
-    And my git repository has a submodule ".exosphere/boilr-spark" with remote "https://github.com/tmrts/boilr-spark.git"
+    When running "exo template add boilr-license https://github.com/tmrts/boilr-license.git" in my application directory
+    Then my application contains the directory ".exosphere/boilr-license"
+    And my git repository has a submodule ".exosphere/boilr-license" with remote "https://github.com/tmrts/boilr-license.git"
+    And my git repository has a submodule ".exosphere/boilr-license" at commit "afb2fa6"
```
cheahuychou: This step will fail when there are new commits in the `tmrts/boilr-license` repo, correct? I guess that's unlikely but it would be nice if we could make this step check for the latest commit instead.

charlierudolph: Hmm okay. I'll just remove this step. I wanted to show it being different from the next test case but I don't think its worth validating its the latest commit.

charlierudolph:
@hugobho @mcharawi bump

cheahuychou (approved):

PR #434, opened 2017-08-09
move src to the top level

charlierudolph:
Removes monorepo structure and morula by moving the go src to the top level


kevgo:
Thoughts around moving the entire directory content up one notch? I.e. have the `src` folder directly in the root directory?

charlierudolph:
@kevgo done

kevgo (commented):
Cool!

PR #435, opened 2017-08-10
Exo test runs in individual services

mcharawi:


exo-go/features/test/test-service.feature
```diff
@@ -0,0 +1,23 @@
...
+
+  As an Exosphere developer working with a variety of services written in different languages
+  I want to have a standardized way to run all tests for a service
+  So that I don't have to waste time investigating how to use language-specific tools.
+
+  Rules:
+  - when running "exo test" in a service directory, it runs the test for the respective service
+
+  Scenario: testing a service
+    Given I am in the "exoservice" directory of the "tests-cucumber" example application
```
charlierudolph: `test-cucumber` isn't a good example for this since it contains a single service. I'd suggest using something like the tests-passing application, where you run in one of the services. We should also validate it does not run tests in any other service.

exo-go/src/application/tester.go
```diff
@@ -38,9 +38,12 @@ func NewTester(appConfig types.AppConfig, logger *Logger, appDir, homeDir string
 }
 
 // RunAppTests runs the tests for the entire application
-func (a *Tester) RunAppTests() error {
+func (a *Tester) RunAppTests(serviceName string) error {
 	a.logChannel <- fmt.Sprintf("Testing application %s", a.AppConfig.Name)
 	numFailed := 0
+	if serviceName != "" {
+		a.InternalServiceConfigs = map[string]types.ServiceConfig{serviceName: a.InternalServiceConfigs[serviceName]}
+	}
```
charlierudolph: I don't like overwriting `a.InternalServiceConfigs`. I think we should create another public function `RunServiceTests(serviceName string)`. We could have a helper which takes something of the same type of `a.InternalServiceConfigs` which we call either with `a.InternalServiceConfigs` or the custom object you build here when running a single service

exo-go/src/cmd/test.go
```diff
@@ -21,6 +23,14 @@ var testCmd = &cobra.Command{
 		if err != nil {
 			panic(err)
 		}
+		serviceName := ""
+		if _, err = os.Stat("service.yml"); err == nil {
+			_, serviceName = path.Split(appDir)
```
charlierudolph: I don't think the directory name will always match the serviceName. See space-tweet for an example of when they are different.

exo-go/test_helpers/run_feature_context.go
```diff
@@ -40,6 +40,12 @@ func RunFeatureContext(s *godog.Suite) {
 		return CheckoutApp(cwd, appName)
 	})
 
+	s.Step(`^I am in the "([^"]*)" directory of the "([^"]*)" example application$`, func(serviceName, appDirName string) error {
+		appDir = path.Join(cwd, "tmp", appDirName, serviceName)
```
charlierudolph: This feels wrong since the appDir doesn't include the serviceName. I'd suggest using the existing step for picking what application you are in and creating a new step for something like
```
When starting "exo test" in the "exoservice" directory
```
Where the dir your run is would be appDir plus exoservice.

mcharawi:

src/application/tester.go
```diff
@@ -61,6 +58,24 @@ func (a *Tester) RunAppTests(serviceName string) error {
...
+		if testPassed, err := a.runServiceTests(serviceName, a.InternalServiceConfigs[serviceName]); err != nil {
+			a.logChannel <- fmt.Sprintf("error running '%s' tests:", err)
+		} else if !testPassed {
+			numFailed++
+		}
+	}
+	if numFailed == 0 {
+		return a.Logger.Log("exo-test", "All tests passed", true)
+	}
+	return a.Logger.Log("exo-test", fmt.Sprintf("%d tests failed", numFailed), true)
```
charlierudolph: I don't think we need some of these log messages. "Testing application" / "All tests passed" / "%d tests failed" aren't really needed since we are only running the tests for a single service.

features/test/test-service.feature
```diff
@@ -8,14 +8,11 @@ Feature: testing an Exosphere service
...
-    And I eventually see:
-    """
-    1 scenario (1 passed)
-    """
-    And it prints "'exoservice' tests passed" in the terminal
+    Given I am in the root directory of the "tests-passing" example application
+    When starting "exo test" in the "tweets-service" directory
+    Then it prints "Testing service 'tweets-service'" in the terminal
+    And it prints "'tweets-service' tests passed" in the terminal
+    And it does not print "'users-service' tests passed" in the terminal
```
charlierudolph: Thoughts on having this be `Testing service 'users-service'` instead of `'users-service' tests passed` as the latter wouldn't be printed if the user tests failed

test_helpers/shared_feature_context.go
```diff
@@ -115,6 +130,16 @@ func SharedFeatureContext(s *godog.Suite) {
 		return childCmdPlus.Start()
 	})
 
+	s.Step(`^starting "([^"]*)" in the "([^"]*)" directory$`, func(command, dirName string) error {
+		commandWords, err := util.ParseCommand(command)
+		if err != nil {
+			return err
+		}
+		childCmdPlus = execplus.NewCmdPlus(commandWords...)
+		appDir = path.Join(appDir, dirName)
```
charlierudolph: Do we need to update the appDir variable here? I think we can pass `path.Join(appDir, dirName)` directly into    `childCmdPlus.SetDir`. Again, I don't like that its being set to something thats not an application directory.

test_helpers/run_feature_context.go
```diff
@@ -40,6 +40,12 @@ func RunFeatureContext(s *godog.Suite) {
 		return CheckoutApp(cwd, appName)
 	})
 
+	s.Step(`^I am in the "([^"]*)" directory of the "([^"]*)" example application$`, func(serviceName, appDirName string) error {
+		appDir = path.Join(cwd, "tmp", appDirName, serviceName)
+		appName = appDirName
+		return CheckoutApp(cwd, appName)
+	})
```
charlierudolph: I believe this step can now be deleted since its unused

src/cmd/test.go
```diff
@@ -21,6 +26,20 @@ var testCmd = &cobra.Command{
...
+		serviceName := ""
+		if _, err = os.Stat("service.yml"); err == nil {
+			serviceConfig := types.ServiceConfig{}
+			yamlFile, err := ioutil.ReadFile(path.Join(appDir, "service.yml"))
+			if err != nil {
+				panic(err)
+			}
+			err = yaml.Unmarshal(yamlFile, &serviceConfig)
+			serviceName = serviceConfig.Type
+			appDir = path.Join(appDir, "/..")
```
charlierudolph: Thoughts on renaming the initial `appDir` variable (set from the call to `os.Getwd`) to `currentDir` and updating this branching tree to the following:
```go
if _, err = os.Stat("service.yml"); err == nil {
  // Assign serviceName 
  // Assign appDir as the parent of the currenDir 
} else if _, err = os.Stat("application.yml"); err == nil {
  // Assign appDir as currentDir
} else {
  // Print error message
}
```
I think thats more clear than assigning the currentDir as the appDir and overwriting it when that is wrong

features/test/test-service.feature
```diff
@@ -0,0 +1,20 @@
...
+
+  Scenario: testing a service
+    Given I am in the root directory of the "tests-passing" example application
+    When starting "exo test" in the "tweets-service" directory
+    Then it prints "Testing service 'tweets-service'" in the terminal
+    And it prints "'tweets-service' tests passed" in the terminal
+    And it does not print "'users-service' tests passed" in the terminal
+
+  Scenario: executing "exo test" in a directory that isn't a service or application
+    Given I am in the root directory of a non-exosphere application called "empty"
```
charlierudolph: Thoughts on removing the `called "empty"` portion of this step? I don't think thats needed

test_helpers/shared_feature_context.go
```diff
@@ -61,11 +61,26 @@ func SharedFeatureContext(s *godog.Suite) {
...
 	s.Step(`^I am in the root directory of an empty application called "([^"]*)"$`, func(appName string) error {
 		appDir = path.Join(os.TempDir(), appName)
 		return createEmptyApp(appName, cwd)
 	})
 
+	s.Step(`^it doesn\'t run any tests$`, func() error {
+		if childCmdPlus != nil {
+			return childCmdPlus.WaitForText("Not an application or service directory, exiting...", time.Minute)
+		}
+		return validateTextContains(childOutput, "Not an application or service directory, exiting...")
```
charlierudolph: Thoughts on having a variable for holding the expectedText to reduce the duplication?

mcharawi:

charlierudolph (approved):

charlierudolph:
Good to merge once green or the lint and "exo test" tests passing

PR #436, opened 2017-08-10
[exo-deploy 3] Image management

hugoesthere:
Build and push images to ECR, as well as pass their repository URIs to the Terraform scripts.


src/aws/ecr_helpers.go
```diff
@@ -0,0 +1,127 @@
...
+
+// returns a mapping from service/dependency names to image name on the user's machine
+func getImageNames(deployConfig types.DeployConfig, dockerComposeDir string, dockerCompose types.DockerCompose) (map[string]string, error) {
+	images := map[string]string{}
+	// get service image names
+	for _, serviceName := range deployConfig.AppConfig.GetServiceNames() {
+		dockerConfig := dockerCompose.Services[serviceName]
+		images[serviceName] = buildImageName(dockerConfig, dockerComposeDir, serviceName)
+	}
+	// get dependency image names
```
charlierudolph: Instead of having this comments could we break this into two functions and merge the maps?

src/aws/ecr_helpers.go
```diff
@@ -0,0 +1,127 @@
...
+		imagesMap[serviceName] = taggedImage
+		encodedAuth, err := getECRCredentials(ecrClient)
+		if err != nil {
+			return nil, err
+		}
+		err = docker.PushImage(dockerClient, taggedImage, encodedAuth)
+		if err != nil {
+			return nil, err
+		}
+	}
```
charlierudolph: Can this function be broken up a bit more. Maybe extract whats in the for loop into its own function?

charlierudolph (commented):
Code looks fine. Is there anyway we can add some tests here? Maybe `getImageNames` could be tested?

hugoesthere:
@charlierudolph comments addressed and test added.

src/aws/index_test.go
```diff
@@ -49,3 +55,46 @@ var4="val4"`
 		Expect(expectedTfString).To(Equal(actualSecrets.TfString()))
 	})
 })
+
+var _ = Describe("ECR helpers", func() {
+	It("compiles the list of image names", func() {
```
charlierudolph: Please put this in its own file `ecr_helpers_test.go` and add another `Describe` block for `GetImageNames` as that is what this it block applies to (not the entire ecr helpers)

charlierudolph (approved):

PR #440, opened 2017-08-11
Exo Test: exit with status 1 on failures

mcharawi:

### DIFF
--- a/src/application/tester.go
+++ b/src/application/tester.go
@@ -61,12 +61,12 @@ func (a *Tester) RunAppTests() (bool, error) {
 // RunServiceTest runs the tests for a single service
 func (a *Tester) RunServiceTest(serviceName string) (bool, error) {
 	testsPassed := true
+	var err error
 	if a.InternalServiceConfigs[serviceName].Tests == "" {
 		a.logChannel <- fmt.Sprintf("%s has no tests, skipping", serviceName)
 	} else {
-		if _, err := a.runServiceTests(serviceName, a.InternalServiceConfigs[serviceName]); err != nil {
+		if testsPassed, err = a.runServiceTests(serviceName, a.InternalServiceConfigs[serviceName]); err != nil {
 			a.logChannel <- fmt.Sprintf("error running '%s' tests:", err)
-			testsPassed = false
 		}
 	}
 	return testsPassed, nil
--- a/src/application/tester.go
+++ b/src/application/tester.go
@@ -2,7 +2,6 @@ package application
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/Originate/exosphere/src/config"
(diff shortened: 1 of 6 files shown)


src/cmd/test.go
```diff
@@ -71,7 +71,7 @@ var testCmd = &cobra.Command{
 				panic(err)
 			}
 		}
-
+		os.Exit(0)
```
charlierudolph: Is this needed? I think the default exit code is 0

test_helpers/shared_feature_context.go
```diff
@@ -201,6 +201,8 @@ func SharedFeatureContext(s *godog.Suite) {
 			} else {
 				return fmt.Errorf("cmd.Wait: %v", err)
 			}
+		} else {
+			return nil
```
charlierudolph: Shouldn't this branch validate that the expected exit code is 0?

charlierudolph (commented):
Please also add exit code validation to the other features in `features/test`

mcharawi:

src/application/tester.go
```diff
@@ -55,7 +56,9 @@ func (a *Tester) RunAppTests() error {
 	if numFailed == 0 {
 		return a.Logger.Log("exo-test", "All tests passed", true)
 	}
-	return a.Logger.Log("exo-test", fmt.Sprintf("%d tests failed", numFailed), true)
+	err := a.Logger.Log("exo-test", fmt.Sprintf("%d tests failed", numFailed), true)
+	os.Exit(1)
```
charlierudolph: Please no exiting from helpers. We should only have exit statements in the `cmd` directory. Please update these methods to return a bool of whether or not the tests ran successfully.

mcharawi:

src/application/tester.go
```diff
@@ -53,21 +53,23 @@ func (a *Tester) RunAppTests() error {
...
 }
 
 // RunServiceTest runs the tests for a single service
-func (a *Tester) RunServiceTest(serviceName string) error {
+func (a *Tester) RunServiceTest(serviceName string) (bool, error) {
+	testsPassed := true
 	if a.InternalServiceConfigs[serviceName].Tests == "" {
 		a.logChannel <- fmt.Sprintf("%s has no tests, skipping", serviceName)
 	} else {
 		if _, err := a.runServiceTests(serviceName, a.InternalServiceConfigs[serviceName]); err != nil {
```
charlierudolph: `runServiceTests` returns whether or not service tests pass. I believe we need to use that value. If you add a test case for running the tests for a single service where the tests fail, I believe you would currently get an exit code of 0 when we want 1.

mcharawi:

charlierudolph (approved):

Issue #441, opened 2017-08-11
feature testing: remove and kill all containers


charlierudolph:
Lets do this before starting tests since orphaned containers cause errors.

charlierudolph:
@hugobho @mcharawi @cheahuychou thoughts?

hugoesthere:
This only includes all containers listed in `docker-compose.yml` right? @mcharawi was working on adding a feature to `exo clean` that does essentially this. We'd just have to run `exo clean` against the main `docker-compose.yml` as well as the test `docker-compose.yml` file.

charlierudolph:
closing in favor of #444

charlierudolph:
This would need a disclaimer that running the exosphere feature tests will stop and remove all containers. This is because if you interrupt running feature tests you may be left with orphaned containers that can cause the exo-run to fail. @hugobho thoughts?

hugoesthere:
Who/what is stopping and removing the containers?

charlierudolph:
We would add some code to do that in a `BeforeSuite`

hugoesthere:
Would we use the new `exo clean` command which reads your `docker-compose.yml` files and removes only those containers? If so then we'll only remove containers that exosphere touches and not the ones the user has running themselves. Sure we'll stop service containers if they have `exo run` running in the background but that's not as bad as stopping and removing their non-exosphere containers.

charlierudolph:
This is only for running our feature tests when developing exopshere. I want to get our test suite reliably passing and then we can make this better about not removing all containers

hugoesthere:
Ok sure, let's just put the disclaimer in for now then.

charlierudolph:
Closing with won't fix. We can come back to this if needed

PR #442, opened 2017-08-11
better error messages

charlierudolph:
* Updated the os helpers to all return errors so we don't hide anything.
* Updated the error message for is empty directory so we see what are the contents.
* Moved cleaning up the app dir from after suite to after scenario as one error message made it appear we were sharing an app dir between a couple tests


src/docker/compose_helpers.go
```diff
@@ -17,7 +17,7 @@ func CreateNewContainer(serviceName string, env []string, dockerComposeDir strin
 
 // KillAllContainers kills all the containers
 func KillAllContainers(dockerComposeDir string, logChannel chan string) (*execplus.CmdPlus, error) {
-	cmdPlus := execplus.NewCmdPlus("docker-compose", "down")
+	cmdPlus := execplus.NewCmdPlus("docker-compose", "down", "--remove-orphans")
```
hugoesthere: Are we sure we want to do this? This could possibly remove containers on the user's machine that Exosphere doesn't manage. I know we were running into an issue with orphaned containers, is that something the new `exo clean` can fix? (Removing all containers listed in `docker-compose.yml`).

charlierudolph: I'm good removing this. Was just hitting it locally and that seemed to be what it suggested. I created #441 to do something like it in our tests.

src/cmd/add.go
```diff
@@ -83,6 +93,11 @@ var addCmd = &cobra.Command{
 	},
 }
 
+func exitWithNoTemplates() {
```
kevgo: Name this `exitBecauseNoTemplates`?

src/cmd/add.go
```diff
@@ -26,14 +26,20 @@ var addCmd = &cobra.Command{
 		if err != nil {
 			panic(err)
 		}
-		if !template.HasTemplatesDir(appDir) {
-			fmt.Println("no templates found\n\nPlease add templates to the \".exosphere\" folder of your code base.")
-			os.Exit(1)
+		hasTemplates, err := template.HasTemplatesDir(appDir)
+		if err != nil {
+			panic(err)
```
kevgo: Thoughts about using `log.Fatal` for such user errors? Looking at the Go documentation, it seems like `panic` is more for programming errors like accessing a slice index out of bounds or a failing type assertion. Therefore, it prints a stack trace of the issue to stderr, so that the bug in the code can be fixed.

`log.fatal` is for user errors. It logs to the configured log output, without a stack trace which is irrelevant and distracting in this case. 

* https://stackoverflow.com/questions/35996966/when-should-i-use-panic-vs-log-fatalln-in-golang
* https://stackoverflow.com/questions/28472922/when-to-use-os-exit-and-panic-in-golang

charlierudolph: Okay. I'll create separate issue for this.

charlierudolph: Actually. I'll still debate this. These panics are not for user errors. They are for fully unexpected errors such as the filesystem failures. I personally like the idea of having a stacktrace in the rare case that a user ever encounters them as having the stacktrace gives use a much better chance of being able to fix the problem if there is anything we need to do.. The `exitWithNoTemplates` func is an example of a user error to me (and one where we could use `log.Fatal`)

src/cmd/add.go
charlierudolph: Sounds good

src/cmd/add.go
kevgo: Yes, we should keep using panics for developer errors like  when an internal file isn't there. In the case here, unless I am misreading this, the error  is that there is no template directory in the repo. I consider that a user error, no?

charlierudolph: I believe you are misreading this. `hasTemplates` is a boolean for whether or not the user has templates and that is checked against in the next if statement. The error is a possible filesystem error that may be hit when determining this

charlierudolph: Just renamed `hasTemplates` to `hasTemplateDir`

hugoesthere (approved):

charlierudolph:
Updated with #444 as I kept running into those errors

src/cmd/add.go
kevgo: I see. Is it a programming error that requires seeing the stack track (in order to change something in the program so that we can avoid this error in the future)? In that case let's use a panic. 

Or is that a runtime error where all we can do is show the user the error message and abort? Then we should use `log.fatalf`.

PR #445, opened 2017-08-14
[exo-deploy 5] Source git urls for terraform modules

hugoesthere:


src/terraform/index_test.go
```diff
@@ -69,7 +69,7 @@ provider "aws" {
 }
 
 module "aws" {
-  source = "./aws"
+  source = "git@github.com:Originate/exosphere.git//src//terraform//modules//aws?ref=8786f912"
```
charlierudolph, line 72: Why the `//` for the file? Also should we specify a commit or anything? Currently this is pulling from latest master always

hugoesthere: `//` is the Terraform syntax to specify subdirs. The `ref=8786f912` at the end specifies the commit.

charlierudolph: Can we put the commit in a variable so if we want to update it we only need to do it in one place?

hugoesthere: I think for now when they're all the same that makes sense. But in the future it's possible modules would change individually of each other, in which case having a single variable might be problematic. Or we could be sourcing modules from different repositories.

I also like that the commit hash is hardcoded into the module themselves, so there isn't margin for error when generating or interpolating the git url.

charlierudolph: > change individually of each other

I don't know if I understand this. Why would we keep some referencing older versions of the codebase? If we only changed one, whats the harm with updating all of them?

hugoesthere: Yeah that's true. I'll update this to reflect that.

hugoesthere:
@charlierudolph a terraform commit hash const has been added.

src/terraform/index.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+// TerraformModulesCommitHash is the git commit hash that reflects
+// which of the Terraform modules in Originate/exosphere we are using
+const TerraformModulesCommitHash = "8786f912"
```
charlierudolph: I don't think this needs to be public since it appears to only appear in this package. Thus it could become: `terraformModulesCommitHash`

charlierudolph (approved):

PR #446, opened 2017-08-14
Check if file exists before removing secrets.tfvars

hugoesthere:


src/application/deployer.go
```diff
@@ -52,9 +53,12 @@ func writeSecretsFile(deployConfig types.DeployConfig) error {
...
 func RemoveSecretsFile(secretsPath string) error {
-	err := os.Remove(secretsPath)
-	if err != nil {
-		return errors.Wrap(err, fmt.Sprintf("Error removing secrets file: %s. Manual removal recommended", secretsPath))
+	if util.DoesFileExist(secretsPath) {
+		err := os.Remove(secretsPath)
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("Error removing secrets file: %s. Manual removal recommended", secretsPath))
+		}
+		return nil
```
charlierudolph: I don't believe this return is necessary. Since it will fall to `return nil` outside

charlierudolph (approved):

PR #450, opened 2017-08-15
[exo-deploy 6] Extract into two commands

hugoesthere:
Extracts `exo-deploy` into two commands: one for initializing deployment and the other for running actual deployment. Also adds step that actually runs `terraform apply`


src/cmd/deploy.go
```diff
@@ -4,57 +4,45 @@ import (
...
-	"github.com/Originate/exosphere/src/util"
 	"github.com/spf13/cobra"
 )
 
-//deployCmd represents the deploy command
-var deployCmd = &cobra.Command{
-	Use:   "deploy",
-	Short: "Deploys Exosphere application to the cloud",
-	Long:  "Deploys Exosphere application to the cloud",
+var deployInit = &cobra.Command{
```
kevgo: Extract this into a file `deploy_init.go`?

src/cmd/deploy.go
```diff
@@ -4,57 +4,45 @@ import (
...
+			log.Fatalf("Deploy initialization failed: %s", err)
 		}
-		awsConfig, err := getAwsConfig()
-		if err != nil {
-			log.Fatalf("Failed to read secrest configurations: %s", err)
+	},
+}
+
+var deployCmd = &cobra.Command{
+	Use:   "deploy",
```
kevgo: How is this command called? `exo deploy` or `exo deploy deploy`?

hugoesthere: Just `exo deploy`

src/cmd/deploy.go
hugoesthere: We've been keeping subcommands in the same file as the root command (see `cmd/configure.go`). This is a subcommand of `exo deploy` invoked by `exo deploy init`.

charlierudolph:
Whats the error message look like if you run `exo deploy` without `exo deploy init`?

hugoesthere:
@charlierudolph I've added an error message if you run `exo deploy` before running the init command, as well as tests for that. I've also changed `exo deploy init`  to `exo deploy build` since that command can be run more than once and isn't really just an initialization.

hugoesthere:
Although it seems that this could be causing a race condition. Do you know why that could be?

src/util/exec.go
```diff
@@ -28,23 +28,23 @@ func Run(dir string, commandWords ...string) (string, error) {
...
 	cmdPlus := execplus.NewCmdPlus(commandWords...)
 	cmdPlus.SetDir(dir)
 	cmdPlus.SetEnv(append(env, os.Environ()...))
 	ConnectLogChannel(cmdPlus, logChannel)
 	if err := cmdPlus.Run(); err != nil {
-		return errors.Wrapf(err, "Error running '%s'. Output:\n%s", strings.Join(commandWords, " "), cmdPlus.GetOutput())
+		return &execplus.CmdPlus{}, errors.Wrapf(err, "Error running '%s'. Output:\n%s", strings.Join(commandWords, " "), cmdPlus.GetOutput())
 	}
-	return nil
+	return cmdPlus, nil
```
hugoesthere: I think returning this `cmdPlus` object is causing some issues with race conditions and missing output logs. Is there an easier way to do this? All I want to do is listen for a string and emit an error if that string appears in the output. @charlierudolph

hugoesthere:
Nevermind, I went with a different implementation to check that `exo deploy init` has been run. Instead of grabbing the process output I just check that the `main.tf` file exists before running any Terraform commands. Should be good for another look @charlierudolph

src/terraform/commands.go
```diff
@@ -10,6 +10,9 @@ import (
 
 // RunInit runs the 'terraform init' command and force copies the remote state
 func RunInit(deployConfig types.DeployConfig) error {
+	if !util.DoesFileExist(deployConfig.TerraformFile) {
+		return errors.New("terraform initialization attempted in directory with no 'main.tf' file. Please make sure 'exo deploy build' has been run")
+	}
```
charlierudolph: At this point is running `exo deploy build` ever the wrong thing? If not can we just log something like `Missing 'main.tf' file. Initializing deployment...` and run the deployment initializer?

hugoesthere: The only problem with that is automatically running `exo deploy build` will overwrite the existing `main.tf` file. One of the points of breaking this up into `exo deploy build` and `exo deploy` is that after running the first command, the user can inspect the `main.tf` file and make or add changes, and then deploy with those modifications.

However, overwriting the `main.tf` file is a problem we will have to deal with at some point. I was thinking maybe `exo deploy init` could be passed a flag to tell it whether or not to generate a new `main.tf` file. We could even have it generate a new `main.tf` file just with a different name. @alexdavid thoughts?

kevgo:
I would prefer that if a command has subcommands, that all possible actions have to be done via subcommands. I.e. change `exo deploy` to `exo deploy run` or something. What do you think?

src/terraform/commands.go
charlierudolph: I was expecting this to **never** overwrite `main.tf` and instead only initialize it if not present.

charlierudolph:
I agree with @kevgo. I am not a fan of having `exo deploy` do something and also have subcommands

src/terraform/commands.go
hugoesthere: When the user makes changes to the application then then `exo deploy build` needs to be run again so `main.tf` can be updated, so that deployment will reflect the latest version of the app.

hugoesthere:
Yeah, that felt weird too. I'll make it `exo deploy build` and `exo deploy apply`.

src/terraform/commands.go
charlierudolph: I meant if you run `exo deploy apply` without running `exo deploy build` then it runs `exo deploy build` for you instead of telling you to run it. If the `main.tf` exists, it just applies it.

hugoesthere: Oh I see, so only when `main.tf` doesn't exist does it run `exo deploy build` automatically, otherwise it's up to the user whether or not they overwrite the existing file.

charlierudolph: Correct. I'm not certain if we want to stop after building if we want the user to be able to inspect the `main.tf` first or just apply it immediately.

src/cmd/shared.go
```diff
@@ -47,6 +51,46 @@ func getAwsConfig() (types.AwsConfig, error) {
...
+	if err != nil {
+		panic(err)
+	}
+	homeDir, err := util.GetHomeDirectory()
+	if err != nil {
+		panic(err)
+	}
+	appConfig, err := types.NewAppConfig(appDir)
+	if err != nil {
+		return types.DeployConfig{}, fmt.Errorf("Cannot read application configuration: %s", err)
```
charlierudolph: Please use [errors.Wrap](https://godoc.org/github.com/pkg/errors) instead of manually wrapping with `fmt.Errorf`

src/cmd/shared.go
```diff
@@ -47,6 +51,46 @@ func getAwsConfig() (types.AwsConfig, error) {
...
 }
 
+func getDeployConfig() (types.DeployConfig, error) {
+	appDir, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
+	homeDir, err := util.GetHomeDirectory()
+	if err != nil {
+		panic(err)
```
charlierudolph: Why are we panic-ing instead of returning the error since this function returns an error?

src/terraform/commands.go
hugoesthere: True, I'll stop the process after the build portion finishes.

features/help.feature
```diff
@@ -63,10 +63,15 @@ Feature: help command
...
-      Deploys Exosphere application to the cloud
+      Deploys an Exosphere application to the cloud
 
       Usage:
-        exo deploy
+        exo deploy [command]
+
+      Available Commands:
+        apply       Deploys an Exosphere application to the cloud
+        build       Prepares an Exosphere application to be deployed
```
alexdavid, line 73: This is only run the first time deploying right? Maybe this should be called `exo deploy init`?

hugoesthere: Thats what I called it initially, but `build` also pushes your Docker images to ECR and generates your `.tf` files, so it would theoretically be run each time the user makes changes to their app

hugoesthere:
We're going with a different implementation of `exo deploy` as we discussed.

PR #471, opened 2017-08-17
fix service tester

charlierudolph:
ran into another random failure where the tests passed/failed message was not printed when shutdown encountered an error. Update to print the result message before attempting shutdown


cheahuychou (approved):

hugoesthere (approved):

charlierudolph:
Merging this with #470

Issue #429, opened 2017-08-08
Env vars for services


hugoesthere:
Environment variable support for individual services will be provided in the form:
```
#service.yml

environment:
  development:
    public:
      ENV1: some value
      ENV2: some value
    secret:
      - ENV3
  production:
    public:
      ENV1: some value
      ENV2: some value
    secret:
      - ENV3
```

The secret keys can be set in development via https://github.com/joho/godotenv. The user will define a `.env` file containing the values of the secret env vars. Both `public` and `secret` env vars and values will be rendered into `tmp/docker-compose.yml` which is not checked in with the app. 

As for production, the `public` env vars and values will be rendered into the terraform directly, whereas the `secret` env vars will be represented as variables in the terraform, the values of which should be managed via `exo configure`. During `exo deploy` these secrets will be read from the remote store that `exo configure` manages.

hugoesthere:
@kevgo @charlierudolph thoughts?

kevgo:
What is the use case for this?

hugoesthere:
The use case is for any service that would ever need to set an environment variable. Right now exosphere provides no way of doing this. For the `secret` env vars, there could be a need for things like API keys, db credentials, etc. 

 This issue was ported over from: https://github.com/Originate/exosphere-sdk/issues/301.

hugoesthere:
@kevgo where do we land on this? This is needed for `exo-deploy`.

It would probably be a better idea for the production field of `service.yml` to look like this:
```
  production:
    public:
      ENV1: some value
      ENV2: some other value
    secret:
      ENV3: secret_key
```

Eventually we need some way to render an env var block in Terraform that looks like this:
```
environment_variables = {
  ENV1 = "some value"
  ENV2 = "some other value"
  ENV3 = "${var.secret_key}"
}
```

The value of `secret_key` is stored in the `secrets.tfvars` file we're managing for the user. That file is pulled down during `terrafrom plan/apply` and pointed to via `-var-file=secrets.tf`. This means the `secret_key` would have to be unique across services.

hugoesthere:
@alexdavid

charlierudolph:
hmm. @hugobho can we keep the structure the same between dev and production? Maybe for dev we can use 

```
public:
  ENV1: some value
  ENV2: some value
secret:
  ENV3: ENV_VARIABLE_NAME
```

hugoesthere:
[...]

PR #477, opened 2017-08-23
[exo-deploy 2] Read service env vars

hugoesthere:
Resolves the production part of https://github.com/Originate/exosphere/issues/429

This PR does two things: read service env vars and pass them in as command line flags instead of writing a `secret.tfvars` file. We avoid writing secrets to disk and having to manage/delete that file this way.


src/types/service_config.go
```diff
@@ -13,5 +13,31 @@ type ServiceConfig struct {
...
+	Environment     envVars            `yaml:",omitempty"`
 	Production      map[string]string  `yaml:",omitempty"`
 }
+
+type envVars struct {
+	Default     map[string]string `yaml:",omitempty"`
+	Development map[string]string `yaml:",omitempty"`
+	Production  map[string]string `yaml:",omitempty"`
+	Secrets     []string          `yaml:",omitempty"`
+}
```
charlierudolph: Please define this struct in its own file. We are trying to have a single struct per file

src/types/service_config.go
```diff
@@ -13,5 +13,31 @@ type ServiceConfig struct {
...
+	case "production":
+		envVars = s.Environment.Production
+	case "development":
+		envVars = s.Environment.Development
+	}
+	for k, v := range envVars {
+		defaultVars[k] = v
+	}
+	return defaultVars, s.Environment.Secrets
+}
```
charlierudolph: Can we add unit tests for this method?

src/terraform/commands.go
```diff
@@ -17,11 +18,63 @@ func RunInit(deployConfig types.DeployConfig) error {
...
+	envVarsJSON, err := json.Marshal(terraformEnvVars)
+	if err != nil {
+		return "", err
+	}
+	envVarsEscaped, err := json.Marshal(string(envVarsJSON))
+	if err != nil {
+		return "", err
+	}
+	return string(envVarsEscaped), nil
+}
```
charlierudolph: Can we add unit test these new methods?

charlierudolph (commented):
very cool

src/types/service_config.go
hugoesthere: What if it's not being used outside of this file? This seems more like a sub-struct of `ServiceConfig`.

src/types/service_config.go
```diff
@@ -13,5 +13,27 @@ type ServiceConfig struct {
...
 }
+
+// GetEnvVars compiles a service's environment variables
+// It overwrites default variables with environemnt specific ones,
+// returning a map of public env vars and a list of private env var keys
+func (s ServiceConfig) GetEnvVars(environment string) (map[string]string, []string) {
+	defaultVars := map[string]string{}
+	for k, v := range s.Environment.Default {
+		defaultVars[k] = v
+	}
```
charlierudolph: Can we create a merge function which takes in two `map[string]string` and merges the second into the first? I think that will help simplify this and even allow us to remove the envVars variable.

src/types/service_config.go
```diff
@@ -13,5 +13,27 @@ type ServiceConfig struct {
...
 	Dependencies    []DependencyConfig `yaml:",omitempty"`
+	Environment     EnvVars            `yaml:",omitempty"`
 	Production      map[string]string  `yaml:",omitempty"`
 }
+
+// GetEnvVars compiles a service's environment variables
+// It overwrites default variables with environemnt specific ones,
+// returning a map of public env vars and a list of private env var keys
+func (s ServiceConfig) GetEnvVars(environment string) (map[string]string, []string) {
+	defaultVars := map[string]string{}
```
charlierudolph: Can we name this result?

src/util/map.go
```diff
@@ -10,3 +10,11 @@ func JoinStringMaps(maps ...map[string]string) map[string]string {
 	}
 	return result
 }
+
+// Merge merges two maps. It overwrites existing keys of map1 if conflicting keys exist
+func Merge(map1 map[string]string, map2 map[string]string) map[string]string {
+	for k, v := range map2 {
+		map1[k] = v
+	}
+	return map1
```
charlierudolph: I don't think we should return map1. maps are like pointers so map1 gets updated automatically.

src/terraform/commands.go
```diff
@@ -17,11 +18,72 @@ func RunInit(deployConfig types.DeployConfig) error {
...
+	command = append(command, vars...)
+	err = util.RunAndLog(deployConfig.TerraformDir, []string{}, deployConfig.LogChannel, command...)
 	if err != nil {
 		return errors.Wrap(err, "'terraform plan' failed")
 	}
 	return err
 }
+
+// CompileVarFlags compiles the variable flags passed into a Terraform command
+func CompileVarFlags(deployConfig types.DeployConfig, secrets types.Secrets) ([]string, error) {
```
charlierudolph: Thoughts on moving this and its helper functions into its own file. Then can also put the tests for this in its own file.

charlierudolph (approved):

PR #478, opened 2017-08-23
[exo-deploy 3] Round off exo deploy

hugoesthere:
Run `terraform apply` and make final changes to Terraform files to make `exo deploy` work.


src/application/deployer.go
```diff
@@ -53,5 +54,15 @@ func StartDeploy(deployConfig types.DeployConfig) error {
 	}
 
 	deployConfig.LogChannel <- "Planning deployment..."
-	return terraform.RunPlan(deployConfig, secrets)
+	terraform.RunPlan(deployConfig, secrets)
+	if err != nil {
+		return err
+	}
+
+	if ok := prompt.Confirm("Do you want to continue?"); ok {
```
charlierudolph: Can this prompt me more descriptive? This is confirming you want toi actually deploy correct? Also why is this prompt here versus at the beginning?

src/terraform/commands.go
```diff
@@ -33,6 +33,21 @@ func RunPlan(deployConfig types.DeployConfig, secrets types.Secrets) error {
 	return err
 }
 
+// RunApply runs the 'terraform apply' command and passes variables in as command flags
+func RunApply(deployConfig types.DeployConfig, secrets types.Secrets) error {
+	command := []string{"terraform", "apply"}
```
charlierudolph: Thoughts on removing this variable? Could do the following
```go
err = util.RunAndLog(deployConfig.TerraformDir, []string{}, deployConfig.LogChannel, "terraform", "apply", vars...)

File: src/terraform/commands.go [hunk 3]
```diff
@@ -33,6 +33,21 @@ func RunPlan(deployConfig types.DeployConfig, secrets types.Secrets) error {
...
+	vars, err := CompileVarFlags(deployConfig, secrets)
+	if err != nil {
+		return err
+	}
+	command = append(command, vars...)
+	err = util.RunAndLog(deployConfig.TerraformDir, []string{}, deployConfig.LogChannel, command...)
+	if err != nil {
+		return errors.Wrap(err, "'terraform apply' failed")
```
charlierudolph: I don't think this wrap is needed as its wrapping it with arguments passed into `util.RunAndLog`. Ideally we only wrap it with context not passed to the function and rely on the function to report good errors based on its inputs. 

`util.RunAndLog` already wraps run errors with the command it was running and the output

src/application/deployer.go
hugoesthere: Yes it's prompting to deploy, it's asked after `terraform plan` and before `terraform apply`

charlierudolph (approved):

PR #485, opened 2017-08-24
Update terraform hash

hugoesthere:


src/terraform/index.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // terraformModulesCommitHash is the git commit hash that reflects
 // which of the Terraform modules in Originate/exosphere we are using
-const terraformModulesCommitHash = "8786f912"
+const terraformModulesCommitHash = "48731e6"
```
charlierudolph: Hmm. Should this always be the latest commit (and thus we should try and automate this type of update) or one we want to update manually?

hugoesthere: It doesn't have to be the latest commit, but it should be the latest commit that changed anything in `terraform/modules`. What would that automatic update look like?

charlierudolph: We can write two small scripts. One that verifies this value if the latest commit that changed anything in terraform modules. Another that updates it to the latest. We would probably use it in the release process.

Alternatively do we need to have a commit hash? Is there another way we could get by without it?

hugoesthere: We'll have to be updating this more often than during release if we want a working version of this command.

I think we'll always have to specify commit hashes, because as we release new versions of Exosphere we want to lock which Terraform modules are being used, instead of always pointing to the latest ones.

bin/pre-release
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+hash=$(git log --format="%h" -n 1 src/terraform/modules/)
+
+if [ -z "$(grep "$hash" src/terraform/index.go)" ]; then
+  echo "Terraform module hash needs to be updated to $hash"
+  exit 1
+fi
+
+
```
charlierudolph, line 10: Please remove these extra newlines

charlierudolph (approved):

PR #488, opened 2017-08-25
fix template tests

charlierudolph:
I merged #487 before CI finished running, where the tests failed. Fixing it here


charlierudolph:
Merged to get master back to passing

PR #494, opened 2017-08-29
Hh networks

hugoesthere:
resolves #

### DIFF
--- a/features/run/run.feature
+++ b/features/run/run.feature
@@ -13,18 +13,24 @@ Feature: running Exosphere applications
     When starting "exo run" in my application directory
     Then it prints "setup complete" in the terminal
     And my machine has acquired the Docker images:
-      | tmp_users        |
-      | tmp_web          |
-      | originate/exocom |
+      | exampleapplication_users        |
+      | exampleapplication_web          |
+      | originate/exocom                |
     And the docker images have the following folders:
-      | IMAGE       | FOLDER       |
-      | tmp_users   | node_modules |
-      | tmp_web     | node_modules |
+      | IMAGE                      | FOLDER       |
+      | exampleapplication_users   | node_modules |
+      | exampleapplication_web     | node_modules |
     And it prints "all dependencies online" in the terminal
     And it prints "all services online" in the terminal
     And my machine is running the services:
       | NAME  |
       | web   |
+    # And the network "exampleapplication_default" contains the running services:
+    #   | NAME             |
+    #   | web              |
+    #   | users            |
+    #   | originate/exocom |
+
 
 
   Scenario: booting an Exosphere application with external docker images
@@ -38,3 +44,7 @@ Feature: running Exosphere applications
     And my machine is running the services:
       | NAME             |
       | external-service |
+    And the network "exampleapplication_default" contains the running services:
+      | NAME             |
+      | external-service |
+      | originate/exocom |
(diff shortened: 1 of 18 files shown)


features/run/run.feature
```diff
@@ -13,18 +13,24 @@ Feature: running Exosphere applications
...
     And it prints "all dependencies online" in the terminal
     And it prints "all services online" in the terminal
     And my machine is running the services:
       | NAME  |
       | web   |
+    # And the network "exampleapplication_default" contains the running services:
+    #   | NAME             |
+    #   | web              |
+    #   | users            |
+    #   | originate/exocom |
```
charlierudolph, line 32: Can this be removed or uncommented?

src/cmd/shared.go
```diff
@@ -64,3 +66,10 @@ func prettyPrintSecrets(secrets map[string]string) {
 	}
 	fmt.Printf("%s\n\n", string(secretsPretty))
 }
+
+// GetDockerComposeProjectName creates a docker compose project name the same way docker-compose mutates the COMPOSE_PROJECT_NAME env var
+func GetDockerComposeProjectName(appName string) string {
+	reg := regexp.MustCompile("[^a-zA-Z0-9]")
+	replacedStr := reg.ReplaceAllString(appName, "")
+	return strings.ToLower(replacedStr)
+}
```
charlierudolph, line 75: Thoughts on putting this in the `docker/compose` package and naming it `GetProjectName()`?

src/application/deployer.go
```diff
@@ -19,7 +19,13 @@ func StartDeploy(deployConfig types.DeployConfig) error {
...
 
 	deployConfig.LogChannel <- fmt.Sprintf("Building %s %s...\n\n", deployConfig.AppConfig.Name, deployConfig.AppConfig.Version)
-	initializer, err := NewInitializer(deployConfig.AppConfig, deployConfig.LogChannel, "exo-deploy", deployConfig.AppDir, deployConfig.HomeDir)
+	initializer, err := NewInitializer(
+		deployConfig.AppConfig,
+		deployConfig.LogChannel,
+		"exo-deploy",
+		deployConfig.AppDir,
+		deployConfig.HomeDir,
+		deployConfig.DockerComposeProjectName)
```
charlierudolph, line 28: Thoughts on adding a `GetDockerComposeProjectName` to the `AppConfig` struct so we don't need this extra parameter to initializer / runner / deployer?

hugoesthere:
Sorry this is unfinished, I don't know how this was opened.

src/application/deployer.go
hugoesthere: I liked created the project name at the highest level because that ensures we're creating it in one place and passing the same name down, as opposed to each of `initializer` / `runnner` creating their own for example. If we run `docker-compose -p app1 initializer build` then `docker-compose -p app2 initializer up` then docker compose will think we're running two different projects and it won't behave as expected. I already ran into some problems with this writing tests.

Also, `deployer` and `tester` both use `initializer`, but they require different project names so that needs to be passed down instead of generated inside initializer.

hugoesthere: If we wanted to clean it up a little bit we could create the project name in `application/deployer` and `application/tester` since `application/initializer` is the only one being reused, but for the runner we'd still have to create it in `cmd/run`, b/c that's where both `initializer` and `runner` are created.

PR #501, opened 2017-08-30
Add worker service [part 1]

### LINKED ISSUE #500: Add concept of "private" service
In production we need public, private and worker services. These services are defined as follows:

- Public: has an external facing ALB that lives in the public subnet, plus an internal ALB in the private subnet
- Private: has an internal facing ALB in the private subnet
- Worker: has no load balancers

All services themselves live in the private subnet. Currently we have public and worker services (though we call them private). This ticket is to add the private service and rename what we currently call private to worker. There is no affect on the dev environment.

hugoesthere:
Partially resolves https://github.com/Originate/exosphere/issues/500

Adds worker service to dev environment, working on PR that actually makes this meaningful in prod


src/types/app_config.go
```diff
@@ -46,6 +46,9 @@ func (a AppConfig) GetDependencyNames() []string {
 // GetServiceData returns the configurations data for the given services
 func (a AppConfig) GetServiceData() map[string]ServiceData {
 	result := make(map[string]ServiceData)
+	for serviceName, data := range a.Services.Worker {
```
charlierudolph: Gahh, I hate that we have three for loops so much in this file. Thoughts on creating a helper function that does the following
```go
func (a AppConfig) forEachService(fn func(string, string, ServiceData)) {
  for serviceName, data := range a.Services.Worker {
    fn("worker", serviceName, data) 
  }
  // similar for private and public
}
```
That would make this method be
```go
func (a AppConfig) GetServiceData() map[string]ServiceData {
  result := make(map[string]ServiceData)
  a.forEachService(func(type, name string, data ServiceData) {
    result[serviceName] = data 
  })
  return result
}
```
Likewise it would simplify quite a few methods in here

charlierudolph (approved):

PR #502, opened 2017-08-30
Add worker service [part 2]

### LINKED ISSUE #500: Add concept of "private" service

hugoesthere:
Partially resolves https://github.com/Originate/exosphere/issues/500

Renames private service to worker service and then actually adds the private service Terraform.


src/terraform/index.go
```diff
@@ -72,6 +72,8 @@ func generateServiceModules(serviceConfigs map[string]types.ServiceConfig, servi
 			module, err = generateServiceModule(serviceName, serviceConfig, imagesMap, "public_service.tf")
 		case "private":
 			module, err = generateServiceModule(serviceName, serviceConfig, imagesMap, "private_service.tf")
+		case "worker":
+			module, err = generateServiceModule(serviceName, serviceConfig, imagesMap, "worker_service.tf")
```
charlierudolph: I think you can remove the select by using the protectionLevel to generate the filename

charlierudolph (approved):

PR #503, opened 2017-08-30
exo run: update environment variables

charlierudolph:
* make the exocom port hardcoded. There is no reason I can think of for anyone to override it since we are using docker images.
* put env variables into the docker compose file instead of using env variables passed to docker run


hugoesthere (approved):
This is much cleaner. We might even consider hardcoding a default port 80 in Exocom and in exoservice/relay if `EXOCOM_PORT` isn't defined. We haven't actually been needing to configure that value anyways.

charlierudolph:
The custom configuration only made sense to me if running locally but when running with docker images, there is no reason I see that you would override it

PR #509, opened 2017-08-31
exo-clean cleans containers

### LINKED ISSUE #320: "exo clean" should remove existing Docker containers for the app
> When there are running Docker containers that break "exo run"
I want to be able to stop and remove them cleanly using "exo clean"
So that I can start my application normally without having to google how to remove Docker containers manually.

#### Rules:
- running "exo clean" stops and removes all Docker containers that are listed in application.yml 
- running Docker containers are stopped by running `docker stop <container names>`
- stopped Docker containers are removed by running `docker rm <container names>`

#### Notes:
This fixes this error:
```
exo-run  Creating web ... error
               exo-run  

               exo-run  ERROR: for web  Cannot create container for service web: Conflict. The container name "/web" is already in use by container    "f266b77ba98a3212d1ad727c1659195136c6c4968ad92c0961d4afe20087b06a". 
                               You have to remove (or rename) that container to be able to reuse that name.
```

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/320


test_helpers/clean_feature_context.go
```diff
@@ -49,6 +90,74 @@ func CleanFeatureContext(s *godog.Suite) {
...
+		}
+		testContainerProcess = execplus.NewCmdPlus("docker-compose", "-p", testNetwork, "create")
+		testContainerProcess.SetDir(path.Join(appDir, "service", "tests", "tmp"))
+		return testContainerProcess.Start()
+	})
+
+	s.Step(`^my machine has running third party containers$`, func() error {
+		thirdPartyContainerProcess = execplus.NewCmdPlus("docker", "run", "--name", thirdPartyContainer, "--rm", "alpine", "sleep", "20")
+		err := thirdPartyContainerProcess.Start()
+		time.Sleep(2500 * time.Millisecond)
```
charlierudolph: Whats the need for the sleep here?

test_helpers/clean_feature_context.go
```diff
@@ -49,6 +90,74 @@ func CleanFeatureContext(s *godog.Suite) {
...
+		return err
+	})
+
+	s.Step(`^it removes application and test containers$`, func() error {
+		hasNetworkBool, err := hasNetwork(dockerClient, appNetwork)
```
charlierudolph: Thoughts on renaming `hasNetworkBool` -> `hasAppNetwork`. Similar for `hasTestNetwork`

src/application/clean.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	serviceData := config.GetServiceData(appConfig.Services)
+	for _, data := range serviceData {
+		dockerComposeFile := path.Join(appDir, data.Location, "tests", "tmp", "docker-compose.yml")
+		exists, err := util.DoesFileExist(dockerComposeFile)
+		if err != nil {
+			return err
+		}
+		if exists {
+			composeProjectName := fmt.Sprintf("%stests", composebuilder.GetDockerComposeProjectName(appDir))
```
charlierudolph: This is duplicated in `src/cmd/test.go`. Can we extract this into a function somewhere. Could be something like `GetTestsDockerComposeProjectName`?

src/application/clean.go
```diff
@@ -0,0 +1,63 @@
...
+	if exists {
+		composeProjectName := composebuilder.GetDockerComposeProjectName(appDir)
+		_, err = compose.KillAllContainers(compose.BaseOptions{
+			DockerComposeDir: path.Join(appDir, "tmp"),
+			LogChannel:       logChannel,
+			Env:              []string{fmt.Sprintf("COMPOSE_PROJECT_NAME=%s", composeProjectName)},
+		})
+		if err != nil {
+			return err
+		}
```
charlierudolph: There appears to be a good amount of duplication here between does the docker compose file exist and killing all containers. Can we dry that up a bit? Maybe a function that takes in the `dockerComposeDir` and `composeProjectName`

src/cmd/clean.go
```diff
@@ -25,12 +29,26 @@ var cleanCmd = &cobra.Command{
 		if err != nil {
 			panic(err)
 		}
-		fmt.Println("removed all dangling images")
+		logger.Log("exo-clean", "removed all dangling images")
```
charlierudolph, line 33: This could be `logChannel <- "removed all dangling images"`

test_helpers/clean_feature_context.go
```diff
@@ -49,6 +90,74 @@ func CleanFeatureContext(s *godog.Suite) {
...
+		appContainerProcess = execplus.NewCmdPlus("docker-compose", "-p", appNetwork, "up")
+		appContainerProcess.SetDir(path.Join(appDir, "tmp"))
+		err := appContainerProcess.Start()
+		if err != nil {
+			return err
+		}
+		testContainerProcess = execplus.NewCmdPlus("docker-compose", "-p", testNetwork, "up")
+		testContainerProcess.SetDir(path.Join(appDir, "service", "tests", "tmp"))
+		err = testContainerProcess.Start()
+		time.Sleep(2500 * time.Millisecond)
```
charlierudolph: Why is this sleep needed?

test_helpers/clean_feature_context.go
```diff
@@ -49,6 +90,74 @@ func CleanFeatureContext(s *godog.Suite) {
...
+	s.Step(`^my machine has stopped application and test containers$`, func() error {
+		appContainerProcess = execplus.NewCmdPlus("docker-compose", "-p", appNetwork, "create")
+		appContainerProcess.SetDir(path.Join(appDir, "tmp"))
+		err := appContainerProcess.Start()
+		if err != nil {
+			return err
+		}
+		testContainerProcess = execplus.NewCmdPlus("docker-compose", "-p", testNetwork, "create")
+		testContainerProcess.SetDir(path.Join(appDir, "service", "tests", "tmp"))
+		return testContainerProcess.Start()
```
charlierudolph: This and the previous step seep quite similar with the only difference being "up" and "create". Can we dry that up?

test_helpers/clean_feature_context.go
hugoesthere: To give the containers time to "start up". I think there was a point in the test between when `docker-compose up` has been executed and the actual containers show up as running in `docker ps`

test_helpers/clean_feature_context.go
hugoesthere: Same reason as above.

test_helpers/clean_feature_context.go
charlierudolph: Can we wait for them to show up in `docker ps` then? I'd prefer we have deterministic waits instead of a time we hope is long enough

src/cmd/clean.go
hugoesthere: I think we avoided using logChannels here because that gave us a problem where the process would finish before all statements were printed.

charlierudolph: True, nevermind good as is

charlierudolph (approved):
Good to merge when green (or when confident nothing broke as the builds have been failing every now and then for me)

Issue #497, opened 2017-08-29
exo deploy: automate exocom env vars


charlierudolph:
When deploying an application with exocom, the env vars each service needs should be auto-populated so they don't need to be listed in the `service.yml`.

See https://github.com/Originate/space-tweet/pull/34/files#r135900724

charlierudolph:
@hugobho I'd like to get this fixed for EDX. So to me there are a couple updates that we need to make. Couple questions:

* Where is the exocom host name created? (or how to mirror what the name terraform gives it?)
* Can I unit test this? Since we had sme unit tests for addinng the service routes as an env variable, I am guessing yes but am unsure.

hugoesthere:
> Where is the exocom host name created?

That's fixed in the `.tf` modules, it's created across multiple `.tf` files but the only dynamic part of `exocom.app-name.local` is `app-name`. Everything else is hard-coded so we can easily mirror the name as long as we know the application name.
> Can I unit test this?

I think we can. Just test the function that generates environment variables, and if Exocom is a listed app dependency then check that those fields are there.

PR #515, opened 2017-09-06
exo deploy: automate env vars - service role

charlierudolph:
part of #497

### DIFF
--- a/src/terraform/command_helpers.go
+++ b/src/terraform/command_helpers.go
@@ -34,6 +34,7 @@ func compileEnvVars(deployConfig types.DeployConfig, secrets types.Secrets) ([]s
 		for _, secretKey := range serviceSecrets {
 			serviceEnvVars[secretKey] = secrets[secretKey]
 		}
+		serviceEnvVars["ROLE"] = serviceName
 		serviceEnvVarsStr, err := createEnvVarString(serviceEnvVars)
 		if err != nil {
 			return []string{}, err
--- a/src/terraform/command_helpers.go
+++ b/src/terraform/command_helpers.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/Originate/exosphere/src/types"
+	"github.com/Originate/exosphere/src/util"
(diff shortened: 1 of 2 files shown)


src/terraform/command_helpers.go
```diff
@@ -34,6 +34,7 @@ func compileEnvVars(deployConfig types.DeployConfig, secrets types.Secrets) ([]s
 		for _, secretKey := range serviceSecrets {
 			serviceEnvVars[secretKey] = secrets[secretKey]
 		}
+		serviceEnvVars["ROLE"] = serviceName
```
hugoesthere: `ROLE` is an ExoCom specific env var, it probably shouldn't be passed to every service.

hugoesthere: We'll have to do something similar where `ROLE` is set only when `appDependency` is `exocomDependency`

charlierudolph: `ROLE` is passed to everything in the local environment: https://github.com/Originate/exosphere/blob/v0.23.0.alpha.4/src/docker/composebuilder/index.go#L40

Its also not exocom specific to me. Its the name given the service in the `application.yml` and would be there whether or not you are using exocom. The role allows you to give labels to a service that may be a configurable reusable service.

hugoesthere: Hm that must've slipped through a PR.

Yes it's true that we can use `ROLE` for anything, but the reason we set that variable in the first place is so exoservice/relay has a name to give Exocom when it registers. It's technically not needed if you're not using Exocom. I'm fine with leaving it as is, but that just means we'll always be defining the `ROLE` env var for the user and that'll be a variable name they can't use.

charlierudolph: I think its a good generic one for now. We can certainly make it exocom specific if ever is annoying to someone. I'll update this so similar to in the local version where it is assigned first so it could be overridden.

charlierudolph:
@hugobho comments addressed

hugoesthere (approved):

PR #517, opened 2017-09-06
update exocom

charlierudolph:
released some new versions of the templates and now remove the need for the EXOCOM_PORT env variable


features/tutorial.feature
```diff
@@ -182,16 +182,20 @@ Feature: Following the tutorial
...
 
         index(req, res) {
-          this.send('todo.list', {}, (todos) => {
-            res.render('index', {todos})
+          this.send('todo.list', {}, (messageName, payload) => {
+            if (messageName === 'todo.listing') {
+              res.render('index', {todos: payload})
+            } else {
+              res.sendStatus(500)
+            }
```
hugoesthere, line 190: Were these changes meant to be included as well?

hugoesthere (approved):
Looks good, `exocom.tf` will need to be updated as well. This line can be removed: https://github.com/Originate/exosphere/blob/master/src/terraform/templates/exocom.tf#L30 since the default value when one isn't provided is `80`

features/tutorial.feature
charlierudolph: Yeah. They were a result of a change in exorelay

charlierudolph:
@hugobho want to take one last look?

src/terraform/modules/aws/custom/exocom/exocom-service/exocom-service.tf
```diff
@@ -2,7 +2,6 @@ module "task_definition" {
   source = "./exocom-task-definition"
 
   command               = "${var.command}"
-  container_port        = "${var.container_port}"
```
hugoesthere: Let's keep this variable in, that way we are in line with behavior on Exocom: container ports come online at port 80 by default but can be overridden if one is passed down. Removing this line will prevent us from passing one down at all. Just removing `container_port = "3100"` in `exocom.tf` and in the corresponding test is sufficient.

hugoesthere (approved):
Good to go after that comment is addressed.

Issue #520, opened 2017-09-07
Go AWS library: allow support for profiles


hugoesthere:
Seeing as a tech lead using `exo-deploy` will most likely have multiple AWS profiles on their machine, we should allow support for profile specification that way the library doesn't always default to the profile named `default`.

PR #534, opened 2017-09-08
Add AWS profile support

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/520

### DIFF
--- a/src/cmd/shared.go
+++ b/src/cmd/shared.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path"
 
 	"github.com/Originate/exosphere/src/aws"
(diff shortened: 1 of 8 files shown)


src/cmd/shared.go
```diff
@@ -35,13 +36,15 @@ func getAppConfig() (types.AppConfig, error) {
 	return appConfig, nil
 }
 
-func getAwsConfig() (types.AwsConfig, error) {
+func getAwsConfig(homeDir, profile string) (types.AwsConfig, error) {
```
charlierudolph: Can we get the home dir from within this function instead of requiring it to be passed in?

charlierudolph (approved):
~Do we need to do anything else~

PR #537, opened 2017-09-08
exo deploy: fix terraform error

charlierudolph:
Appears the else statement was being evaluated no matter what which was causing an error


hugoesthere (approved):

hugoesthere:
Make sure to get this in before shipping the release

Issue #528, opened 2017-09-07
exo deploy: validate production sections of app/service.yml 


charlierudolph:
For example setting the cpu and memory

kevgo:
FYI @alexdavid has the idea to extract this part into Terraform files.

charlierudolph:
+ whatever we add to the app yaml in other tickets

hugoesthere:
Closed in https://github.com/Originate/exosphere/commit/1f3f1eb5a6d4974ad44fae4b709a2e55f8a9ffaf

PR #548, opened 2017-09-12
Validate application and service production fields

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/528

### DIFF
--- a/src/types/app_config.go
+++ b/src/types/app_config.go
@@ -124,7 +124,7 @@ func (a AppConfig) ValidateProductionFields() error {
 	}
 	for _, field := range requiredFields {
 		if a.Production[field] == "" {
-			return fmt.Errorf("application.yml missing required field '%s'", field)
+			return fmt.Errorf("application.yml missing required field 'production.%s'", field)
 		}
 	}
 	return nil
--- a/src/types/app_config.go
+++ b/src/types/app_config.go
@@ -120,7 +120,7 @@ func (a AppConfig) forEachService(fn func(string, string, ServiceData)) {
 func (a AppConfig) ValidateProductionFields() error {
 	requiredFields := []string{"url", "region", "account-id", "ssl-certificate-arn"}
 	if a.Production == nil {
-		return errors.New("application.yml missing required field 'production'")
+		return errors.New("application.yml missing required section 'production'")
 	}
 	for _, field := range requiredFields {
 		if a.Production[field] == "" {
--- a/src/types/app_config.go
+++ b/src/types/app_config.go
@@ -114,3 +114,18 @@ func (a AppConfig) forEachService(fn func(string, string, ServiceData)) {
 		fn("public", serviceName, data)
 	}
 }
+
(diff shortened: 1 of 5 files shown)


src/types/app_config_test.go
```diff
@@ -11,6 +11,33 @@ import (
...
+		It("should throw an error when AppConfig is missing fields", func() {
+			appConfig = types.AppConfig{
+				Production: map[string]string{
+					"url":        "originate.com",
+					"account-id": "123",
+					"region":     "us-west-2",
+				},
+			}
+			err := appConfig.ValidateProductionFields()
+			Expect(err).To(HaveOccurred())
```
charlierudolph: Can we also test what the error message? And maybe add an test case for when there is no production section?

src/types/service_config_test.go
```diff
@@ -8,6 +8,40 @@ import (
...
+		workerConfig := types.ServiceConfig{
+			Production: map[string]string{
+				"cpu":    "128",
+				"memory": "128",
+			},
+		}
+
+		It("throws an error if public production fields are missing", func() {
+			err := publicConfig.ValidateProductionFields("public-service", "public")
+			Expect(err).To(HaveOccurred())
```
charlierudolph: Can we test the error message here as well?

hugoesthere:
@charlierudolph  comments addressed

src/types/app_config.go
```diff
@@ -114,3 +114,18 @@ func (a AppConfig) forEachService(fn func(string, string, ServiceData)) {
 		fn("public", serviceName, data)
 	}
 }
+
+// ValidateProductionFields validates that service.yml contiains a production field
+// and the required fields
+func (a AppConfig) ValidateProductionFields() error {
+	requiredFields := []string{"url", "region", "account-id", "ssl-certificate-arn"}
+	if a.Production == nil {
+		return errors.New("application.yml missing required field 'production'")
```
charlierudolph: Thoughts on changing "required field" -> "required section" or "required block"?

src/types/service_config_test.go
```diff
@@ -8,6 +8,53 @@ import (
...
+			Production: map[string]string{
+				"cpu":    "128",
+				"memory": "128",
+			},
+		}
+
+		It("throws an error if the production field is missing", func() {
+			err := missingConfig.ValidateProductionFields("missing-service", "public")
+			Expect(err).To(HaveOccurred())
+			expectedErrorString := "service.yml for 'missing-service' missing required field 'production'"
```
charlierudolph: Thoughts on updating this error message to be
```
missing-service/service.yml missing required field 'production'
```
Would need to pass in the service location instead of the serviceName but I think that would be easier for the user to figure out where they need to fix

src/types/app_config.go
charlierudolph: Same for the service config

charlierudolph (approved):

PR #551, opened 2017-09-12
Use dev dockerfile

alexdavid:
Initial implementation of dev dockerfile.

Next PR will remove the `startup` section from `service.yml` so the redundancy will be removed


example-apps/app-with-external-docker-images/local-service/service.yml
```diff
@@ -12,3 +12,7 @@ restart:
 messages:
   sends:
   receives:
+
+development:
+  scripts:
+    run: node server.js
```
kevgo, line 18: Didn't we say this line should be in the Dockerfile?

kevgo (commented):
Nice!

src/types/service_config.go
```diff
@@ -2,6 +2,10 @@ package types
 
 import "github.com/Originate/exosphere/src/util"
 
+type DevelopmentConfig struct {
+	Scripts map[string]string `yaml:",omitempty"`
+}
```
charlierudolph: Please put this in its own file. We have been doing one struct per file.

src/types/service_config.go
```diff
@@ -2,6 +2,10 @@ package types
 
 import "github.com/Originate/exosphere/src/util"
 
+type DevelopmentConfig struct {
+	Scripts map[string]string `yaml:",omitempty"`
```
charlierudolph: Thoughts on Scripts being a struct with Run and Test properties instead of a map of string to string?

charlierudolph (commented):
Appears you have a lint issue to fix (enforced documentation)

example-apps/complex-setup-app/todo-service/service.yml
```diff
@@ -5,7 +5,6 @@ author: test-author
 startup:
   command: echo "does not run"
```
hugoesthere: Does this command have any use now?

hugoesthere (commented):
Looks good. How does this fit in with the multi-stage Dockerfile you were working on?

example-apps/app-with-external-docker-images/local-service/service.yml
alexdavid: It's in the dockerfile for production, but since the dev dockerfile may do many things (test, lint, run specs, run, etc...) we must specify them here.

src/types/service_config.go
alexdavid: I figured the scripts would be user specifiable. Any script you want to run in the dev docker image should go here much how `npm`/`yarn` do it. I imagine there being cases where we want to run other scripts in the dev docker image and we shouldn't restrict the user to just those two use cases

example-apps/complex-setup-app/todo-service/service.yml
alexdavid: It's temporary for now, next PR will remove this

alexdavid:
The multi-stage docker file is separate from this since both `Dockerfile` and `Dockerfile.dev` are able to use it if necessary. However, it is probably only going to be useful in `Dockerfile` which is what we are going to use for production deployment.

@hugobho @charlierudolph what are your thoughts on calling it `Dockerfile.prod`?

hugoesthere:
That's fine with me, makes it more clear.

charlierudolph:
I like the naming `Dockerfile.prod`. Makes it more clear to me

src/types/service_config.go
charlierudolph: We don't support something like that at the moment though and I'm also skeptical what that would be used for. I'd like to see a use case before we consider it

alexdavid: It would be useful for things like reinstalling dependencies on nodejs / go and it would also be useful for people who want to hack on services written in languages they do not have compilers/runtimes installed

alexdavid: But I'm also ok making this a struct until we need it to be a map

alexdavid: Also it would be good to have the option of splitting up lint, unit-tests and cucumber tests into separate commands

alexdavid:
@charlierudolph All comments addressed. I will do the `Dockerfile.prod` and support production in the next PR

charlierudolph (approved):

PR #552, opened 2017-09-12
Modify goprompt to accept spaces between strings

### LINKED ISSUE #524: exo configure - cannot deal with secret values having spaces
When entering a value which contains a whitespace, the part after the whitespace is used as the next key

hugoesthere:
Meant to resolve https://github.com/Originate/exosphere/issues/524

#524 is an issue stemming from the package we're using, not our code itself. I'm not sure if it's okay updating the vendor files manually like this. We could always fork segmentio/go-prompt. Alternatively, there are people who have already made these changes:
https://github.com/jandre/go-prompt
https://github.com/kofalt/go-prompt


charlierudolph:
I think we need to update glide to install from the fork. Otherwise this could be overwritten the next time we update dependencies (which happens anytime we add/remove a dependency)

hugoesthere:
Ok, should we just use one of those forks then?

charlierudolph:
Yes please

hugoesthere:
@charlierudolph switched to forked repo

charlierudolph (approved):

PR #555, opened 2017-09-13
Fix terraform bug

hugoesthere:
`key_name` needs to be a defined variable whether or not exocom is used in a project. This change ensures that the `key_name` block is always generated in `main.tf`


charlierudolph (approved):
Appears a unit test needs to be updated. Ship when green

PR #557, opened 2017-09-13
Prod dockerfile

alexdavid:
Finishes the dockerfile refactor we talked about


src/terraform/index.go
```diff
@@ -77,13 +76,9 @@ func generateServiceModules(deployConfig types.DeployConfig, serviceProtectionLe
...
 
 func generateServiceModule(serviceName string, awsConfig types.AwsConfig, serviceConfig types.ServiceConfig, imagesMap map[string]string, filename string) (string, error) {
-	command, err := json.Marshal(strings.Split(serviceConfig.Startup["command"], " "))
-	if err != nil {
-		return "", errors.Wrap(err, "Failed to marshal service startup command")
-	}
 	varsMap := map[string]string{
 		"serviceRole":         serviceName,
-		"startupCommand":      string(command),
+		"startupCommand":      "[]",
```
charlierudolph: Could we remove this variable?

charlierudolph (commented):
What about the rename to `Dockerfile.prod`?

src/terraform/index.go
alexdavid: I removed it from the templated terraform file which should allow it to be defaulted in the terraform modules

hugoesthere (approved):
Good once tests are green

PR #562, opened 2017-09-15
update gometalinter

### LINKED ISSUE #527: exo deploy: catch terraform plan errors
When terraform plan reports errors (such as when a service name is less than 32 characters), the user is still being asked: "Do you want to apply this plan?""

charlierudolph:
* we were using an outdated gometalinter. Updated to the latest
* Make us aware of warnings. There was a warning that an unused variable was causing most linters to error. The only warning currently is caused by the fact that we some nolint statments. I want to remove the need for those and then we can validate there are no warnings.
* Remove gotype as not much use and it was causing a warning as it couldn't run because gometalinter was passing it the arguments wrong

resolves #527


src/aws/terraform_helpers.go
```diff
@@ -43,7 +44,7 @@ func createRemoteState(currSession *session.Session, config *aws.Config, bucketN
 }
 
 // creates dynamodb table to store terraform lock file
-func createLockTable(currSession *session.Session, config *aws.Config, tableName string) error {
+func createLockTable(currSession client.ConfigProvider, config *aws.Config, tableName string) error {
```
hugoesthere, line 47: What was the warning for this just out of curiosity?

hugoesthere (approved):

src/aws/terraform_helpers.go
charlierudolph: That `*session.Session` is more specific than the requirements of its use in this function.

PR #567, opened 2017-09-16
Replace restarting services with volume mouting

alexdavid:
This mounts the service directory in development to `/mnt` on the conatiner. This allows the choice up to the service if they want to make use of this or not. This is not a breaking change (other than the service watcher going away but it was broken on macos anyway). All existing services/templates will work fine, but they can now assume the service directory is mounted in `/mnt`


features/run/mounting-service-directories.feature
```diff
@@ -0,0 +1,22 @@
+Feature: Mounting service directories in docker
+
+  As an Exosphere developer
+  I want the services inside my application to mount as volumes in the docker containers
+  So that I can use existing tooling to restart services on change
+
+  Rules:
+  - "exo run" mounts service directories in in /mnt on the docker container
```
charlierudolph: typo with `in in`

features/run/mounting-service-directories.feature
```diff
@@ -0,0 +1,22 @@
...
+  Scenario: changes made in the service directory appear in docker
+    Given I am in the root directory of the "frontend-with-webpack" example application
+    And starting "exo run" in my application directory
+    And it prints "all services online" in the terminal
+    Then http://localhost:8080 displays:
+      """
+      Hello world
+      """
+    When modifying frontend-service/src/index.html to "Foobar"
+    Then http://localhost:8080 displays:
```
charlierudolph: Will this pass consistently or is there a chance timing issues could pop up here? If timing issues are possible can we wait for `exo run` to print something like `webpack recompiled` or whatever status message we would get?

charlierudolph (approved):
Beautiful. This is so cool

features/run/mounting-service-directories.feature
alexdavid: Yeah this is a good point. The problem is webpack doesn't announce anything in terminal when the root html file is changed. I originally had it modifying a js file, but this doesn't reflect changes when making an http request because those changes would be happening client side. Would you prefer to just verify the changes in the terminal, or should I modify both files?

charlierudolph: Another way to combat a timing issue is to modify the next step to retry loading the app if foobar doesn't appear (and keep retrying with a timeout of 5 seconds)

alexdavid: 🏁

features/run/mounting-service-directories.feature
alexdavid: 🏁

Issue #559, opened 2017-09-14
Long term solution for dbs and third party dependencies


hugoesthere:
Per our discussion here: https://github.com/Originate/exosphere/issues/508, we are going with a more permanent solution to third party dependencies in production. What do we think about this proposed format in `application.yml` taken from the discussion in #508:
```
development:
  dependencies:
    - name: exocom
      version: 0.24.0
    - name: mongo
      version: 3.4.0
       silent: true
       config:
          ...

production:
  dependencies
    - name: exocom
      version: 0.24.0
    - name: mongo
      version: 3.4.0
      config:
         ...
```

charlierudolph:
To be clear: `development` refers to using `exo run` and `production` refers to using `exo deploy`?

For the first version. I suggest no new features: no external in local, no overriding the version in dev, no `type` field

I would also prefer that dependencies only has `name/version` (as in this PR). 

For `development/dependencies`, I would prefer that we move the `silent` and `config` options there, but don't introduce a `type` as I don't think its needed to retain current functionality.

For `production/dependencies`. We would move the `external-in-production` flag there and probably rename to `external`.

I think that would be a minimal first start and we can build on it after that.

hugoesthere:
Ok, I've updated the comment above to reflect that. ~Renamed `development` -> `local`.~ Do we allow dependencies to be listed under `production` and not in `dependencies`?

charlierudolph:
In the example shouldn't `production` be `production/dependencies`?

Hmm. Thoughts on not having the top level dependencies? Since its an array, I think overriding is not very straightforward. We only save a few lines and could have the entire dependency config be in `local/dependencies` and `production/dependencies`

hugoesthere:
Yeah I agree, it's just kind of a source of confusion right now. Let's just get rid of it.

hugoesthere:
At which point couldn't we also just get rid of the `external` flag? That way you only list a dependency under `production/dependencies` if you need it, and you don't otherwise.

charlierudolph:
Perfect. Looks good to me. Do we have any current production config for dependencies?

hugoesthere:
Not currently, but I assume as we move onto supporting RDS probably

charlierudolph:
Okay cool. Just wanted to check that for production dependencies we don't currently need to add the config property.

PR #568, opened 2017-09-16
Dependencies refactor part 1

hugoesthere:
Resolving https://github.com/Originate/exosphere/issues/559


example-apps/simple/application.yml
```diff
@@ -3,9 +3,15 @@ description: For making testing fast
...
-    version: 0.26.1
+development:
+  dependencies:
+    - name: exocom
+      version: 0.26.1
+
+production:
+  dependencies:
+    - name: exocom
+      version: 0.26.1
```
kevgo, line 14: Rather than duplicating information here, can we add an application-wide `dependencies` block, and have the environments only provide overrides?

```yml
dependencies:
  - name: exocom
    version: 0.26.1

development:
  - name: exocom
    version: 0.27   # new exocom version in dev

production:
  # no overrides here, uses exocom 0.26.1
```

hugoesthere: We considered that but ultimately decided that it was only a source of confusion and didn't provide that much convenience. See: https://github.com/Originate/exosphere/issues/559#issuecomment-329852538

charlierudolph: I don't like the thought of merging arrays. Merging objects is much simpler and easier to understand, so if we really want to reduce duplication (and this is not much currently), I think we need to make the dependencies an object.

hugoesthere: It's more likely that a dependency's `config` will have to change between environments, in which case there will be one entry per dependency per environment containing at least `name` and `config`. At that point having a generic `dependencies` block might actually be more lines and doesn't make things any clearer.

hugoesthere:
@charlierudolph should be good for a real review now. Only test that's failing is the data race one.

charlierudolph (approved):

PR #571, opened 2017-09-18
fix `exo test` with mounted services

### LINKED ISSUE #570: using mnt with `exo test` is broken
The path to mounting the volume is incorrect since the directory of the docker compose file is `trade-service/tests/tmp` and the volume is `../trade_service:/mnt`. For reference, it appears we could get this working by using the value of the `build/context` as the where we want to mount

charlierudolph:
resolves #570


hugoesthere (approved):

charlierudolph:
@hugobho please review again. Switched the docker compose from using relative paths (which change in the tests) to absolute paths so they don't need to be updated in the tests

hugoesthere (approved):
👍

PR #580, opened 2017-09-19
Add production mode

### LINKED ISSUE #576: ability to run in production mode locally
* Add an option to `exo run` that will use the production docker images in order to test that they work as expected.

alexdavid:
Resolves #576


example-apps/static-asset-service/nginx/Dockerfile.dev
```diff
@@ -0,0 +1,5 @@
+FROM nginx
+
+COPY index.dev.html /usr/share/nginx/html/index.html
+
+CMD sh -c "nginx && sleep 1 && echo nginx online && tail -f /dev/null"
```
alexdavid, line 5: This was necessary since nginx doesn't output anything on boot up and we still rely on stdout to tell if a service is online. We will be able to remove this when we implement better ways to tell if services are online.

kevgo: Does nginx have a verbose mode?

alexdavid: It does not

kevgo: Well, Heilandsack!

charlierudolph (approved):
Looks great! Please merge with the latest. Would be great to have this in to test on EDX

charlierudolph:
@alexdavid @hugobho please take another look. This required some updates to work with the latest refactor

hugoesthere (approved):

PR #584, opened 2017-09-20
Rds integration

### LINKED ISSUE #559: Long term solution for dbs and third party dependencies

hugoesthere:
The end of a long journey to resolve: https://github.com/Originate/exosphere/issues/559


src/config/rds_production_dependency.go
```diff
@@ -0,0 +1,44 @@
...
+
+type rdsProductionDependency struct {
+	config    types.ProductionDependencyConfig
+	appConfig types.AppConfig
+}
+
+// GetDockerConfig returns docker configuration and an error if any
+func (r *rdsProductionDependency) GetDockerConfig() (types.DockerConfig, error) {
+	return types.DockerConfig{
+		Image: fmt.Sprintf("%s:%s", r.config.Name, r.config.Version),
```
hugoesthere: Wondering what we should do here as we still have the un-needed step where a DB dependency's docker image is uploaded to ECR

src/types/production_dependency_config_options.go
```diff
@@ -2,7 +2,12 @@ package types
 
 // ProductionDependencyConfigOptions represents the configuration of a development dependency
 type ProductionDependencyConfigOptions struct {
-	//TODO: populate with RDS fields
+	AllocatedStorage string `yaml:"allocated-storage,omitempty"`
+	InstanceClass    string `yaml:"instance-class,omitempty"`
+	DbName           string `yaml:"db-name,omitempty"`
+	Username         string `yaml:",omitempty"`
+	PasswordEnvVar   string `yaml:"password-env-var,omitempty"` //TODO: verify this is set via exo configure
+	StorageType      string `yaml:"storage-type,omitempty"`
```
charlierudolph: Weren't we going to group this in an RDS config?

src/config/rds_production_dependency.go
charlierudolph: Can we update the ProductionDependency interface to have something like `HasDockerConfig` or something? And the builder doesn't add those when they return false. 

Alternatively, we could make it so a return value from this function of nil means a similar thing. We may need to update the return type to be a pointer to allow nil to be returned

src/types/production_dependency_config_options.go
hugoesthere: Right, I'll add that in

src/config/rds_production_dependency.go
hugoesthere: What about generic dependencies? They would all either have to be pushed to ECR or not. Should we add yet another field to `ProductionDependencyConfig`?

charlierudolph: Aren't generic dependencies currently based on a docker image (which means we would always need to push)

hugoesthere: I'm not even sure what a generic dependency is at this point. I guess we'll just always push for now and then add a field if it becomes necessary in the future.

hugoesthere:

src/config/rds_production_dependency.go
charlierudolph: Can this now be updated to return an empty docker config?

charlierudolph (approved):

Issue #566, opened 2017-09-15
Compile service production dependencies


hugoesthere:
Currently we only generate terraform for application dependencies, we need to do this for dependencies listed under each service as well. This ticket should be completed after https://github.com/Originate/exosphere/issues/559 is done.

PR #585, opened 2017-09-20
Compile service production dependenceis

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/566


charlierudolph:
Does this work for the case where we have postgres as an application dependency and then list it in each service that uses it? I believe thats a requirement at the moment (but I could be wrong). I think anything listed as an application dependency should be available to all services (or the services its available to are defined in the app.yml).

hugoesthere:
Anything listed as an application dependency will have a terraform module generated for it, regardless of if it's also listed in service.yml. This PR is for the case where a dependency is listed in `service.yml` and not in `application.yml`.

charlierudolph (approved):

PR #591, opened 2017-09-22
add no-mount flag

charlierudolph:
Adds the no-mount flag to `exo run` and `exo test`


example-apps/fail-if-mounted/test-service/tests/tmp/docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: "3"
+services:
+  test-service:
+    build:
+      context: /Users/charlesrudolph/go/src/github.com/Originate/exosphere/example-apps/fail-if-mounted/test-service
+      dockerfile: Dockerfile.dev
+    command: /mnt/test
+    container_name: test-service
+    environment:
+      ROLE: test-service
```
hugoesthere: Were these `tmp/` dirs meant to be created? If not we'll probably have to checkout an app to run tests.

hugoesthere (approved):

PR #592, opened 2017-09-22
cleanup: remove tmp dirs

charlierudolph:
Also gitignore them to prevent us from committing any more in the future.


charlierudolph:
Closing as I believe this tmp dir is required

PR #594, opened 2017-09-24
release 0.26.1

charlierudolph:
* The no-mount flag is the only change
* Missed updating the terraform hash in 0.25.1 and 0.26.0

### DIFF
--- a/src/terraform/index.go
+++ b/src/terraform/index.go
@@ -11,7 +11,7 @@ import (
 
 // terraformModulesCommitHash is the git commit hash that reflects
 // which of the Terraform modules in Originate/exosphere we are using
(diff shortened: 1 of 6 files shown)


src/terraform/index.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // terraformModulesCommitHash is the git commit hash that reflects
 // which of the Terraform modules in Originate/exosphere we are using
-const terraformModulesCommitHash = "fa599050"
+const terraformModulesCommitHash = "16663974"
```
hugoesthere: My latest PR made this change, you can just do a `git sync`.

hugoesthere (approved):

PR #595, opened 2017-09-25
Rds: add endpoint to route53

hugoesthere:
Actually passes rds endpoint to a service that uses it.

### DIFF
--- a/example-apps/rds/application.yml
+++ b/example-apps/rds/application.yml
@@ -1,4 +1,4 @@
-name: Exosphere-application-with-rds-dependency
+name: exosphere-application-with-rds-dependency
 description: For testing generation of Terraform files
 author: exospheredev
 version: '1.0'
--- a/src/config/rds_production_dependency.go
+++ b/src/config/rds_production_dependency.go
@@ -45,6 +45,6 @@ func (r *rdsProductionDependency) GetDeploymentConfig() (map[string]string, erro
 // GetDeploymentServiceEnvVariables returns configuration needed for each service in deployment
 func (r *rdsProductionDependency) GetDeploymentServiceEnvVariables() map[string]string {
 	return map[string]string{
-		fmt.Sprintf("%s_ENDPOINT", strings.ToUpper(r.config.Name)): fmt.Sprintf("%s.%s.local", r.config.Name, r.appConfig.Name),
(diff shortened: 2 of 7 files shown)


src/config/rds_production_dependency.go
```diff
@@ -41,5 +44,7 @@ func (r *rdsProductionDependency) GetDeploymentConfig() (map[string]string, erro
 
 // GetDeploymentServiceEnvVariables returns configuration needed for each service in deployment
 func (r *rdsProductionDependency) GetDeploymentServiceEnvVariables() map[string]string {
-	return map[string]string{}
+	return map[string]string{
+		fmt.Sprintf("%s_ENDPOINT", strings.ToUpper(r.config.Name)): fmt.Sprintf("%s.%s.local", r.config.Config.Rds.DbName, r.appConfig.Name),
```
charlierudolph: So for postgres this would be something like `POSTGRES_ENDPOINT`? In dev this is currently `POSTGRES`, thoughts on removing the `_ENDPOINT` from the name? Mainly just think it would nice if its consistent between dev and prod

charlierudolph (approved):

src/config/rds_production_dependency.go
hugoesthere: Yep, that works.

src/config/rds_production_dependency.go
```diff
@@ -41,5 +44,7 @@ func (r *rdsProductionDependency) GetDeploymentConfig() (map[string]string, erro
 
 // GetDeploymentServiceEnvVariables returns configuration needed for each service in deployment
 func (r *rdsProductionDependency) GetDeploymentServiceEnvVariables() map[string]string {
-	return map[string]string{}
+	return map[string]string{
+		strings.ToUpper(r.config.Name): fmt.Sprintf("%s.%s.local", r.config.Config.Rds.DbName, r.appConfig.Name),
+	}
```
charlierudolph, line 49: Can we add a test for this too?

example-apps/rds/application.yml
```diff
@@ -1,4 +1,4 @@
-name: Exosphere-application-with-rds-dependency
+name: exosphere-application-with-rds-dependency
```
hugoesthere, line 1: This is the second place where we need the app name to be DNS compliant. I know we throw and error somewhere when setting up an AWS account, but should we just put this mandate somewhere in the documentation or in maybe the CLI help messages @charlierudolph ?

charlierudolph: We could update the error message that enforces this and add this it needs to dns compliant

PR #596, opened 2017-09-25
Add profile field to backend block

hugoesthere:
Discovered this just now - profile needs to be passed into the backend block so `terraform init` can use the proper credentials.


charlierudolph (approved):
merge on green

PR #601, opened 2017-09-26
Add name to rds module

hugoesthere:
Terraform throws and error when there are modules with the same name. This can happen when we have more than one type of rds dependency.


charlierudolph (approved):
merge on green

PR #602, opened 2017-09-26
Validate fields for db dependency

hugoesthere:


src/types/production_dependency_config_test.go
```diff
@@ -0,0 +1,46 @@
...
+			Name:    "postgres",
+			Version: "0.0.1",
+			Config: types.ProductionDependencyConfigOptions{
+				Rds: types.RdsConfig{
+					DbName:         "test",
+					Username:       "test-user",
+					PasswordEnvVar: "TEST_PASSWORD",
+				},
+			},
+		}
```
charlierudolph: Can we move the creation of these config objects to within the it block that uses them?

charlierudolph (approved):

PR #603, opened 2017-09-26
Define terraform var for rds password

hugoesthere:
Also update terraform module hash


charlierudolph (approved):
In the future please split this into two PRs (or leave updating the module to a new release).

Merge on green

hugoesthere:
Looks like the `pre-release` script picked up a commit hash from a feature branch instead of master. Will be separating into two PRs anyways then.

PR #607, opened 2017-09-26
Rds modules: some more cleanup

hugoesthere:
Should be the last of the cleanup needed

1) Make all rds fields required
2) Add `final_snapshot_identifier` field and along with it db name validation.
3) Missing `ttl` field in rds route 53 record
4) Create proper security groups


hugoesthere:
Actually, need to manage some security groups first.

hugoesthere:
@charlierudolph should be good for a review now.

charlierudolph (approved):

PR #608, opened 2017-09-27
Refactor exo test

alexdavid:
First step in my general refactor to help clean up the tight coupling in exosphere. To keep PRs small this initial refactor moves most of the mess into a single file (service_tester), which will be removed in my next PR.

It also introduces the idea of an `AppContext` and a `ServiceContext` which is a way of telling helper methods and docker compose about an app or a service.


src/application/service_tester.go
```diff
@@ -16,12 +17,42 @@ type ServiceTester struct {
...
 
 // NewServiceTester is ServiceTester's constructor
-func NewServiceTester(role string, serviceConfig types.ServiceConfig, builtDependencies map[string]config.AppDevelopmentDependency, appDir, serviceDir string, initializer *Initializer, runner *Runner) (*ServiceTester, error) {
+func NewServiceTester(serviceContext types.ServiceContext, logger *util.Logger, mode composebuilder.BuildMode) (*ServiceTester, error) {
+	appConfig := serviceContext.AppContext.Config
+	serviceConfig := serviceContext.Config
+	serviceDir := serviceContext.Location
+	role := serviceContext.Name
+	appDir := serviceContext.AppContext.Location
+	dockerComposeProjectName := fmt.Sprintf("%stests", composebuilder.GetDockerComposeProjectName(appDir))
```
hugoesthere, line 31: Can we still get this from a shared function like `getTestDockerComposeProjectName`? I think `exo clean` may need it.

src/types/context.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Context represents contextual information about what application/service the user is running
+type Context struct {
+	AppContext        AppContext
+	ServiceContext    ServiceContext
+	HasServiceContext bool
+}
+
+// GetContext returns a Context for the current working direcotry
+func GetContext() (Context, error) {
```
hugoesthere, line 13: 🏅

src/types/service_context.go
```diff
@@ -0,0 +1,36 @@
...
+	yaml "gopkg.in/yaml.v2"
+)
+
+// ServiceContext represents the exosphere service the user is running
+type ServiceContext struct {
+	Name       string
+	Location   string
+	Config     ServiceConfig
+	AppContext AppContext
+
```
hugoesthere: extra line

hugoesthere (approved):
Looks good to me. I like populating all the context fields in `application/...` more than when we used to do it in `cmd/...`. Hope this plays well with the rest of the commands

src/application/service_tester.go
alexdavid: Yeah I haven't cleaned up the service tester yet. Ideally I don't want anything in application to know about docker compose stuff. The docker package should actually take care of this IMO

Issue #617, opened 2017-09-29
rds app dependency: auto inject variables


charlierudolph:
database name, username, password to service

Should not need to be defined in all services. For config may need to determine what those vars get injected as

hugoesthere:
The new format as we discussed:
```
name: postgres
version: 9.6.4
config:
  rds:
    allocated-storage: 10
    db-name: edx
    username: edxdevuser
    password-secret-name: RDS_MASTER_PASSWORD
    instance-class: db.t2.micro
    service-env-var-names:
      db-name: DATABASE_NAME
      username: DATABASE_USERNAME
      password: DATABASE_PASSWORD
    storage-type: gp2
```

PR #622, opened 2017-10-02
Rds env var injection

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/617

### DIFF
--- a/src/types/rds_config.go
+++ b/src/types/rds_config.go
@@ -10,12 +10,20 @@ import (
 
 // RdsConfig holds configuration fields for an rds dependency
 type RdsConfig struct {
-	AllocatedStorage string `yaml:"allocated-storage,omitempty"`
-	InstanceClass    string `yaml:"instance-class,omitempty"`
-	DbName           string `yaml:"db-name,omitempty"`
-	Username         string `yaml:",omitempty"`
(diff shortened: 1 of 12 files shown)


src/types/rds_config.go
```diff
@@ -24,24 +32,38 @@ func (d *RdsConfig) IsEmpty() bool {
...
+		return fmt.Errorf("missing required field 'Rds.%s'", missingField)
+	}
+	dbNameRegex := regexp.MustCompile("^[a-zA-Z0-9-]+$")
+	if !dbNameRegex.MatchString(d.DbName) {
+		return errors.New("only alphanumeric characters and hyphens allowed in rds.db-name")
+	}
+	requiredServiceEnvVars := []string{"DbName", "Username", "Password"}
+	missingField = checkRequiredFields(d.ServiceEnvVarNames, requiredServiceEnvVars)
+	if missingField != "" {
+		return fmt.Errorf("missing required field 'Rds.ServiceEnvVarNames.%s", missingField) //TODO address camelcase issue
```
hugoesthere: Wanted to bring up that our error messages are in the form `Rds.ServiceEnvVarNames.DbName` when in a yaml folder they should be `rds.service-env-var-names.db-name`. Not sure if these are potentially misleading to the user.

charlierudolph: We should update them to the how they appear in the yaml file.

charlierudolph (approved):

PR #638, opened 2017-10-03
Remove terraform ref from tests

alexdavid:
We shouldn't need to update test files anytime we update the terraform hash.

Also I think we eventually might want to make this user configurable or at least generate a `terraform/ref.lock` to make sure we don't go swapping out terraform modules out from under the user for applications that need extremely stable deployments.

For right now I think it's good that we can update since we're still iterating quickly, but when we land on a more stable API we should semantically version this or something

### DIFF
--- a/bin/pre-release
+++ b/bin/pre-release
@@ -2,7 +2,7 @@
 
 hash=$(git log --format="%h" -n 1 src/terraform/modules/)
 
-if [ -z "$(grep "$hash" src/terraform/index.go)" ]; then
+if [ -z "$(grep "$hash" src/cmd/deploy.go)" ]; then
   echo "Terraform module hash needs to be updated to $hash"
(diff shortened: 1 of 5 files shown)


hugoesthere (approved):

charlierudolph:
Nice! Please update `bin/pre-release` as that is used to ensure we keep the hash up to date.

hugoesthere:
@charlierudolph Isn't that script still useful? Maybe when we start making this user configurable we should remove it but we still have to update that hash even if it's not in the tests.

charlierudolph:
I was trying to say update the script so it works with these changes as the file where the hash exists changed.

alexdavid:
Updated

charlierudolph (approved):
Thanks!

Issue #626, opened 2017-10-02
exo deploy: service version numbers


charlierudolph:
Make the service version numbers be the last SHA that changed them (`git rev-list -1 HEAD <service-dir>`) instead of the version listed in the `application.yml`

PR #646, opened 2017-10-05
Use commit sha as version numbers

hugoesthere:
Resolves: https://github.com/Originate/exosphere/issues/626

### DIFF
--- a/src/aws/ecr_helpers.go
+++ b/src/aws/ecr_helpers.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


src/aws/ecr_helpers.go
```diff
@@ -138,14 +143,25 @@ func buildImageName(dockerConfig types.DockerConfig, dockerComposeProjectName, s
...
+
+func getCommitSHA(deployConfig types.DeployConfig, serviceLocation string) (string, error) {
+	cmd := exec.Command("git", "rev-list", "-1", "HEAD", serviceLocation)
+	cmd.Dir = deployConfig.AppDir
+	output, err := cmd.Output()
+	if err != nil {
+		return "", err
 	}
-	return repositoryName, version
+	return strings.Trim(string(output), "\n"), err
```
charlierudolph, line 166: You can use `strings.TrimSpace` here. Also please use `return ..., nil` to make it more clear we are not returning an error here

charlierudolph (approved):

PR #647, opened 2017-10-05
Fix os stat bug

hugoesthere:
This was throwing an error when `main.tf` did not exist, when we really just want to return an empty byte array


charlierudolph:
We have a helper you can use: `util.DoesFileExist`

hugoesthere:
@charlierudolph switched to util function

charlierudolph (approved):

Issue #628, opened 2017-10-03
Use more unique s3 bucket names


hugoesthere:
Since s3 buckets are global if we simply name them `#{app-name}-terraform` and `#{app-name}-secrets` then we can't deploy to multiple accounts. We should figure out a more unique way of naming them.

hugoesthere:
@charlierudolph @alexdavid any suggestions for what we should name them? Some combination of app name and account id?

charlierudolph:
Thoughts on also including the region as I thought we could do multiple environments by deploying to another region?

```
#{account-id}-#{region}-#{app-name}-terraform
```

hugoesthere:
I think Chris had shot down the idea of deploying to multiple envs by deploying to another region

charlierudolph:
Auhh okay. Then `#{account-id}-#{app-name}-terraform` seems fine to me

PR #654, opened 2017-10-06
S3 Bucket cleanup

hugoesthere:
Resolves: https://github.com/Originate/exosphere/issues/628

Also removes unnecessary `dev` key in front of file key. I think that will become confusing as we start deploying to multiple environments.

Something to note: if we update existing projects (EDX) to use this, we might have to tear down their existing infrastructure before re-deploying, since we're putting`terraform.tfstate` in a whole new location. Alternatively we can try copying `edx-terraform/dev/terraform.tfstate` to `#{account-id}-edx-terraform/terraform.tfstate`. Same goes for application secrets.


charlierudolph (approved):

charlierudolph:
@alexdavid since you are deployed using exosphere, just want to let you know this is now in master and will be in the next release.

PR #659, opened 2017-10-10
exo deploy: load aws credentials from env variables

charlierudolph:
Needed for CI. The credentials file defaults to `~/.aws/credentials` 

Copied some of the code for the default credentials chain: https://github.com/aws/aws-sdk-go/blob/4a5fb7e221f1be0cb59b4cd4a50b56669b1d099b/aws/defaults/defaults.go#L90

### DIFF
--- a/src/aws/index.go
+++ b/src/aws/index.go
@@ -145,7 +145,12 @@ func getEcrAuth(ecrClient *ecr.ECR) (string, string, error) {
 
 func createAwsConfig(awsConfig types.AwsConfig) *aws.Config {
 	return &aws.Config{
-		Region:      aws.String(awsConfig.Region),
-		Credentials: credentials.NewSharedCredentials(awsConfig.CredentialsFile, awsConfig.Profile),
+		Region: aws.String(awsConfig.Region),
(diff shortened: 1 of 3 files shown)


src/aws/index.go
```diff
@@ -145,7 +145,12 @@ func getEcrAuth(ecrClient *ecr.ECR) (string, string, error) {
...
 	return &aws.Config{
-		Region:      aws.String(awsConfig.Region),
-		Credentials: credentials.NewSharedCredentials(awsConfig.CredentialsFile, awsConfig.Profile),
+		Region: aws.String(awsConfig.Region),
+		Credentials: credentials.NewCredentials(&credentials.ChainProvider{
+			Providers: []credentials.Provider{
+				&credentials.EnvProvider{},
+				&credentials.SharedCredentialsProvider{Profile: awsConfig.Profile},
+			},
+		}),
```
hugoesthere, line 154: I'm assuming the expected behavior is to first check if env vars are set if not default to `~/.aws/config`?

hugoesthere (approved):

src/aws/index.go
charlierudolph: Yep

PR #660, opened 2017-10-10
exo test: remove dependency on service role

### LINKED ISSUE #650: exo test: dependecy on service role
exo test depends on the fact that the service name / role (key in app yaml) matches the service directory. Remove this requirement as some services may be used multiple types in an applicaiton.

`exo test` should iterate through the locations (not the keys)

charlierudolph:
resolves #650


src/application/tester.go
```diff
@@ -17,7 +17,12 @@ func TestApp(appContext types.AppContext, logger *util.Logger, mode composebuild
 	}
 
 	numFailed := 0
+	locations := []string{}
 	for serviceName, serviceContext := range serviceContexts {
+		if util.DoesStringArrayContain(locations, serviceContext.Location) {
+			continue
+		}
+		locations = append(locations, serviceContext.Location)
```
charlierudolph: This is preventing from running the same tests twice when a service is reused. Wasn't exactly certain how to test this

src/application/service_tester.go
```diff
@@ -88,11 +88,13 @@ func (s *ServiceTester) getDockerComposeConfig() (types.DockerCompose, error) {
 	if err != nil {
 		return result, err
 	}
-	dockerConfigs := types.DockerConfigs{}
-	serviceDockerConfig := appDockerConfigs[s.Role]
-	serviceDockerConfig.DependsOn = s.getDependencyContainerNames()
-	serviceDockerConfig.Command = s.ServiceConfig.Development.Scripts["test"]
-	dockerConfigs[s.Role] = serviceDockerConfig
+	dockerConfigs, err := s.Initializer.getServiceDockerConfig(s.Role, s.ServiceConfig, types.ServiceData{Location: s.Role})
```
hugoesthere: Why would `Location` be `s.Role` here?

src/application/service_tester.go
```diff
@@ -88,11 +88,13 @@ func (s *ServiceTester) getDockerComposeConfig() (types.DockerCompose, error) {
...
+	if err != nil {
+		return result, err
+	}
+	serviceConfig := dockerConfigs[s.Role]
+	serviceConfig.DependsOn = s.getDependencyContainerNames()
+	serviceConfig.Command = s.ServiceConfig.Development.Scripts["test"]
```
hugoesthere: It doesn't look like these `serviceConfig` objects actually being used anywhere, beyond being assigned values here.

src/application/tester.go
hugoesthere: This seems simple and sound enough. If you want, you could extract a `forServiceDirs` method like we have `forEachService` in the `types/app-config.go` that just loops through service locations and maybe test that method. We might be using something like that in `exo clean` as well you said?

src/application/service_tester.go
charlierudolph: The role in this case is the service dir (we need your rename pr in)

src/application/service_tester.go
charlierudolph: The `serviceConfig` variable is only used to update the command and dependsOn. Its a value in the `dockerConfigs` map which is what is returned

src/application/tester.go
charlierudolph: That sounds good

src/application/service_tester.go
hugoesthere: It doesn't seem like this is actually modifying the underlying `dockerConfigs` object. I had to reassign `serviceConfig` to `dockerConfigs[s.Role]` in order for this method to return the right objects

hugoesthere:

PR #667, opened 2017-10-11
example apps: update dockerfiles

### LINKED ISSUE #657: example apps: update dockerfiles
We have a few example-apps that pull a dockerfile from `FROM nodesource/node`. Update this to `FROM node` as it has yarn preinstalled (we have this working on EDX).

charlierudolph:
resolves #657

All you @hugobho. I updated tests-cucumbers to use the latest exoservice. Apparently that was all we needed. I'll review your work, if you review mine.


charlierudolph:
Looks good. Thanks for doing this @hugobho. Merge on green.

charlierudolph:
@hugobho I ran through the example apps and switched to simpler docker images for all the places we didn't actually need node.

Issue #641, opened 2017-10-04
exo test: catch interrupt and shutdown containers


charlierudolph:
Removes the need for running exo clean after

hugoesthere:
This would be best done after https://github.com/Originate/exosphere/issues/653 and https://github.com/Originate/exosphere/issues/650 are completed

PR #673, opened 2017-10-11
Catch interrupt and shutdown test containers

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/641


src/application/tester.go
```diff
@@ -37,24 +44,40 @@ func TestApp(appContext types.AppContext, logger *util.Logger, mode composebuild
...
 	logger.Logf("%d tests failed", numFailed)
-	return false, nil
+	return false, false, nil
 }
 
-// TestService runs the tests for the service and return true if the tests passed
+// TestService runs the tests for the service and return true if the tests passed, if they were interrupted,
 // and an error if any
-func TestService(serviceContext types.ServiceContext, logger *util.Logger, mode composebuilder.BuildMode) (bool, error) {
+func TestService(serviceContext types.ServiceContext, logger *util.Logger, mode composebuilder.BuildMode) (bool, bool, error) {
```
charlierudolph: Can we create a `TestResult` struct which has passed and interrupted fields instead of returning two bools? Same for `TestApp`

src/application/tester.go
```diff
@@ -37,24 +44,40 @@ func TestApp(appContext types.AppContext, logger *util.Logger, mode composebuild
...
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt)
+		<-c
+		signal.Stop(c)
+		err = serviceTester.Shutdown()
+		isInterrupted = true
+		wg.Done()
+	}()
+	if isInterrupted {
+		return false, true, err
```
charlierudolph: Thoughts on moving this into ServiceTester? I'd prefer if `ServiceTester.Run()` captures interrupts and shuts itself down if there is one and otherwise shutdowns before returning. This would prevent this method from having to call `Shutdown` as well.

src/application/service_tester.go
```diff
@@ -28,7 +28,7 @@ func NewServiceTester(serviceContext types.ServiceContext, logger *util.Logger,
 	serviceLocation := serviceContext.Location
 	serviceDir := serviceContext.Dir
 	appDir := serviceContext.AppContext.Location
-	dockerComposeProjectName := fmt.Sprintf("%stests", composebuilder.GetDockerComposeProjectName(appDir))
+	dockerComposeProjectName := serviceContext.AppContext.DockerComposeProjectName
```
charlierudolph: @hugobho this change and updating `AppContext` doesn't seem related to this PR. I don't really like its addition anyway, so please create a new PR where can discuss it in isolation.

hugoesthere:

src/types/test_result.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+// TestResult represents the result of a test
+type TestResult struct {
+	Passed      bool
+	Interrupted bool
+	Error       error
```
charlierudolph: I don't like having an error inside of this and would prefer keeping the error separate as that is the common paradigm in go.

src/application/service_tester.go
```diff
@@ -141,11 +144,46 @@ func (s *ServiceTester) setup() error {
...
 // Run runs the tests for the service and return true if the tests passed
 // and an error if any
-func (s *ServiceTester) Run() (int, error) {
+func (s *ServiceTester) Run() (types.TestResult, error) {
 	if err := s.setup(); err != nil {
-		return 1, err
+		return types.TestResult{
+			Passed:      false,
+			Interrupted: false,
+		}, err
```
charlierudolph: Since you are returning an error the values of passed/interrupted don't matter. This could be 
```
return types.TestResult{}, err
```
Please apply this in similar places elsewhere

src/application/service_tester.go
```diff
@@ -141,11 +144,46 @@ func (s *ServiceTester) setup() error {
...
+		signal.Notify(c, os.Interrupt)
+		<-c
+		signal.Stop(c)
+		err = s.Shutdown()
+		isInterrupted = true
+		wg.Done()
+	}()
+	if isInterrupted {
+		return types.TestResult{
+			Passed:      false,
```
charlierudolph: Booleans default to false so this could be `types.TestResult{Interrupted: true}`. Please apply this in similar places elsewhere.

src/application/service_tester.go
```diff
@@ -141,11 +144,38 @@ func (s *ServiceTester) setup() error {
...
+	wg.Add(1)
+	go func() {
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt)
+		<-c
+		signal.Stop(c)
+		err = s.Shutdown()
+		isInterrupted = true
+		wg.Done()
+	}()
```
charlierudolph: I don't understand how this works `wait group` works as we never wait on it. Can we add a feature test for this where the test prints "started", sleeps for a minute, and prints "finished". We hit ctrl-c after "started" is seen, and check that the containers are shut down, we don't see finished, and no tests run, and it exits wth status 1?

hugoesthere: I'm not sure exactly where `wg.Wait()` should go, or even `wg.Add(1)`. In `exo run` we would wait indefinitely until the user hit ctrl-c but in exo-clean that's not the case, so I'm not sure where to increment the counter of the wait group

charlierudolph: Please write a test first. I'm uncertain what is necessary here. I don't think we need the wait group either way. One thing though is that the goroutine needs to communicate with the main thread via a channel (otherwise there is potential data race around the interrupted variable). We may need to move the `runTests` into a goroutine as well and if either complete, they pass something into the channel which the main thread reads from:

```go
done := make(chan TestResult)
// create a goroutine listening for sigInt, if caught push a result into channel
// create a goroutine running the tests, if finished push result into channel
return <- done // wait for one to finish
```
To get the error in case runTests hits an error, we may need a custom struct that contains a testResult and an error.

hugobho:
Closing this in favor of a refactor. Will reopen when that is done.

src/application/tester.go
```diff
@@ -26,45 +28,60 @@ func TestApp(appContext types.AppContext, logger *util.Logger, mode composebuild
...
+			return
+		}
+		testExit <- exitCode
+	}()
+
+	select {
+	case <-shutdown:
+		return types.TestResult{Interrupted: true}, serviceTester.Shutdown()
+	case err := <-testError:
+		serviceTester.Shutdown() // nolint errcheck
```
charlierudolph, line 76: I don't think we need to shutdown in this case and don't believe we were before either.

src/application/tester.go
```diff
@@ -26,45 +28,60 @@ func TestApp(appContext types.AppContext, logger *util.Logger, mode composebuild
 		if serviceContext.Config.Development.Scripts["test"] == "" {
 			logger.Logf("%s has no tests, skipping", serviceContext.Dir)
 		} else {
-			testPassed, err := TestService(serviceContext, logger, mode)
-			if err != nil {
+			testResult, err := TestService(serviceContext, logger, mode, shutdown)
+			switch {
```
charlierudolph, line 32: Thoughts on converting this to three ifs as I don't think these should be mutually exclusive. 

First, log the error if there is one. 
Second, return if interrupted (but don't return the error). 
Finally increment numFailed if needed

src/application/service_tester.go
```diff
@@ -139,13 +139,14 @@ func (s *ServiceTester) setup() error {
...
-func (s *ServiceTester) Run() (int, error) {
+// Start sets up and runs the test for a service, return its result and an error if any
+func (s *ServiceTester) Start() (int, error) {
 	if err := s.setup(); err != nil {
 		return 1, err
 	}
-	return s.runTests()
+
+	exitCode, err := s.runTests()
+	return exitCode, err
```
charlierudolph, line 149: I think this change can be reverted and we can go back to `return s.runTests()`

charlierudolph (commented):
Still need a feature test for this. Let me know if you want to pair on that.

hugoesthere:
@alexdavid do you remember why we held off on this PR? I think it was some refactor you were doing, can't remember if it was the docker-compose refactor or something else

hugoesthere:
Closing in favor of some of the refactor we're doing. Will re-open once the codebase stablizes

PR #674, opened 2017-10-11
Move alb creation into service modules

### LINKED ISSUE #655: exo deploy: bug in initial deploy
@alexdavid says this is due to a bug in our initial deploy where we are missing a depends on

screen shot 2017-10-06 at 6 15 46 pm

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/655


charlierudolph (approved):
Looks good to me although I assume having someone with more terraform experience would be better to review this. @alexdavid can you please review?

src/terraform/modules/aws/public-service/alb.tf
```diff
@@ -51,16 +49,3 @@ resource "aws_alb_listener" "external" {
     type             = "forward"
   }
 }
-
-resource "aws_alb_listener" "internal" {
-  count = "${var.internal ? 1 : 0}"
```
alexdavid, line 56: Even though there's some repetition because of the `depends_on` I don't hate it because differences in this abl_listener makes this more easily configurable in the future.

👍

alexdavid (approved):
Looks great! Thanks for fixing this

PR #681, opened 2017-10-12
exo-deploy: pass exocom service routes as variable

### LINKED ISSUE #672: exo deploy: reduce terraform changes
Currently terraform changes if the `exocom` env service routes change (which happens when any service sends or receives a new message). Can we make the exocom env variables a variable so that these type of changes do not create the need for a manual deploy?

@hugobho

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/672


src/terraform/templates/exocom.tf
```diff
@@ -27,10 +31,8 @@ module "exocom_service" {
   docker_image          = "{{{dockerImage}}}"
   env                   = "production"
   environment_variables = {
-    ROLE = "exocom"
-    SERVICE_ROUTES = <<EOF
-{{{serviceRoutes}}}
-EOF
+    ROLE           = "exocom"
+    SERVICE_ROUTES = "${var.exocom_service_routes}"
```
charlierudolph: thoughts on updating this to be `environment_variables = "${var.exocom_env_vars}"`? That is what we are doing with the services and I thought it might make sense to do the same here.

I don't think the role is needed (but we can remove that separately).

charlierudolph (approved):

PR #682, opened 2017-10-13
release 0.28.1

hugoesthere:


hugoesthere:
Actually, holding off on this release for now as there seems to be an issue with how exocom service routes are being passed.

PR #683, opened 2017-10-13
exo deploy: don't build images that won't be pushed

### LINKED ISSUE #669: exo deploy: don't build images that won't be pushed
If an image would not be pushed to ECR, lets not build it.

charlierudolph:
resolves #669

This includes a refactor of the deployment code.Tried to pull more of the application logic into application/deployer and made aws/ecr_helpers more generic and not reliant on any of our application types.


src/aws/ecr_client_helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return "", err
+	}
+	for _, repositoryInfo := range result.Repositories {
+		if *repositoryInfo.RepositoryName == repositoryName {
+			return *repositoryInfo.RepositoryUri, nil
+		}
+	}
+	return "", nil
+}
```
charlierudolph, line 78: No new code here. Moved functions from `aws/ecr_helpers.go` and `aws/index.go` into this file. Did make the top two methods public

src/aws/ecr_helpers.go
```diff
@@ -1,194 +0,0 @@
-package aws
```
charlierudolph, line 1: Most of this logic was extracted into into `application/deployer/push_application_images.go`, `application/deployer/push_service_image.go` and `aws/push_image_helper`

src/application/deployer/push_service_image.go
```diff
@@ -0,0 +1,107 @@
...
+	DeployConfig     types.DeployConfig
+	DockerComposeDir string
+	EcrAuth          string
+	EcrClient        *ecr.ECR
+	ImageName        string
+	ServiceLocation  string
+	BuildImage       bool
+}
+
+// PushServiceImage pushes a single service image to ECR, building or pulling if needed
```
charlierudolph: This is also used for pushing a dependency image. So open to a better name

hugoesthere: What about just `PushImage`?

src/application/deployer/push_service_image.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func buildOrPullImage(options PushServiceImageOptions) error {
+	opts := compose.ImageOptions{
+		DockerComposeDir: options.DockerComposeDir,
+		Logger:           options.DeployConfig.Logger,
+		Env:              []string{fmt.Sprintf("COMPOSE_PROJECT_NAME=%s", options.DeployConfig.DockerComposeProjectName)},
+		ImageName:        options.ImageName,
+	}
+	if options.BuildImage {
```
hugoesthere: I can't seem to find where this boolean is being set for dependencies

src/application/deployer/push_application_images.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/Originate/exosphere/src/aws"
+	"github.com/Originate/exosphere/src/docker/tools"
+	"github.com/Originate/exosphere/src/types"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecr"
+)
+
+// PushApplicationImagesOptions is the options to PushApplicationImages
+type PushApplicationImagesOptions struct {
```
hugoesthere: It feels a little more clear to me (but not that much better) if we call these files and their respective structs `PushImage`/`push_images.go` and `PushImage`/`push_image.go`. 

Also, this struct seems a little unnecessary as it only contains two fields that are only used once by `PushApplicationImages`. It would almost make more sense to me to pass in two params instead of overloading the words `options` and having slightly confusing struct names.

What do you think?

src/application/deployer/push_service_image.go
```diff
@@ -0,0 +1,107 @@
...
+	"strings"
+
+	"github.com/Originate/exosphere/src/aws"
+	"github.com/Originate/exosphere/src/docker/compose"
+	"github.com/Originate/exosphere/src/types"
+	"github.com/aws/aws-sdk-go/service/ecr"
+)
+
+// PushServiceImageOptions is the options to PushServiceImage
+type PushServiceImageOptions struct {
```
hugoesthere: How are we defining the difference between `options` and `config` in this codebase? I think we might be using both interchangeably because `config` seems to make more sense to me here. There is added confusion too because now it seems we have a `PushImageHelper` struct with very similar fields. Would it be possible to combine those structs at all?

src/aws/push_image_helper.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	for _, imageDetail := range result.ImageDetails {
+		for _, tag := range imageDetail.ImageTags {
+			if *tag == p.ImageVersion {
+				return true, nil
+			}
+		}
+	}
+	return false, nil
+}
```
hugoesthere: I might even call this file `RepositoryHelper` or something as it contains functions pertaining to what's in a image repo

src/application/deployer/push_application_images.go
```diff
@@ -0,0 +1,52 @@
...
+			ImageName:        imageName,
+			ServiceLocation:  serviceData[serviceRole].Location,
+		})
+		if err != nil {
+			return nil, err
+		}
+		imagesMap[serviceRole] = taggedImage
+	}
+	return imagesMap, nil
+}
```
hugoesthere, line 47: It seems like we usually give file names the noun version of what they do, like `application_image_pusher.go`. Although that sounds weird here. What do you think?

src/application/deployer/push_service_image.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	if options.BuildImage {
+		options.DeployConfig.Logger.Logf("Building image: %s...", options.ImageName)
+		return compose.BuildImage(opts)
+	}
+	options.DeployConfig.Logger.Logf("Pulling image: %s...", options.ImageName)
+	return compose.PullImage(opts)
+}
+
+func getPushImageHelper(options PushServiceImageOptions) (*aws.PushImageHelper, error) {
```
hugoesthere: It seems confusing to me what a `PushImageHelper` is exactly (see comment above), but perhaps this will be more clear when we switch to calling it a `RepositoryHelper` (see comment below).

src/application/deployer/push_service_image.go
```diff
@@ -0,0 +1,107 @@
...
+		EcrAuth:        options.EcrAuth,
+		EcrClient:      options.EcrClient,
+		ImageName:      options.ImageName,
+		ImageVersion:   version,
+		RepositoryName: repositoryName,
+		RepositoryURI:  repositoryURI,
+	}, nil
+}
+
+// returns an image with version tag if applicable. uses the application version otherwise
```
hugoesthere: This is something I forgot to change, but the comment needs to be updated to say something like "uses the commit sha otherwise".

src/application/deployer/push_service_image.go
```diff
@@ -0,0 +1,107 @@
...
+func getRepositoryConfig(appDir string, serviceLocation, imageName string) (string, string, error) {
```
hugoesthere: What do you think about splitting this function up into `getRepositoryName` and `getImageVersion`? It seems to be doing two things and lumping them under `getRepositoryConfig` (which I did, admittedly) seems confusing. It's also confusing that what Docker calls an "image" is kind of similar to what AWS calls "repository"

hugoesthere (commented):
Looks good, thanks for doing this refactor - it's been a long time coming. Just some comments on naming to make things more clear, otherwise I agree with the move from `aws/` to `application/deployer`.

src/application/deployer/push_application_images.go
charlierudolph: I'll move the options struct into its own file. The options is only used for what goes into the public function.

I'm good passing in it as two arguments

src/application/deployer/push_application_images.go
charlierudolph: I think having the file name match the only public function exposed is fine. I think the noun makes the most sense when you export a struct. Otherwise this could be something like `application_image_helpers`

src/application/deployer/push_service_image.go
charlierudolph: Options / Config isn't very different to me. For objects that are only used as function arguments (and have not alternative meaning), I think the Options suffix is a good idea. Config is more for things that are actually user configuration. That said, the comment should describe well what it represents. 

The `PushImageHelper` is ECR specific and doesn't know about our application types. It has its own API that is specific to what its doing. This is attempting to make the break between when we are using ECR and our application logic. I think it makes it easier to think about and allows us to switch out ECR if we wanted to host our images elsewhere.

src/application/deployer/push_service_image.go
charlierudolph: Removed and used whether or not we have a serviceLocation

charlierudolph:
@hugobho comments addressed. Tried this out with EDX and appears to be working as expected. Please take over this PR next week.

PR #684, opened 2017-10-13
Fix exocom routes bug

hugoesthere:
Had to triple encode the service routes, but it works. Unit and deploy tested.

### DIFF
--- a/src/config/exocom_production_dependency.go
+++ b/src/config/exocom_production_dependency.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/Originate/exosphere/src/types"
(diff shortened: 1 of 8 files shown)


src/config/exocom_production_dependency.go
```diff
@@ -47,6 +48,13 @@ func (e *exocomProductionDependency) GetDeploymentServiceEnvVariables(secrets ty
...
-	return map[string]string{"exocom_service_routes": serviceRoutes}, err
+	serviceRoutes, err := exocomDevelopmentDependency.compileServiceRoutes()
+	if err != nil {
+		return map[string]string{}, err
+	}
+	serviceRoutesJSON, err := json.Marshal(serviceRoutes)
+	if err != nil {
+		return map[string]string{}, err
+	}
+	return map[string]string{"SERVICE_ROUTES": string(serviceRoutesJSON)}, err
```
charlierudolph, line 61: Can we please add a comment (or even better a test) so we know why this encoding in this way?

charlierudolph (approved):

PR #694, opened 2017-10-19
Initial logger refactor work

hugoesthere:
Seems like we could only port part of one command over. Everything else uses docker-compose :/

screen shot 2017-10-19 at 4 46 08 pm


hugoesthere:
Closing in favor of: https://github.com/Originate/exosphere/issues/699

PR #696, opened 2017-10-23
refactor: remove app/initializer part 2

### LINKED ISSUE #666: refactor A2: remove app/initializer and app/runner
* Create a docker/runner which combines simplified app/initializer and app/runner

dependent on #665

charlierudolph:
resolves #666


src/docker/runner/image_group.go
```diff
@@ -0,0 +1,8 @@
+package runner
+
+// ImageGroup is a group of images to run at once
+type ImageGroup struct {
+	ID          string
+	Names       []string
```
hugoesthere: I think we have to give this field a better name. This is really the key for a service/dependency in `docker-compose.yml` right? `ImageGroup.Names` makes me think of docker image names which are like `tmp_user-service`. Docker-compose calls those keys service names, that wouldn't be a bad name although I'm not sure if that will cause confusion with what we call service names.

src/application/runner.go
```diff
@@ -67,76 +69,43 @@ func (r *Runner) getDependencyContainerNames() []string {
...
-		}
-		go func(role string, onlineTextRegex *regexp.Regexp) {
-			r.waitForOnlineText(cmdPlus, role, onlineTextRegex)
-			wg.Done()
-		}(role, onlineTextRegex)
+	runOptions := runner.RunOptions{
+		ImageGroups: []runner.ImageGroup{
+			{
+				ID:          "dependencies",
+				Names:       r.getDependencyContainerNames(),
```
hugoesthere: Should we just start calling this `getDependencyNames()` or something? Right now it seems like we're using `dependencyName`, `serviceRole`, and `ImageGroup.Name` interchangeably. I know why we do that but I feel like this could potentially get confusing if we return to this code months from now.

It seems like in this file at least they _are_ just dependency names, or even dependency keys.

src/application/service_tester.go
```diff
@@ -126,42 +120,43 @@ func (s *ServiceTester) getServiceOnlineTexts() map[string]string {
...
+	dockerComposeDir := path.Join(s.ServiceLocation, "tests", "tmp")
+	dockerConfigs, err := s.getDockerComposeConfig()
+	if err != nil {
+		return runner.RunOptions{}, err
+	}
+	return runner.RunOptions{
+		ImageGroups: []runner.ImageGroup{
+			{
+				ID:          "dependencies",
+				Names:       s.getDependencyContainerNames(),
```
hugoesthere: Same comment here as above

src/docker/tools/index.go
```diff
@@ -42,6 +42,22 @@ func GetDockerCompose(dockerComposePath string) (result types.DockerCompose, err
 	return result, nil
 }
 
+// GetExitCode returns the exit code for the given container
+func GetExitCode(containerName string) (int, error) {
+	c, err := client.NewEnvClient()
+	if err != nil {
+		return 1, err
+	}
+	containerJSON, err := c.ContainerInspect(context.Background(), containerName)
```
hugoesthere, line 51: 👍

src/docker/runner/shutdown.go
```diff
@@ -0,0 +1,30 @@
...
+	err := compose.KillAllContainers(compose.BaseOptions{
+		DockerComposeDir: options.DockerComposeDir,
+		Logger:           options.Logger,
+		Env:              []string{fmt.Sprintf("COMPOSE_PROJECT_NAME=%s", options.DockerComposeProjectName)},
+	})
+	if err != nil {
+		return errors.Wrap(err, "Failed to shutdown the app")
+	}
+	return nil
+}
```
hugoesthere: What do you think about moving everything from `docker/runner` into `docker/compose/runner` or something similar? I'm worried we'll get `docker` and `docker-compose` processes confused. It'd be nice to keep them in just two separate dirs.

hugoesthere (commented):
Great work! I would make sure @alexdavid gets a look at this to ensure that it lines up with the roadmap he and Kevin were talking about last Friday.

src/docker/runner/image_group.go
charlierudolph: How about `containerNames` since they ended up being used as the name of the container?

hugoesthere: I think technically they're still different from containerNames for the regular services right? Those service container names are prefixed. That would also be confusing because we don't list a container name for services in `docker-compose.yml`, what we're really referring to are the keys

src/docker/runner/shutdown.go
charlierudolph: I'll move it to `composerunner` to match `composebuilder`

src/docker/runner/image_group.go
charlierudolph: The image names are prefixed with the compose project name, but the container names match the keys in the `docker-compose.yml`. I figured this out when creating the get exit code function.

src/application/runner.go
charlierudolph: I'd like to leave this as is given that #698 will remove this.

charlierudolph:

hugoesthere (approved):

PR #701, opened 2017-10-24
Add bastion sg to rds sg

hugoesthere:
Not sure if we should be doing this for every case, but thought I'd put up a PR for it.

### DIFF
--- a/src/terraform/modules/aws/dependencies/rds/rds.tf
+++ b/src/terraform/modules/aws/dependencies/rds/rds.tf
@@ -46,6 +46,13 @@ resource "aws_security_group" "rds" {
     security_groups = ["${var.ecs_security_group}"]
   }
 
+  ingress {
+    from_port       = 5432
+    to_port         = 0
+    protocol        = -1
+    security_groups = ["${var.bastion_security_group}"]
+  }
+
   tags {
     Name        = "${var.name}-rds"
     Environment = "${var.env}"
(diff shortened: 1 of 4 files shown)


alexdavid:
Just talked to @chrizkim about this and he suggests that we don't do this. If needed we can double port forward to the ecs cluster.

hugoesthere:
👍

alexdavid (approved):

src/terraform/modules/aws/dependencies/rds/rds.tf
```diff
@@ -46,6 +46,13 @@ resource "aws_security_group" "rds" {
     security_groups = ["${var.ecs_security_group}"]
   }
 
+  ingress {
+    from_port       = 5432
+    to_port         = 0
+    protocol        = -1
```
alexdavid: I think this should be
```
from_port       = 5432
to_port         = 5432
protocol        = 6
```

alexdavid (changes requested):

src/terraform/modules/aws/dependencies/rds/rds.tf
alexdavid: You can also use 
```
protocol        = "tcp"
```
I think

hugoesthere:
@alexdavid comments addressed

alexdavid (approved):

PR #705, opened 2017-10-25
exo clean: simplify

charlierudolph:
This doesn't really have anything to do with exosphere and causes the command to be slower.


hugoesthere (commented):
Hmm..we had this because we used to have problems with exosphere building so many new images that a lot of dangling images would appear and eventually use of all the memory reserved for docker. Is that not something we're worried about anymore?

charlierudolph:
Okay, I didn't every understand what exactly this was doing. Well having the headers before starting the command will be nice. I'll close this.

Issue #699, opened 2017-10-23
simplify output


charlierudolph:
Stop parsing the docker compose output and instead just output it directly. With #698, we shouldn't be injecting any more output into the `docker-compose up` process.

For other processes, like `exo clean` and `exo deploy`, output the child command directly (this will probably mean stop using `go-execplus`) and instead only add banner rows to signify that something is being run.

This may result in a complete removal of the logger.

@alexdavid @hugobho please review

hugoesthere:
I think we'll still need to logger for logging things like `exo clean removing test containers` or something, we can simply pipe the docker-compose output to the logger without parsing it. 

This might mess up the indentation though - I'm not sure how important that is.

charlierudolph:
That would be a banner row to me which separates calls to child processes). Another example would be for `exo run` which has four sections `docker-compose pull`, `docker-compose build`, `docker-compose up`, and `docker-compose down`

charlierudolph:
See `printHeader` in `src/cmd/template` for an example of what I am considering a banner row

charlierudolph:
For breaking this up lets do this one command at a time.
* [x] exo clean
  * Remove `exo clean` from the start of each line
  * add banner rows and then pipe the rest of the output directly
* [x] exo deploy
  * Remove `exo deploy` from the start of each line
  * add banner rows and then pipe the rest of the output directly
* [x] exo run
* [x] exo test

Pass in an `io.Writer` instead of a logger and use `fmt.Fprint` and assign child process stdout / stderr to it

hugoesthere:
When you say pipe the output directly do you mean use this simplified logger which takes in `io.Writer` and uses `fmt.Fprint`?

charlierudolph:
For `pipe the output directly` I mean we will switch from using go-execplus and instead be using the ordinary `exec`. When you create a `Cmd` instance, you can assign as io.Writer to `Stdout` and `Stderr` https://golang.org/pkg/os/exec/#Cmd

hugoesthere:
When do we decide to switch from go-execplus to exec? Didn't we mostly use go-execplus for monitoring online text? Does that mean we won't need go-execplus either?

charlierudolph:
I don't think we'll need go-execplus in the source code anymore, but will continue to use it in the exosphere tests. Please start with exo clean / exo deploy since those don't touch online text. Lets not cleanup the output in exo run / exo test until #698 is done

charlierudolph:
[...]

PR #706, opened 2017-10-25
Switch exo deploy to io.writer

hugoesthere:
Resolving https://github.com/Originate/exosphere/issues/699


charlierudolph:
Can you please add an example of what the output of `exo-deploy` will look like?

hugoesthere:
Example output:
screen shot 2017-10-25 at 2 13 27 pm
screen shot 2017-10-25 at 2 14 59 pm
screen shot 2017-10-25 at 2 14 32 pm

charlierudolph:
Lets remove "We are about to deploy an application". Ideally I think the banners should only be used when they are followed by the output of a child process. Unless the validations steps take a long time, I think we can remove those headers.

charlierudolph:
Thoughts on the following: only use the banner rows when dealing with child processes. If there isn't going to be any output don't use a banner row and instead just print it. Thus the first 5 would all appear after each other with no lines of stars. The child process part could be updated to something like
```
Generating terraform files...
Retrieving remove state...
*******************************
<output>
*******************************
```
Then I'm fine keeping all the current messages and we are using the lines of stars only to separate output for child processes.

hugoesthere:
So are we putting the banner around the child process itself or the "command" that signifies the start of the child process? Because in exo clean/template test we're doing the latter.

charlierudolph:
I think we should move to putting it around the child process output as that what we want to separate from the rest.

hugoesthere:
Ok, output looks closer to this now:
screen shot 2017-10-25 at 3 49 20 pm
screen shot 2017-10-25 at 3 49 37 pm

charlierudolph (approved):

PR #707, opened 2017-10-25
Release 0.28.4

hugoesthere:


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.28.4 (2017-10-25)
+
+#### New Features
+
+* `exo-deploy`: allow traffic from bastion instances to RDS secruity groups
+
+#### Bug fixes
+
+* `exo-run`: fix ctrl-c shutdown bug introduced in previous release
```
charlierudolph: I don't think bug was every released.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.28.4 (2017-10-25)
+
+#### New Features
+
+* `exo-deploy`: allow traffic from bastion instances to RDS secruity groups
```
charlierudolph: Were you able to test this with a deploy of EDX and be able to connect? Also thinking back on the implementation the 5432 port is specific to Postgres. Mysql for instance uses a different port but would also use RDS

hugoesthere: Good point. For this release I'll note that it only allows postgres traffic. I did a test deploy and let it go until the plan step. Then I read the plan to verify that the sg was being modified, but I didn't actually apply the deploy. Should I test that right now?

@alexdavid should we allow all traffic from bastion for each of the respective db ports?

charlierudolph: I don't think we necessarily need to support the other versions at the moment but at least build this in a way that we can add support for the other dbs down the line

hugoesthere: Yes it's certainly built in a way to allow support for other dbs, even right now when we're only supporting postgres

charlierudolph: Shouldn't the port be a variable though? Where 5432 is defined once as part of the postgres dependency and we would have a different one for mysql

alexdavid: Good point, didn't think about that. We honestly could just open up all communication to the bastion instance like we do for the ECS cluster.

charlierudolph: I like the simpler approach

src/cmd/deploy.go
```diff
@@ -62,7 +62,7 @@ var deployCmd = &cobra.Command{
 			DeployServicesOnly: deployServicesFlag,
 
 			// git commit hash of the Terraform modules in Originate/exosphere we are using
-			TerraformModulesRef: "01891552",
+			TerraformModulesRef: "f95d4b0a",
```
charlierudolph: I think this needs to be updated.

charlierudolph (commented):
Good to merge after updating the terraform module hash

PR #709, opened 2017-10-25
allow bastion traffic from all ports

hugoesthere:
per our discussion https://github.com/Originate/exosphere/pull/707#discussion_r147002957


alexdavid (approved):

alexdavid:
Although at this point we might as well just set protocol to `"all"` to allow ICMP/etc...

PR #710, opened 2017-10-25
remove ability to silence services / dependencies

charlierudolph:
Ran into this when working on stopping the spin up of dependencies. This was a feature that I didn't think of a good way to update and with our change in the logger, this feature doesn't make sense to me anymore. We can reintroduce this down the line using a docker-compose feature if we want it.


hugoesthere (approved):
👍 I had gotten rid of this in an initial logger refactor - definitely agree with this decision.

PR #711, opened 2017-10-26
stop spinning up dependencies

### LINKED ISSUE #698: stop spinning up dependencies first
Instead of using docker-compose to spin up dependencies and then services, remove this feature and spin up all at once. This is similar to what is done in production so its important that services can handle coming up before their dependencies.

This may remove the need for "online text", unless we want to keep telling the user when everything has booted up. I don't think thats very important though.

@alexdavid @hugobho please review

charlierudolph:
resolves #698

* also removed `restart` during the cleanup (very minimal)


example-apps/external-dependency/mongo/server.js
```diff
@@ -4,7 +4,7 @@ const N = require('nitroglycerin')
 
 bootstrap({
   beforeAll: function (done) {
-    MongoClient.connect(getMongoAddress(), N(function() {
+    MongoClient.connect(getMongoAddress(), {autoReconnect: true}, N(function() {
```
hugoesthere, line 7: 😂

src/application/service_tester.go
```diff
@@ -110,20 +103,14 @@ func (s *ServiceTester) getDockerComposeConfig() (types.DockerConfigs, error) {
...
-
 // Run runs the tests for the service and return true if the tests passed
 // and an error if any
 func (s *ServiceTester) Run() (int, error) {
-	err := composerunner.Run(s.RunOptions)
+	cmdPlus, err := composerunner.Run(s.RunOptions)
+	if err != nil {
+		return 1, err
+	}
+	err = cmdPlus.WaitForText(fmt.Sprintf("%s exited with code", s.DirName), time.Hour)
```
hugoesthere, line 113: How do we plan on handling this after switching to `exec`?

hugoesthere (approved):
Good riddance, online text.

src/application/service_tester.go
charlierudolph: Hmm, I would guess we'll need to poll docker checking if the container has exited?

charlierudolph: Actually `docker-compose up` has a pretty useful option `--abort-on-container-exit`. We would use that only in tests

PR #713, opened 2017-10-26
Refactor build mode struct

hugoesthere:


src/application/deployer/index.go
```diff
@@ -28,8 +28,11 @@ func StartDeploy(deployConfig types.DeployConfig) error {
 	dockerConfigs, err := composebuilder.GetApplicationDockerConfigs(composebuilder.ApplicationOptions{
 		AppConfig: deployConfig.AppConfig,
 		AppDir:    deployConfig.AppDir,
-		BuildMode: composebuilder.BuildModeDeployProduction,
-		HomeDir:   deployConfig.HomeDir,
+		BuildMode: composebuilder.BuildMode{
+			Type:        composebuilder.BuildModeTypeDeploy,
+			Environment: composebuilder.BuildModeEnvironmentProduction,
```
charlierudolph: I don't think the environment is needed when we are deploying

src/docker/composebuilder/application_test.go
```diff
@@ -30,8 +30,11 @@ var _ = Describe("composebuilder", func() {
 			dockerConfigs, err := composebuilder.GetApplicationDockerConfigs(composebuilder.ApplicationOptions{
 				AppConfig: appConfig,
 				AppDir:    appDir,
-				BuildMode: composebuilder.BuildModeDeployProduction,
-				HomeDir:   homeDir,
+				BuildMode: composebuilder.BuildMode{
+					Type:        composebuilder.BuildModeTypeDeploy,
+					Environment: composebuilder.BuildModeEnvironmentProduction,
```
charlierudolph: Same here. I don't think this should be needed.

charlierudolph (approved):

PR #714, opened 2017-10-26
Simplify exo test and exo run logger

hugoesthere:
exo test output looks pretty ugly now imo. open to suggestions
screen shot 2017-10-26 at 4 45 43 pm
screen shot 2017-10-26 at 4 46 23 pm
screen shot 2017-10-26 at 4 46 40 pm
screen shot 2017-10-26 at 4 46 48 pm
screen shot 2017-10-26 at 4 46 58 pm


src/docker/compose/index.go
```diff
@@ -12,44 +11,30 @@ func BuildImage(opts ImageOptions) error {
...
+	return util.RunAndPipe(opts.DockerComposeDir, opts.Env, opts.Writer, "docker-compose", "build")
 }
 
 // KillAllContainers kills all the containers
 func KillAllContainers(opts BaseOptions) error {
 	return util.RunAndPipe(opts.DockerComposeDir, opts.Env, opts.Writer, "docker-compose", "down")
 }
 
-// KillContainer kills the docker container of the given service
-func KillContainer(opts ImageOptions) error {
```
hugoesthere, line 29: This was dead code that was never being called. Does our linter not catch things like this?

charlierudolph: Its a public function so that might be a function we as a library are exporting. If it was private and unused the linter would catch it

features/run/production.feature
```diff
@@ -12,15 +12,15 @@ Feature: running Exosphere applications
 
   Scenario: booting an exosphere application in development mode
     When starting "exo run" in my application directory
-    And it prints "nginx nginx online" in the terminal
+    And it prints "Attaching to nginx" in the terminal
```
hugoesthere, line 15: The output is actually now `nginx    | nginx online`, but for some reason execplus doesn't pick that up if I tell it to wait on that text. I had to wait on the line right before `nginx    | nginx online`, which is this.

hugoesthere: Running into a similar problem on `features/test/complex-test.feature` , on the step:
```
    And I eventually see:
      """
      1 scenario (1 passed)
      """
```
screen shot 2017-10-27 at 1 16 04 pm

charlierudolph: For the nginx part, your solution is fine. My PR addressed the complex-test feature

PR #716, opened 2017-10-27
Gracefully shutdown test containers

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/641


src/cmd/test.go
```diff
@@ -32,16 +34,21 @@ var testCmd = &cobra.Command{
 			buildMode.Mount = false
 		}
 
-		var testsPassed bool
+		// Listen to SIGINT
```
charlierudolph: I don't think this comment is need.

src/cmd/test.go
```diff
@@ -32,16 +34,21 @@ var testCmd = &cobra.Command{
...
 		} else {
-			testsPassed, err = application.TestApp(context.AppContext, writer, buildMode)
+			testResult, err = application.TestApp(context.AppContext, writer, buildMode, shutdownChannel)
 		}
 		if err != nil {
 			panic(err)
 		}
-		if !testsPassed {
+		signal.Stop(shutdownChannel)
+		if !testResult.Passed && !testResult.Interrupted {
```
charlierudolph: I think we can go leave this as it was: `if !testsPassed`. passed should be false on interrupt and the exit code also doesn't really matter during interrupt.

src/application/tester.go
```diff
@@ -29,46 +30,61 @@ func TestApp(appContext types.AppContext, writer io.Writer, mode composebuilder.
...
+	case err := <-testError:
+		serviceTester.Shutdown() // nolint errcheck
+		return types.TestResult{}, err
+	case exitCode := <-testExit:
+		testResult := types.TestResult{Passed: exitCode == 0}
+		result := "failed"
+		if testResult.Passed {
+			result = "passed"
+		}
+		fmt.Fprintf(writer, "'%s' tests %s\n", serviceContext.Dir, result)
```
charlierudolph: Note: when merging, the logger updates removes this print statement.

charlierudolph (approved):

hugoesthere:
@charlierudolph can you take one more look? I want to make sure I merged everything correctly.

example-apps/tests-long-running/test-service/test-script
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -e
+
+echo "tests starting"
+sleep 1000s
+echo "tests finished"
+
```
charlierudolph: this trailing empty line can be deleted

features/test/interrupt.feature
```diff
@@ -0,0 +1,18 @@
...
+  - press ctrl-c to interrupt running tests
+  - Exosphere cleans up running test containers
+
+
+  Scenario: clean up test containers
+    Given I am in the root directory of the "tests-long-running" example application
+    When starting "exo test" in my application directory
+    Then it prints "Testing service 'test-service'" in the terminal
+    And it prints "tests starting" in the terminal
+    And I send an interrupt signal
```
charlierudolph: This should be the `When` and thus everything above uses Given / And. Thus it would be something like:
```cucumber
Given ...
And ...
And ...
And ...
When I send an interrupt signal
```

charlierudolph (approved):

PR #722, opened 2017-10-31
Build compose project name from app name

### LINKED ISSUE #671: exo deploy: remove dependency on folder name
Currently, the prefix added to pushed images has the prefix based on the folder name instead of the app name (from application.yml). For example on CircleCI, if I don't set the working directory then the project is checked out in a folder called "project" and the images pushed to ECR have the prefix "project_" even though the app name is "edx"

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/671


charlierudolph (approved):
Merge on green

PR #723, opened 2017-10-31
Print time metrics

### LINKED ISSUE #680: time metrics
In order to make it easier to address bottlenecks and learn what is slow, I'd like us to add in a bunch of timing information. On EDX, I tried to address "testing" when I think the real slow downs were building and spinning up dependencies.  

For `exo test`  (after each individual and a total metric at the end)
* building 
* spinning up dependencies
* testing

For `exo run`
* building
* spinning up deps
* spinning up services

For `exo deploy`
* building
* pushing
* planning
* applying

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/680

Example output:
screen shot 2017-10-31 at 3 48 22 pm
screen shot 2017-10-31 at 3 47 46 pm


charlierudolph:
Thoughts on updating the header/footer to
```
>>> docker-compose pull
...
...
...
<<< done in 1.8s
```
I don't think repeating the command in the footer is needed

hugoesthere:
I'm open to that, my only reservation is that once in a while I see behavior like this:

```
docker-compose up
...
...

ctrl-c

docker-compose down
...
...
docker-compose down done in x seconds
docker-compose up down in x seconds
```

Since `docker-compose down` is ending the initial `docker-compose up` process we see that double print. I ran into this the first time I tested this but haven't seen it since.

charlierudolph:
Okay how about giving each command an id then that starts from one (ideally using a global so it doesn't need to be passed in:

```
1 >>> docker-compose up
...
...

ctrl-c

2 >>> docker-compose down
...
...
2 <<< done in x seconds
1 <<< done in x seconds
```
I don't want to reprint the commands as we have some that may be long

hugoesthere:
Hmm, I can see that being more confusing than helpful. We could just go back to your original idea of printing `<<< done in x seconds` w/out id, since the only time we have this problem is when a user is closing down `exo run`. I don't see it being particularly useful to know how long you keep your app running, plus it should be obvious the difference between `up` and `down` when one duration is a couple minutes and the other is a couple seconds.

charlierudolph:
K. yeah that shouldn't be a big issue and think minimal is the best place to start

hugoesthere:
Comments addressed. Example output:
screen shot 2017-11-01 at 11 10 27 am

src/util/print.go
```diff
@@ -25,6 +26,15 @@ func PrintSectionHeader(writer io.Writer, text string) {
 	}
 }
 
+// PrintTimeElapsed prints the time elapsed since startTime
+func PrintTimeElapsed(writer io.Writer, startTime time.Time) {
```
charlierudolph: Can we make this `PrintCommandFooter` and thoughts on having `PrintCommandHeader` return the start time?

charlierudolph (approved):

src/util/print.go
hugoesthere: Hm, I really don't like conflating printing commands w/time metrics, because now every time we want to print a command header we have to deal with a start time being returned - or we would have to print a command header in order to get a start time. 

If we keep those ideas separate I would almost argue that we should keep it as `PrintTimeElapsed` to be clear printing commands/measuring time aren't tightly coupled.

charlierudolph: Is this used anywhere aside from in runAndPipe? I agree it is a coupling but I believe these should always be used together so I'd prefer the names to reflect that.

I think we should have this function take in a duration instead of a startTime then (moving the time.Since into runAndPipe)

hugoesthere: It will be used here https://github.com/Originate/exosphere/pull/725, although that PR needs to be updated once this one is shipped.

Okay, I think that makes sense to call if `PrintCommandFooter`. I also moved the `time.Since` into runAndPipe.

src/util/exec.go
```diff
@@ -44,9 +45,11 @@ func RunAndPipe(dir string, env []string, writer io.Writer, commandWords ...stri
 	cmd.Stdout = writer
 	cmd.Stderr = writer
 	PrintCommandHeader(writer, strings.Join(commandWords, " "))
+	startTime := time.Now()
 	if err := cmd.Run(); err != nil {
 		return errors.Wrapf(err, "Error running '%s'", strings.Join(commandWords, " "))
 	}
+	PrintSectionFooter(writer, time.Since(startTime))
```
charlierudolph: Section -> Command

PR #724, opened 2017-11-01
exo deploy: print image push progress

### LINKED ISSUE #670: exo deploy: push image progress
when pushing images to ECR do we get any progress indications that we could display to the user?
This has been the slowest step and I would like to give % complete updates. (especially because I have a slow connection in utah)

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/670

screen shot 2017-11-01 at 11 58 20 am

### DIFF
--- a/src/docker/tools/index.go
+++ b/src/docker/tools/index.go
@@ -1,9 +1,11 @@
 package tools
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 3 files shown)


src/docker/tools/index.go
```diff
@@ -134,38 +136,41 @@ func TagImage(srcImage, targetImage string) error {
...
-	outputArr := strings.Split(string(output), "\n")
-	errorMessage := outputArr[len(outputArr)-2]
-	errorObject := struct {
+func printPushProgress(writer io.Writer, output string) error {
+	outputObject := struct {
+		Status      string      `json:"status"`
+		ID          string      `json:"id"`
 		ErrorDetail interface{} `json:"errorDetail"`
 		Error       string      `json:"error"`
 	}{}
```
charlierudolph, line 166: Can we define this type in its own file? maybe named something like `PushImageProgress`?

charlierudolph (approved):

charlierudolph:
Unrelated to this PR but in the screenshot, it appears the build / push run together. Can we use PrintSectionHeader in the exo deploy output? (in a separate PR)

src/docker/tools/index.go
hugoesthere: Do you think that's useful since it's not used anywhere else besides in this function?

charlierudolph: Good as is.  This is the first time I've seen this.

Issue #697, opened 2017-10-23
commit docker compose files


charlierudolph:
Put docker compose files in a dedicated folder in the repo. Possibly something like the following:

* docker-compose
  * [ ] run_development.yml
  * [ ] run_production.yml
  * [ ] test.yml (when running tests, pick the service that needs to be run)
* ostmpdir
  * [ ] docker-compose.yml (for deployment)

Each are generated on demand, if needed

* [x] Switch `exo clean` to clean from these files.

@alexdavid @hugobho please review

charlierudolph:
Deployment technically also uses a compose file. Do we think that is worth committing? That could go in a os tempdir if not

hugoesthere:
I don't think so, it's really only building images. What if we had `exo-deploy` read from `run_production.yml`? That would make the most sense to me, that way what you test with `exo run --production` is what's pushed to ECR

charlierudolph:
`run_production.yml` may include development dependencies though

charlierudolph:
Lets do this one compose file at a time. The `run` compose files should be pretty easy. The test compose file will be a bit of a refactor as we create one for the entire application. Thus probably do this in two PRs

PR #727, opened 2017-11-02
Commit exo-run docker-compose files

hugoesthere:
Resolves part of: https://github.com/Originate/exosphere/issues/697


src/application/runner.go
```diff
@@ -38,7 +39,8 @@ func NewRunner(appContext types.AppContext, writer io.Writer, homeDir, dockerCom
 		HomeDir:                  homeDir,
 		ServiceConfigs:           serviceConfigs,
 		BuiltDependencies:        allBuiltDependencies,
-		DockerComposeDir:         path.Join(appContext.Location, "tmp"),
+		DockerComposeDir:         path.Join(appContext.Location, "docker-compose"),
+		DockerComposeFileName:    buildMode.GetDockerComposeFileName(),
```
charlierudolph: I don't thin k we need to save this on the runner and it can be computed in `Run`

src/docker/compose/index.go
```diff
@@ -4,34 +4,59 @@ import (
 	"github.com/Originate/exosphere/src/util"
 )
 
+var filename = "docker-compose.yml" //TODO inline opts.DockerComposeFileName once refactor is complete
+
 // BuildImage builds the given docker image
 func BuildImage(opts ImageOptions) error {
-	return util.RunAndPipe(opts.DockerComposeDir, opts.Env, opts.Writer, "docker-compose", "build", opts.ImageName)
+	if opts.DockerComposeFileName != "" {
+		filename = opts.DockerComposeFileName
```
charlierudolph: I don't like this as the global is updated. Thought on creating a getFilename function which takes in opts.DockerComposeFileName and return it if defined or the default value.

hugoesthere: This has been removed with the latest PR: https://github.com/Originate/exosphere/pull/731

It was a temporary fix so that I could work on one command at a time.

src/util/os_helpers.go
```diff
@@ -17,6 +17,18 @@ func CreateEmptyDirectory(dir string) error {
...
+// MakeDirectory creates a directory dir if it doesn't already exist, returns an error if any
+func MakeDirectory(dir string) error {
+	exists, err := DoesDirectoryExist(dir)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return os.MkdirAll(dir, os.FileMode(0777))
+	}
+	return nil
```
charlierudolph: You don't need to check if the directory exists as mkdirAll does nothing in that case. https://golang.org/pkg/os/#MkdirAll

src/util/os_helpers.go
```diff
@@ -17,6 +17,18 @@ func CreateEmptyDirectory(dir string) error {
 	return os.MkdirAll(dir, os.FileMode(0777))
 }
 
+// MakeDirectory creates a directory dir if it doesn't already exist, returns an error if any
+func MakeDirectory(dir string) error {
```
charlierudolph: Please make use of this function elsewhere to dry up the other uses of `os.MkdirAll`

charlierudolph (approved):

PR #731, opened 2017-11-03
Generate exo-deploy docker-compose in tmp dir

hugoesthere:
Partially resolves: https://github.com/Originate/exosphere/issues/697

### DIFF
--- a/src/application/deployer/index.go
+++ b/src/application/deployer/index.go
@@ -2,7 +2,7 @@ package deployer
 
 import (
 	"fmt"
-	"path"
+	"io/ioutil"
 
 	"github.com/Originate/exosphere/src/aws"
 	"github.com/Originate/exosphere/src/docker/composebuilder"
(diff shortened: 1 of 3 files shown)


src/application/deployer/index.go
```diff
@@ -67,6 +51,28 @@ func StartDeploy(deployConfig types.DeployConfig) error {
 	return deployApplication(deployConfig, imagesMap, prevTerraformFileContents)
 }
 
+func pushApplicationImages(deployConfig types.DeployConfig) (map[string]string, error) {
```
hugoesthere: Open to a different name here since we already have a `PushApplicationImages`

charlierudolph: Can we break "writeDockerCoposeFile" into its own function which returns the dockercompse dir? Thats most of the logic in this function

charlierudolph (approved):

Issue #726, opened 2017-11-01
exo run: port assignment


charlierudolph:
- The port under "production" replaces the "public-port"
- the docker section in service.yml goes away

This is part of the build up to #685

Use [cr-ports](https://github.com/Originate/exosphere/compare/master...cr-ports) as a starting point

PR #734, opened 2017-11-07
exo run port assignment

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/726


src/types/service_development_config.go
```diff
@@ -4,4 +4,5 @@ package types
 type ServiceDevelopmentConfig struct {
 	Dependencies []DevelopmentDependencyConfig
 	Scripts      map[string]string `yaml:",omitempty"`
+	Ports        []string          `yaml:",omitempty"`
```
charlierudolph: We want to keep this as a single port for now. We will be getting to multiple ports soon enough.

hugoesthere: But I think we get this for free, no? All we're doing is creating another entry in the `ports` block of the yml file which docker-compose handles for us

charlierudolph: For multiple ports @alexdavid and I had talked about:
```yml
ports: 
  key1: 3000
  key2: 4000
```
where the keys will be used in the env variables we pass to other services. Thus wanted to address this after we have service env variables

hugoesthere: Ah, ok. How are we managing env vars for ports right now? Is it ok to just have `port: 80` in service.yml?

charlierudolph: We aren't doing the env vars for ports yet (that will be the next task). Yes, for now please update this so you would add `port: 80` (separate ports under dev / production)

charlierudolph (approved):
Merge on green

Issue #735, opened 2017-11-07
exo test: pull / build phases


charlierudolph:
Currently when testing any service, it pulls/builds for the entire test file.

When testing a service, we pull/build only the service (does that auto pull/build its deps?)

When testing a application we could optimize to pull/build everything at the beginning and not while testing each service

@hugobho saw this when trying out the latest with EDX

hugoesthere:
> Currently when testing any service, it pulls/builds for the entire test file.

You're saying before or after the release of `0.29.0`? After `0.29.0` it should only be pulling/building per service if we're testing from the service level

> When testing a service, we pull/build only the service (does that auto pull/build its deps?)

Yes, it pulls/builds any service listed in `depends_on`.

> When testing a application we could optimize to pull/build everything at the beginning and not while testing each service

I'm not sure this provides much of an advantage. When you're testing a whole application, the total runtime will be the same no matter when we build everything. Also, if the user wants to interrupt tests when one fails, it makes a lot more sense to build before each service instead of all at once at the beginning where they risk waiting for all services to build only to have a test fail early on.

charlierudolph:
After `0.29.0` it does a pull and build on the entire `test.yml` file without passing in specific images to build / pull.

I was thinking of the optimization for the shared deps. Don't think it would say that much time anyway though

hugoesthere:
Ah yes, we don't specify service name for the build/pull step. I'll open a PR for that now

PR #736, opened 2017-11-07
exo test: build specific service & compose refactor

hugoesthere:
Resolves: https://github.com/Originate/exosphere/issues/735#issuecomment-342558698


charlierudolph:
Can we also do this for `Run` in `src/docker/composerunner/run.go`?

hugoesthere:
`Run` is used for `exo run`, and I think it actually makes more sense to separate the build&pull steps there, since those commands (at least the build) will only run if the image has change, whereas the `--build` flag will run build every time. For running a single service that's fine but when managing a whole application that's probably an unnecessary slowdown.

charlierudolph:
I would expect `docker build` then `docker up` to be the same as `docker up --build`. Is that not true? Does `up --build` ignore the caches?

hugoesthere:
Oh you're right, I thought `docker-compose build` was smart enough to skip if the image was built, like it does in `docker-compose up`. Looks like both cases uses the cache.

charlierudolph (approved):

Issue #685, opened 2017-10-13
Service endpoint environment variables


alexdavid:
I think for web services we should expose env variables `SERVICE_[service-name]_ORIGIN` and `SERVICE_[service-name]_EXTERNAL_ORIGIN`

and worker services should expose `SERVICE_[service-name]_HOST`

Origin would include protocol (http/https) and external is for access to the service not within the network (port forwarding locally in docker & fqdn in production)

For example, suppose we have a web service named "frontend" and a worker service named "ftp-api":
* Frontend service runs on container port `8080` in development mode and `80` in production mode. Exosphere locally binds host port `3000` for this service
* ftp-api service runs on container port `2121` in development and `21` in production mode. Exosphere locally binds host port `3001` for this service and ECS binds host port `65000` for this service

We would inject the following variables into each service:
### `SERVICE_FRONTEND_ORIGIN`
<table>
<tr><th>Local development value</th><td>http://frontend:8080</td></tr>
<tr><th>Local production value</th><td>http://frontend:80</td></tr>
<tr><th>AWS production value</th><td>http://frontend.local</td></tr>
</table>
Useful if one service needs to make an http call to another service. Routes to the ALB in production we will need to enable HTTP alb exposed to only the cluster.

### `SERVICE_FRONTEND_EXTERNAL_ORIGIN`
<table>
<tr><th>Local development value</th><td>http://localhost:3000</td></tr>
<tr><th>Local production value</th><td>http://localhost:3000</td></tr>
<tr><th>AWS production value</th><td>https://frontend.myapp.com</td></tr>
</table>
Useful for services that need to tell programs running outside the network (frontend javascript) where to access other publicly exposed services. Note that we automatically route to the https origin in production.

### `SERVICE_FTP_API_HOST`
<table>
<tr><th>Local development value</th><td>ftp-api:2121</td></tr>
<tr><th>Local production value</th><td>ftp-api:21</td></tr>
<tr><th>AWS production value</th><td>ftp-api.local:65000</td></tr>
</table>
Useful for services that need to communicate directly with non-web services. note that for worker services we need to have them query the ec2metadata api and update their route53 record to match its scheduled EC2 box.

Thoughts @chrizkim @charlierudolph @kevgo @hugobho ?

hugoesthere:
Looks good to me 👍

kevgo:
[...]

PR #737, opened 2017-11-07
service environment variables

charlierudolph:
@alexdavid @hugobho please review to check if this captures a use of `SERVICE_<NAME>_EXTERNAL_ORIGIN` from #685. It is a backend and frontend service which are both public and letting the frontend know where the backend is.

@hugobho please take over this once the ports are done


example-apps/service-external-origin-env/frontend-service/Dockerfile.prod
```diff
@@ -0,0 +1,15 @@
...
+COPY package.json .
+COPY yarn.lock .
+RUN yarn
+
+COPY . .
+RUN node_modules/.bin/webpack
+
+FROM nginx AS prod
+
+COPY --from=build /workspace/dist /usr/share/nginx/html
```
hugoesthere: Are we sure this Dockerfile works?

features/run/service_endpoint_environment_variables.feature
```diff
@@ -0,0 +1,29 @@
...
+    Given I am in the root directory of the "service-exteral-origin-env" example application
+    And starting "exo run" in my application directory
+    And it prints "Listening on port 3000" in the terminal
+    And it prints "webpack: Compiled successfully" in the terminal
+    Then http://localhost:3010 displays:
+      """
+      Backend located at http://localhost:3000
+      """
+
+  Scenario: external origin env var available at build time
```
hugoesthere: Why are these scenarios labeled "at run time" and "at build time"? Isn't the distinction development vs. production?

example-apps/service-external-origin-env/frontend-service/Dockerfile.prod
charlierudolph: Yes, I tested it. The only thing not hooked up is the env variable

features/run/service_endpoint_environment_variables.feature
charlierudolph: "at build time" is when building the docker file. "at run time" is when running the docker container. This is a special case where development needs it at run time and production needs it at build time. build time probably won't be used that much aside from frontend services.

example-apps/service-external-origin-env/backend-service/index.js
```diff
@@ -0,0 +1,6 @@
+const express = require('express')
+const app = express()
+
+app.get('/', (req, res) => res.send('Hello World!'))
+
+app.listen(3000, () => console.log('Listening on port 3000'))
```
alexdavid: To help speed up tests, do you think it would be worthwhile to remove the dependency on express for such simple apps?

Maybe something like
```javascript
require('http')
  .createServer((req, res) => res.end('Hello world!'))
  .listen(3000, () => console.log('Listening on port 3000'))
```

example-apps/service-external-origin-env/backend-service/service.yml
```diff
@@ -0,0 +1,12 @@
+type: backend-service
+description:
+author:
+
+development:
+  ports:
+    - 3000
```
alexdavid: I thought we were going to keep development and production the same, can production handle `ports` yet? I'd suggest leaving this as `port` until we need support for multiple ports

features/run/service_endpoint_environment_variables.feature
```diff
@@ -0,0 +1,29 @@
...
+
+  Scenario: external origin env var available at run time
+    Given I am in the root directory of the "service-exteral-origin-env" example application
+    And starting "exo run" in my application directory
+    And it prints "Listening on port 3000" in the terminal
+    And it prints "webpack: Compiled successfully" in the terminal
+    Then http://localhost:3010 displays:
+      """
+      Backend located at http://localhost:3000
+      """
```
alexdavid, line 20: I'm starting to think that having to manually write up an example app per feature file may start to get unmaintainable quickly.

Do you think it would be possible to have feature files that self-describe their appllications on a higher level and we can have step definitions scaffold them out in a temp directory?

Maybe something like
```gherkin
Scenario: external origin env var available at run time
  Given I have an application with the following services:
    | NAME     | TYPE |
    | frontend | web  |
    | api      | web  |
  When running the web service
  Then I have the following environment variables available at run time:
    | NAME                             | VALUE                 |
    | SERVICE_FRONTEND_EXTERNAL_ORIGIN | http://localhost:3000 |
    | SERVICE_API_EXTERNAL_ORIGIN      | http://localhost:3010 |
```

example-apps/service-external-origin-env/backend-service/service.yml
alexdavid: Also I think we should make this `8080` or some other port to make it clear that this has nothing to do with the host port of 3000 that gets assigned below

alexdavid (commented):
Overall looks good, but I think we should consider using more high level step definitions to really demonstrate what we're testing. Otherwise cucumber just gets in the way here more than it actually helps

features/run/service_endpoint_environment_variables.feature
charlierudolph: I do like that your steps get to the heart of the user value but i ave no idea what we would put as step definitions in order to test it.

I think the example apps are nice as it shows a full usage and provides a nice test bed that you can run in by cd-ing into it. I don't know if what you propose will be flexible enough.

example-apps/service-external-origin-env/backend-service/service.yml
charlierudolph: Was working with an old version of a branch. This should be `port: 3000` now

features/run/service_endpoint_environment_variables.feature
alexdavid: I do like that the example apps are there for real use-cases and I don't think we should remove them fully, but we should have a small manageable amount of them showing large scale features (like communicating with exocom perhaps). But there are a lot of smaller more edge-casey features that I think should be more like this style.

As for implementation, our scaffolded services could prepare all the possible information that we might care about and expose them through a web interface or something. For example, we could have a template example app that get's scaffolded for each service under `Given I have an application with the following services:`.

That service can expose whatever data we need through http, for example, `http://host/build-time-env-vars` could return the environment variables that were availalbe at build time.
This can be done with something like
```dockerfile
RUN env > /usr/app/public/build-time-env-vars
```

and we could do something similar for run-time env vars `http://host/run-time-env-vars`

What are your thoughts?

hugoesthere:
@charlierudolph @alexdavid good for a look, let me know if this is what you guys had in mind.

src/docker/composebuilder/service_endpoints.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return []string{fmt.Sprintf("%s:%s", s.HostPort, s.ContainerPort)}
+}
+
+// GetEndpointMappings returns a map from env var name to env var value of a service endpoint
+func (s *ServiceEndpoints) GetEndpointMappings() map[string]string {
+	if s.ContainerPort == "" {
+		return map[string]string{}
+	}
+	externalKey := fmt.Sprintf("SERVICE_%s_EXTERNAL_ORIGIN", strings.ToUpper(s.ServiceRole))
```
charlierudolph: This should only be for public services. Can we add some unit tests for this class as well?

src/docker/composebuilder/service_endpoints.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// GetEndpointMappings returns a map from env var name to env var value of a service endpoint
+func (s *ServiceEndpoints) GetEndpointMappings() map[string]string {
+	if s.ContainerPort == "" {
+		return map[string]string{}
+	}
+	switch s.ServiceConfig.Type {
+	case "public":
+		externalKey := fmt.Sprintf("SERVICE_%s_EXTERNAL_ORIGIN", strings.ToUpper(s.ServiceRole))
```
charlierudolph: Had discussed with @alexdavid about removing the `SERVICE_` prefix. What do you think?

charlierudolph (commented):
Looks good to me. Are we going add the terraform variables in another PR?

src/docker/composebuilder/service_endpoints.go
hugoesthere: I'm fine with that

hugoesthere:
Yeah terraform vars will be coming in another PR

PR #747, opened 2017-11-08
Fix deploy bug: point docker-compose to filename

hugoesthere:


hugoesthere:
Closing for now, will re-open as bigger PR building images from `run_production.yml` file

PR #754, opened 2017-11-09
simplify dependency logic

charlierudolph:
remove the weird logic tying application dependencies to service dependencies. Any application dependency is available to all services anyway.

@alexdavid was running into an issue with this when he let the config object empty. The isEmpty functions weren't used anywhere else.

### DIFF
--- a/src/config/generic_development_dependency.go
+++ b/src/config/generic_development_dependency.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/Originate/exosphere/src/docker/tools"
 	"github.com/Originate/exosphere/src/types"
@@ -38,5 +39,10 @@ func (g *genericDevelopmentDependency) GetDockerConfig() (types.DockerConfig, er
 // GetServiceEnvVariables returns the environment variables that need to
 // be passed to services that use it
 func (g *genericDevelopmentDependency) GetServiceEnvVariables() map[string]string {
-	return g.config.Config.ServiceEnvironment
+	result := map[string]string{}
+	result[strings.ToUpper(g.config.Name)] = g.GetContainerName()
+	for key, value := range g.config.Config.ServiceEnvironment {
+		result[key] = value
+	}
+	return result
 }
--- a/src/config/generic_development_dependency.go
+++ b/src/config/generic_development_dependency.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/Originate/exosphere/src/docker/tools"
 	"github.com/Originate/exosphere/src/types"
+	"github.com/Originate/exosphere/src/util"
 )
 
 type genericDevelopmentDependency struct {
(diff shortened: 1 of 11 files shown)


src/docker/composebuilder/development.go
```diff
@@ -169,9 +168,6 @@ func (d *DevelopmentDockerComposeBuilder) getDockerEnvVars() map[string]string {
 			result[variable] = value
 		}
 	}
-	for _, dependency := range d.ServiceConfig.Development.Dependencies {
-		result[strings.ToUpper(dependency.Name)] = dependency.Name
-	}
```
charlierudolph: This should be injected in a similar way to service environment endpoints.

src/docker/composebuilder/development.go
```diff
@@ -169,9 +159,6 @@ func (d *DevelopmentDockerComposeBuilder) getDockerEnvVars() map[string]string {
...
```
charlierudolph, line 174: This was moved into the generic development dependency struct

src/docker/composebuilder/development.go
```diff
@@ -149,14 +147,6 @@ func (d *DevelopmentDockerComposeBuilder) getServiceFilePath() string {
 	return path.Join(d.AppDir, d.ServiceData.Location)
 }
 
-func (d *DevelopmentDockerComposeBuilder) getDockerLinks() []string {
-	result := []string{}
-	for _, dependency := range d.ServiceConfig.Development.Dependencies {
-		result = append(result, fmt.Sprintf("%s%s:%s", dependency.Name, dependency.Version, dependency.Name))
-	}
-	return result
```
charlierudolph, line 157: I don't believe this is needed if we switch from put the container name as the env var value instead of the dependency name

src/config/generic_development_dependency.go
```diff
@@ -38,5 +39,10 @@ func (g *genericDevelopmentDependency) GetDockerConfig() (types.DockerConfig, er
 // GetServiceEnvVariables returns the environment variables that need to
 // be passed to services that use it
 func (g *genericDevelopmentDependency) GetServiceEnvVariables() map[string]string {
-	return g.config.Config.ServiceEnvironment
+	result := map[string]string{}
+	result[strings.ToUpper(g.config.Name)] = g.GetContainerName()
+	for key, value := range g.config.Config.ServiceEnvironment {
+		result[key] = value
+	}
+	return result
```
hugoesthere: Can this be reversed to:
```
result := g.config.Config.ServiceEnvironment
result[strings.ToUpper(g.config.Name)] = g.GetContainerName()
return result
```
?

hugoesthere (commented):
So the expected behavior here should be that dependencies are either declared on the app level, in which case all services can use it, or at the service level, in which case only that service can use it? And it's not our responsibility to provide any checks against someone using dependencies in a different way than that.

I think this would simplify a lot of the dependency logic we have.

src/config/generic_development_dependency.go
charlierudolph: That would edit the ServiceEnviorment object which i would prefer not to do. Believe I can use `util.Merge` though.

charlierudolph:
> So the expected behavior here should be that dependencies are either declared on the app level, in which case all services can use it, or at the service level, in which case only that service can use it? 

Correct. Thats basically what we were doing anyway. The idea of listing some deps again in the service never made sense to me.

> And it's not our responsibility to provide any checks against someone using dependencies in a different way than that.

We never had any checks like that as far as i can see. I think we have some more dependency changes coming but we should eventually add a check that a service dep won't override an app dep in the docker-compose (if they have the same name).

hugoesthere (approved):

PR #762, opened 2017-11-10
move exosphere data path

charlierudolph:
There goes all the home dir.


src/docker/tools/index.go
```diff
@@ -61,8 +61,8 @@ func GetExitCode(containerName string) (int, error) {
 }
 
 // GetRenderedVolumes returns the rendered paths to the given volumes
-func GetRenderedVolumes(volumes []string, appName string, role string, homeDir string) ([]string, error) {
-	dataPath := path.Join(homeDir, ".exosphere", appName, role, "data")
+func GetRenderedVolumes(volumes []string, appName string, role string) ([]string, error) {
+	dataPath := path.Join("${APP_PATH}", ".exosphere", "data", role)
```
charlierudolph: This is the primary change.

src/docker/tools/index.go
```diff
@@ -61,12 +61,9 @@ func GetExitCode(containerName string) (int, error) {
...
 	renderedVolumes := []string{}
-	if err := util.MakeDirectory(dataPath); err != nil { //nolint gas
```
hugoesthere: We'll still have to create the dir at whatever `${APP_PATH}/.exosphere/data/role` compiles to right?

hugoesthere (commented):
We still have to set `APP_PATH` when running `docker-compose`: https://docs.docker.com/compose/compose-file/#variable-substitution

We can either do that for the user (i.e. in `docker/compose/index.go` commands) or we require the user to set it themselves. In which case we should probably name it something less generic than `APP_PATH`

src/docker/tools/index.go
charlierudolph: True, I'll add that back

charlierudolph:
`APP_PATH` is already set. Added in https://github.com/Originate/exosphere/pull/750

src/docker/tools/index.go
```diff
@@ -61,14 +61,16 @@ func GetExitCode(containerName string) (int, error) {
 }
 
 // GetRenderedVolumes returns the rendered paths to the given volumes
-func GetRenderedVolumes(volumes []string, appName string, role string, homeDir string) ([]string, error) {
-	dataPath := path.Join(homeDir, ".exosphere", appName, role, "data")
+func GetRenderedVolumes(volumes []string, appDir string, role string) ([]string, error) {
+	appRelativePath := path.Join(".exosphere", "data", role)
+	realPath := path.Join(appDir, appRelativePath)
+	renderedPath := path.Join("${APP_PATH}", appRelativePath)
```
hugoesthere: If `appDir` is already something like `Users/charlie/exosphere/space-tweet`, and `${APP_PATH}` is also `Users/charlie/exosphere/space-tweet`, won't this line combine them into `Users/charlie/exosphere/space-tweet/Users/charlie/exosphere/space-tweet`? That seemed to be the behavior when I tested it out here: https://golang.org/pkg/path/#Join

test_helpers/index.go
```diff
@@ -73,25 +73,22 @@ func createEmptyApp(appName, cwd string) (string, error) {
...
 	if err != nil {
-		return errors.Wrap(err, "Failed to kill test containers")
+		return err
 	}
-	testComposeProjectName := composebuilder.GetTestDockerComposeProjectName(appName)
-	err = application.CleanServiceTestContainers(appDir, testComposeProjectName, writer)
-	if err != nil {
-		return errors.Wrap(err, "Failed to kill test containers")
+	if doesExist {
+		cmdPlus := execplus.NewCmdPlus("exo", "clean") // nolint gas
```
hugoesthere: `killTestContainers` was used for creating dangling images/volumes, I think it was more specific than `exo clean`. Are the tests still doing that?

src/docker/tools/index.go
charlierudolph: This is joining `${APP_PATH}` with `path.Join(".exosphere", "data", role)`.

charlierudolph: I'll rename `appRelativePath` to `dataRelativePath`

test_helpers/index.go
charlierudolph: Auhh. I will revert the changes here and add a new function to clean up at the end of a scenario (which was the other place using this)

charlierudolph:
@hugobho comments addressed

hugoesthere (approved):

PR #767, opened 2017-11-11
move test application / templates

charlierudolph:
* `example-apps` -> `test/fixtures/applications`
* `example-templates` -> `test/fixtures/service_templates`
* `test_helpers` -> `test/helpers`

Added a helper `GetTestApplicationDir` to clean up some unit tests

On titan, Maxime mentioned [gomvpkg](https://godoc.org/golang.org/x/tools/cmd/gomvpkg) which I used here


hugoesthere (approved):

alexdavid (approved):
Awesome, I think this is a much better place for these

PR #768, opened 2017-11-13
Pass filename to deploy processes

hugoesthere:
Fixes no filename bug


src/application/deployer/push_application_images.go
```diff
@@ -37,6 +38,7 @@ func PushApplicationImages(deployConfig types.DeployConfig, dockerComposeDir str
 			ImageName:        imageName,
 			ServiceRole:      serviceRole,
 			ServiceLocation:  serviceData[serviceRole].Location,
+			BuildMode:        buildMode,
```
charlierudolph, line 41: line 23 in this file should also be updated to get the build mode filename

charlierudolph (approved):

PR #769, opened 2017-11-13
exo-generate

### LINKED ISSUE #749: new command: exo generate
Generate all the docker-compose and terraform files. 

`run`, `test` and `deploy` all do this at the beginning of there command.

Pair with `exo needs-generate`which checks if the docker-compose and terraform files are up to date.

@hugobho what do you think? @alexdavid and i were discussing this.

hugoesthere:
partially resolves https://github.com/Originate/exosphere/issues/749

Not sure what the best way to test this is, but seeing as the crux of this command is `GetApplicationDockerConfigs`, I think we have most of our test cases covered.


charlierudolph:
This should also generate the terraform file. Lets also add a feature test that verifies the files are added (without testing the content)

hugoesthere:

charlierudolph:
Should the base branch be updated to `hh-terraform-vars` since it requires #775?

hugoesthere:
Yes, on my machine that's the case I just didn't update it here

bin/pre-release
```diff
@@ -3,6 +3,6 @@
 hash=$(git log --format="%h" -n 1 terraform)
 
 if [ -z "$(grep "$hash" src/cmd/deploy.go)" ]; then
```
charlierudolph: this also needs to be updated to shared

features/generate.feature
```diff
@@ -0,0 +1,17 @@
...
+  Scenario: generate docker-compose and terraform files
+    Given I am in the root directory of the "simple" example application
+    When starting "exo generate" in my application directory
+    Then it prints "docker-compose and terraform files generated" in the terminal
+    And my workspace contains the files:
+      | docker-compose/run_development.yml |
+      | docker-compose/run_production.yml  |
+      | docker-compose/test.yml            |
+      | terraform/main.tf                  |
+
```
charlierudolph: Please remove this extra empty line

src/cmd/generate.go
```diff
@@ -0,0 +1,42 @@
...
+		context, err := GetContext()
+		if err != nil {
+			log.Fatal(err)
+		}
+		err = application.GenerateComposeFiles(context.AppContext)
+		if err != nil {
+			log.Fatal(err)
+		}
+		deployConfig := getBaseDeployConfig(context.AppContext)
+		writer := os.Stdout
```
charlierudolph: is this variable needed? Its only used on the line below

charlierudolph (approved):

src/cmd/generate.go
```diff
@@ -0,0 +1,39 @@
...
+		err = application.GenerateComposeFiles(context.AppContext)
+		if err != nil {
+			log.Fatal(err)
+		}
+		deployConfig := getBaseDeployConfig(context.AppContext)
+		err = application.GenerateTerraformFiles(deployConfig)
+		if err != nil {
+			log.Fatal(err)
+		}
+		fmt.Println("docker-compose and terraform files generated")
```
charlierudolph: I don't think we need either of these messages (generating or generated)

hugoesthere: We don't, I really only have them there so the tests have some text to wait on

charlierudolph: We can wait for the process to exit.

PR #777, opened 2017-11-14
Switch exo run, clean and test to use exo generate

### LINKED ISSUE #749: new command: exo generate

hugoesthere:
Partially resolves https://github.com/Originate/exosphere/issues/749


src/application/clean.go
```diff
@@ -7,26 +7,29 @@ import (
...
+func CleanContainers(appContext types.AppContext, writer io.Writer) error {
+	err := GenerateComposeFiles(appContext)
+	if err != nil {
+		return err
+	}
+	composeProjectName := composebuilder.GetDockerComposeProjectName(appContext.Config.Name)
+	for _, dockerComposeFileName := range composebuilder.GetComposeFileNames() {
+		if dockerComposeFileName == composebuilder.LocalTestComposeFileName {
+			composeProjectName = composebuilder.GetTestDockerComposeProjectName(appContext.Config.Name)
+		}
```
charlierudolph: I don't like changing the variable in the loop as it effects the following loop iterations. I think composeProjectName should be recomputed on each loop iteration.

charlierudolph (approved):

src/application/tester/test_runner.go
```diff
@@ -52,25 +57,12 @@ func (s *TestRunner) Shutdown() error {
...
 		Writer:      s.Writer,
 		AbortOnExit: true,
 	}, nil
 }
-
-func (s *TestRunner) getDockerComposeConfigs() (types.DockerConfigs, error) {
-	return composebuilder.GetApplicationDockerConfigs(composebuilder.ApplicationOptions{
-		AppConfig: s.AppContext.Config,
-		AppDir:    s.AppContext.Location,
-		BuildMode: s.BuildMode,
```
charlierudolph, line 74: This build mode appears to have been lost

src/application/runner.go
```diff
@@ -43,17 +43,12 @@ func NewRunner(appContext types.AppContext, writer io.Writer, dockerComposeProje
 
 // Run runs the application with graceful shutdown
 func (r *Runner) Run() error {
-	dockerConfigs, err := composebuilder.GetApplicationDockerConfigs(composebuilder.ApplicationOptions{
-		AppConfig: r.AppContext.Config,
-		AppDir:    r.AppContext.Location,
-		BuildMode: r.BuildMode,
```
charlierudolph, line 49: same here

src/application/tester/test_runner.go
hugoesthere: Build mode is lost, but that's because `GenerateComposeFiles` is building for all build modes (and all docker-compose.yml files)

charlierudolph (approved):

PR #784, opened 2017-11-15
use named docker volumes

### LINKED ISSUE #763: use named docker volumes
instead of pointing to somewhere on the user's machine. May require some rework of the compose builder, as named volumes need to appear in the docker compose file in a separate volumes block.

charlierudolph:
resolves #763


src/docker/composebuilder/application.go
```diff
@@ -40,15 +40,18 @@ func GetDependenciesDockerConfigs(options ApplicationOptions) (types.DockerConfi
 			if err != nil {
 				return result, err
 			}
-			result[builtDependency.GetContainerName()] = dockerConfig
+			result.Services[builtDependency.GetContainerName()] = dockerConfig
```
hugoesthere, line 43: Unrelated to this PR, but do you think we should start making container name and dependency name synonymous? I think in the past the only reason we made dependency names a concatenation of `#{dependencyName}#{dependencyVersion}` was so that we could support multiple versions of the same dependency. We're not using that now and it's just proving to be a point of confusion

hugoesthere (commented):
I think we have to think about the use case where someone has two exosphere apps on their machine and both use a dependency with the same named volume. That obviously wouldn't work out, we might have to do something like append an application name to the front of a volume name. Unless docker-compose automatically manages that and prepends the project name?

Other than that I really like this change since it lets us use `docker volume inspect` and stuff like that

charlierudolph:
Through testing I see it automatically prepends the compose project name. For example on EDX, if its named `postgres` in the docker compose yml file, while the actual volume created is `edx_postgres`

hugoesthere (approved):

charlierudolph:
@hugobho please take another look as was an additional change to the compose builder to make this work. @alexdavid please review the API here.

src/docker/composebuilder/production.go
```diff
@@ -27,49 +27,49 @@ func NewProductionDockerComposeBuilder(appConfig types.AppConfig, serviceConfig
...
 // getServiceDockerConfigs returns a DockerConfig object for a single service and its dependencies (if any(
-func (d *ProductionDockerComposeBuilder) getServiceDockerConfigs() (types.DockerConfigs, error) {
+func (d *ProductionDockerComposeBuilder) getServiceDockerCompose() (*types.DockerCompose, error) {
 	if d.ServiceData.Location != "" {
-		return d.getInternalServiceDockerConfigs()
+		return d.getInternalServiceDockerCompose()
 	}
 	if d.ServiceData.DockerImage != "" {
-		return d.getExternalServiceDockerConfigs()
+		return d.getInternalServiceDockerCompose()
```
hugoesthere: This should be calling `getExternalServiceDockerCompose`

hugoesthere (approved):

features/tutorial.feature
```diff
@@ -134,8 +134,8 @@ Feature: Following the tutorial
           - name: 'mongo'
             version: '3.4.0'
             config:
-              volumes:
-                - '{{EXO_DATA_PATH}}:/data/db'
+              named-volumes:
+                mongo: '/data/db'
```
alexdavid: So I though we were originally going to be able to do this with `volumes_from` and not bother the user to define these in their service.yml. But it looks like docker-compose deprecated this functionality. So if we're going to make the users manually define volumes, we might as well just name this `volumes` and pass that straight through to docker-compose and use the standard docker syntax:
```yaml
volumes:
  - 'mongo:/data/db'
```

src/config/app_development_dependency.go
```diff
@@ -7,6 +7,7 @@ type AppDevelopmentDependency interface {
 	GetContainerName() string
 	GetDockerConfig() (types.DockerConfig, error)
 	GetServiceEnvVariables() map[string]string
+	GetVolumeNames() []string
```
alexdavid, line 10: If we're making users manually add this to the service.yml, why are dependencies exporting this.

How will this work when dependencies no longer have code?

features/tutorial.feature
charlierudolph: In docker-compose, named docker volumes must also be listed in a separate volumes section at the bottom. We could parse the volumes the user passes in but I want to avoid having to write the code to determine if the volume is a name or a file path.

src/config/app_development_dependency.go
charlierudolph: This is currently where are parsing the the dependency config. This would probably be moved into the compose builder when deps go away

features/tutorial.feature
alexdavid: Ugh that's obnoxious

alexdavid: Can we at that point just use procedurally generated names for these and just call it `persist` or something?
```yaml
persist:
  - /data/db
```

charlierudolph: sounds good.

charlierudolph: Updated with the volume being named `<DEPENDENCY_NAME>_<PATH>` with the with the whole thing converted to snake case

charlierudolph:
@alexdavid comments addressed

features/tutorial.feature
```diff
@@ -134,8 +133,8 @@ Feature: Following the tutorial
           - name: 'mongo'
             version: '3.4.0'
             config:
-              volumes:
-                - '{{EXO_DATA_PATH}}:/data/db'
+              persist:
+                - '/data/db'
```
alexdavid: I don't think we need the quotes here anymore

src/config/generic_development_dependency.go
```diff
@@ -22,15 +21,16 @@ func (g *genericDevelopmentDependency) GetContainerName() string {
 
 // GetDockerConfig returns docker configuration and an error if any
 func (g *genericDevelopmentDependency) GetDockerConfig() (types.DockerConfig, error) {
-	renderedVolumes, err := tools.GetRenderedVolumes(g.config.Config.Volumes, g.appDir, g.config.Name)
-	if err != nil {
-		return types.DockerConfig{}, err
+	volumes := []string{}
+	for _, path := range g.config.Config.Persist {
+		name := util.ToSnake(g.config.Name + "_" + path)
```
alexdavid, line 26: Do you think it's worthwhile to include the application name in here so it's obvious which application the volume belongs to when I run `docker volume ls`?

alexdavid (approved):
Looks great

src/config/generic_development_dependency.go
charlierudolph: docker-compse will prefix the volume with the compose project name which is the app name in our case

PR #790, opened 2017-11-21
Update the Get Started documentation

brian-dlee:
As I get myself introduced to Exosphere, I'm updating the documentation to reflect the current implementation of Exosphere as it has evolved over the past few months.

@hugobho

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.vscode
 bindata.go
 dist/
 /tmp/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.vscode
 bindata.go
 dist/
 /tmp/
--- a/website/tutorial/part_1/03_installation.md
+++ b/website/tutorial/part_1/03_installation.md
@@ -19,8 +19,8 @@ a __package management system__ for installing/updating the various pieces of pl
 ## Dependencies
 
 Before installing Exosphere, ensure you have the following applications installed:
- * Git - [https://git-scm.com/]()
- * Docker - [https://www.docker.com]()
+ * [Git](https://git-scm.com/)
+ * [Docker](https://www.docker.com)
 
 ## General installation instructions
 
--- a/website/tutorial/part_1/03_installation.md
+++ b/website/tutorial/part_1/03_installation.md
@@ -35,7 +35,7 @@ _Note: Depending on where you install these steps could require root privileges_
 ```bash
 curl -L https://github.com/Originate/exosphere/releases/download/vX.Y.Z/exo-darwin-amd64 >/usr/local/bin/exo-X.Y.Z
 chmod +x /usr/local/bin/exo-X.Y.Z
-ln -s /usr/local/bin/exo-X.Y.Z /usr/local/bin/exo
+ln -fns /usr/local/bin/exo-X.Y.Z /usr/local/bin/exo
 ```
 
 <table>
--- a/website/tutorial/part_1/03_installation.md
+++ b/website/tutorial/part_1/03_installation.md
@@ -19,7 +19,7 @@ a __package management system__ for installing/updating the various pieces of pl
 ## Dependencies
 
 Before installing Exosphere, ensure you have the following applications installed:
- * [Git](https://git-scm.com/)
+ * [Git](https://git-scm.com)
  * [Docker](https://www.docker.com)
 
 ## General installation instructions
--- a/website/tutorial/part_1/03_installation.md
+++ b/website/tutorial/part_1/03_installation.md
@@ -32,11 +32,12 @@ Before installing Exosphere, ensure you have the following applications installe
 
 _Note: Depending on the privileges of `/usr/local/bin` on your system these steps could require root privileges_
 
-In a terminal, execute the following:
-```bash
-curl -L https://github.com/Originate/exosphere/releases/download/vX.Y.Z/exo-darwin-amd64 >/usr/local/bin/exo
-chmod +x /usr/local/bin/exo
-```
+1. Download the Exosphere binary for your system from the release page referenced above (for most, this is the 64-bit Darwin build).
+2. Assuming our binary downloaded to `~/Downloads/exo-darwin-amd64`, 
+let's move the binary to a suitable location and make sure it's executable.
+   * `mv ~/Downloads/exo-darwin-amd64 /usr/local/bin/exo`
+   * `chmod +x /usr/local/bin/exo`
+
 
 <table>
   <tr>
--- a/website/tutorial/part_1/03_installation.md
+++ b/website/tutorial/part_1/03_installation.md
@@ -17,7 +17,7 @@ a __package management system__ for installing/updating the various pieces of pl
   frameworks, and other dependencies of your polyglot applications manually.
 
 * Download and install the binary executable for your platform from the 
-[GitHub release page](https://github.com/Originate/exosphere/releases)
+[GitHub release page](https://github.com/Originate/exosphere/releases/latest)
 * Rename the executable to `exo`
 * Check and update your `PATH` environmental variable to include its install location, if required
 
--- a/website/tutorial/part_1/03_installation.md
+++ b/website/tutorial/part_1/03_installation.md
@@ -11,71 +11,29 @@
 
 Exosphere is not just a simple application,
 but a framework for large-scale software development of micro-service based applications.
-Developing applications consisting of many different code bases requires:
-* a __package management system__ for installing/updating the various pieces of platforms and tools as needed.
+Developing applications consisting of many different code bases requires
+a __package management system__ for installing/updating the various pieces of platforms and tools as needed.
   You don't want to get into the business of having to install several programming languages,
   frameworks, and other dependencies of your polyglot applications manually.
 
-The installation instructions below include
-making these infrastructure components available on your system.
+* Download and install the binary executable for your platform from the 
+[GitHub release page](https://github.com/Originate/exosphere/releases)
+* Rename the executable to `exo`
+* Check and update your `PATH` environmental variable to include its install location, if required
 
-### macOS
+## Managing multiple versions of `exo`
 
-To install the SDK manually:
+A common pattern for managing executables on a system is to install the executable with
+the version in the filename and symlinking the desired command name to the executable.
+Using this pattern, old versions are still available in your `PATH`. In the event that you need
+to run an old version of exo it will still be available as `exo-X.Y.Z` unless it is explicitly 
+removed.
 
-* install [Homebrew](http://brew.sh)
-  * run `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
-  * then run `brew doctor` and fix all the errors
-* install [Node.js](https://nodejs.org) version 6 or above: `brew install node`
-* install Exosphere: `npm i -g exosphere-sdk`
-* verify that you can run Exosphere commands: `exo version`
+### Example: Installing version X.Y.Z of exo on Mac
+_Note: Depending on where you install these steps could require root privileges_
 
-
-## Windows
-
-### Windows 7
-
-Some of the commands below have to be run in an __administrative command shell__.
-Instructions for how to open such a shell are available
-for [windows 8-10](http://www.howtogeek.com/194041/how-to-open-the-command-prompt-as-administrator-in-windows-8.1)
-and [Windows 7](http://www.howtogeek.com/howto/windows-vista/run-a-command-as-administrator-from-the-windows-vista-run-box)).
-
-* install the package manager
-  * in an administrative shell, install [Chocolatey](https://chocolatey.org/install)
-  * when done, close the current shell and open a new one to load the environment changes prepared by the installer
-* install [Node.js](http://nodejs.org) version 6 or above
-  * in another administrative shell: `choco install nodejs.install -y`
-* install Exosphere:
-  * in a normal shell: `npm install --global exosphere-sdk`
-  * close this shell, open a normal one, and run `exo` to make sure it works
-* install other tools you will need
-  * in an administrative shell: [Git](https://git-scm.com) (`choco install git.install -y`
-  * optionally [Github Desktop](https://desktop.github.com) (install manually, the choco package is broken)
-
-
-### Contributing
-
-If you want to become an Exosphere contributor and you are on Windows,
-you need some additional infrastructure:
-* [Git](https://git-scm.com) for the Git command line and Git Bash: `choco install git.install -y`
-* you also need to set up SSH keys for Github.
-  The easiest way is via [Github Desktop](https://desktop.github.com): `choco install github -y`
-  (Note: this install is currently broken, you might want to install this manually)
-
-Please perform all
-
-
-
-## Linux
-
-Installation instructions given for [Ubuntu](http://www.ubuntu.com),
-please adapt them to your distro as needed:
-* install [ZeroMQ](http://zeromq.org)
-* install [Node.js](https://nodejs.org) version 6 or above: `sudo apt-get install node`
-
-
-<table>
-  <tr>
-    <td><a href="../part_2/readme.md"><b>&gt;&gt;</b></td>
-  </tr>
-</table>
+```bash
+curl -L https://github.com/Originate/exosphere/releases/download/vX.Y.Z/exo-darwin-amd64 >/usr/local/bin/exo-X.Y.Z
+chmod +x /usr/local/bin/exo-X.Y.Z
+ln -s /usr/local/bin/exo-X.Y.Z /usr/local/bin/exo
+```
--- a/website/tutorial/part_1/03_installation.md
+++ b/website/tutorial/part_1/03_installation.md
@@ -16,6 +16,14 @@ a __package management system__ for installing/updating the various pieces of pl
   You don't want to get into the business of having to install several programming languages,
   frameworks, and other dependencies of your polyglot applications manually.
 
+## Dependencies
+
+Before installing Exosphere, ensure you have the following applications installed:
+ * Git - [https://git-scm.com/]()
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+.vscode
```
kevgo: Tool-specific configuration should be in the [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) on your device, not in every code base you work on.

website/tutorial/part_1/03_installation.md
```diff
@@ -11,68 +11,32 @@
...
-Developing applications consisting of many different code bases requires:
-* a __package management system__ for installing/updating the various pieces of platforms and tools as needed.
+Developing applications consisting of many different code bases requires
+a __package management system__ for installing/updating the various pieces of platforms and tools as needed.
   You don't want to get into the business of having to install several programming languages,
   frameworks, and other dependencies of your polyglot applications manually.
 
-The installation instructions below include
-making these infrastructure components available on your system.
+* Download and install the binary executable for your platform from the
```
kevgo: Please link to the latest release: https://github.com/Originate/exosphere/releases/latest. This way people always install the latest version.

website/tutorial/part_1/03_installation.md
```diff
@@ -11,68 +11,32 @@
...
+a __package management system__ for installing/updating the various pieces of platforms and tools as needed.
   You don't want to get into the business of having to install several programming languages,
   frameworks, and other dependencies of your polyglot applications manually.
 
-The installation instructions below include
-making these infrastructure components available on your system.
+* Download and install the binary executable for your platform from the 
+[GitHub release page](https://github.com/Originate/exosphere/releases)
+* Rename the executable to `exo`
+* Check and update your `PATH` environmental variable to include its install location, if required
```
kevgo: I would suggest moving the `exo` binary into a directory that is in the path. This is common for installation instructions.

website/tutorial/part_1/03_installation.md
```diff
@@ -11,68 +11,32 @@
...
 
-### macOS
+## Managing multiple versions of Exosphere
 
-To install the SDK manually:
+A common pattern for managing executables on a system is to install the executable with
+the version in the filename and symlinking the desired command name to the executable.
+Using this pattern, old versions are still available in your `PATH`. In the event that you need
+to run an old version of Exosphere it will still be available as `exo-X.Y.Z` unless it is explicitly 
+removed.
```
kevgo: Is that really a thing right now? Is there a use case for this? Afaik we don't provide support for older versions since we don't have stable releases yet. I would prefer that everybody always uses the latest version. Once we reach a the second stable release (2.0), we can support older stable releases (1.x) longer.

website/tutorial/part_1/03_installation.md
```diff
@@ -11,68 +11,32 @@
...
+Developing applications consisting of many different code bases requires
+a __package management system__ for installing/updating the various pieces of platforms and tools as needed.
   You don't want to get into the business of having to install several programming languages,
   frameworks, and other dependencies of your polyglot applications manually.
 
-The installation instructions below include
-making these infrastructure components available on your system.
+* Download and install the binary executable for your platform from the 
+[GitHub release page](https://github.com/Originate/exosphere/releases)
+* Rename the executable to `exo`
```
kevgo: Please also include that people need a recent version of Git and Docker on their system.

kevgo (commented):
Thanks for helping with this, @brian-dlee! Some comments below.

website/tutorial/part_1/03_installation.md
brian-dlee: I've reworked the install instructions, but only have specific installation instructions for Mac. I'm not familiar enough with scripting on Windows to write Windows instruction.

Do you think it's safe to assume a Linux user is able to install without explicit instructions? Especially considering Linux instructions will be nearly identical to the Mac instructions.

kevgo: Don't worry about platforms you are not familiar with. Thanks for doing this for the Mac!

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -110,138 +115,161 @@ Here is the current architecture of our application:
...
-Running Todo application 0.0.1
-
- exocom      online at port 8000
-html-server  Server running at port 3000
- exorun      'html-server' is running using exorelay port 8001
- exorun      all services online
- exocom      received routing setup
- exorun      application ready
-```
+* **exocom** is the command that runs Exosphere applications. It starts the
```
brian-dlee: I could use a little assistance here. I'm not sure if replacing exorun with exocom in this text block is actually correct. Is exorun just not used here?

brian-dlee: I think I found an appropriate change to address this. Of course, any and all feedback is welcome.

website/tutorial/part_1/03_installation.md
```diff
@@ -11,68 +11,32 @@
...
   You don't want to get into the business of having to install several programming languages,
   frameworks, and other dependencies of your polyglot applications manually.
 
-The installation instructions below include
-making these infrastructure components available on your system.
+## Dependencies
 
-### macOS
+Before installing Exosphere, ensure you have the following applications installed:
+ * Git - [https://git-scm.com/]()
```
kevgo: I'd format the name of the tools as the link, like so:

```markdown
* [Git](https://git-scm.com)
* [Docker](https://www.docker.com)

File: website/tutorial/part_1/03_installation.md [hunk 8]
```diff
@@ -11,68 +11,32 @@
...
-
-Installation instructions given for [Ubuntu](http://www.ubuntu.com),
-please adapt them to your distro as needed:
-* install [ZeroMQ](http://zeromq.org)
-* install [Node.js](https://nodejs.org) version 6 or above: `sudo apt-get install node`
+_Note: Depending on the privileges of `/usr/local/bin` on your system these steps could require root privileges_
 
+In a terminal, execute the following:
+```bash
+curl -L https://github.com/Originate/exosphere/releases/download/vX.Y.Z/exo-darwin-amd64 >/usr/local/bin/exo
```
kevgo: Not sure how much easier it is to provide a curl command here. Users have to look up the version manually on the website anyways - why not have them download the correct binary right there?

We could tell them to move the binary to `/usr/local/bin` if they don't have a good other place to put it.

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -6,66 +6,68 @@
...
-as separate services.
-Because of this much narrower set of responsibilities,
-the html server is a lot smaller and simpler
-than it would be in a traditional monolithic application.
-
-Since our html server is so simple,
-we'll build it using [ExpressJS](http://expressjs.com).
-Exosphere provides a template for building ExpressJS html servers.
-Let's use it:
+Let's build the first service for our application: the HTML server! If we were
```
kevgo: Please try to match the existing flow of line breaks. The existing text breaks at logical breaking points like commas, subclauses, and each time a thought chain ends. This has several advantages: it makes it much easier to read the markdown code (okay, that's debatable, but I find it easier), and it limits changes to the respective line their are in. If we followed a constant line length, then just adding one world somewhere potentially changes all lines after it, causing massive diffs where it is hard to spot the actual change. Plus, not everybody has a sophisticated editor that automatically breaks text at constant line lengths, and we don't want to exclude these people. Thanks!  :)

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -110,138 +115,159 @@ Here is the current architecture of our application:
...
 
-
 ### The service configuration file
 
-This file tells the Exosphere framework everything it needs to know about this service.
+This file tells the Exosphere framework everything it needs to know about this
+service.
 
-__~/todo-app/html-server/service.yml__
+**~/todo-app/web-service/service.yml**
```
kevgo: Why do you change `__` to `**` here? It doesn't really change anything.

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -110,138 +115,159 @@ Here is the current architecture of our application:
...
+add3ddb21ede: Pull complete
+7b0e69015745: Pull complete
+Digest: sha256:56e176f25e21732ccad671375fba8177a04a3fc7ea8bc1decf9e715c6af961c6
+Status: Downloaded newer image for originate/exocom:0.26.1
+Building web-service
+Step 1/6 : FROM node
+latest: Pulling from library/node
+85b1f47fba49: Pull complete
+ba6bd283713a: Pull complete
+817c8cd48a09: Pull complete
```
kevgo: This output is too lengthy. Can you please eliminate some repeated lines, like these ones here showing `xxxx: Pull complete`? Maybe keep only 2 to show there can be multiple ones...

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -110,138 +115,159 @@ Here is the current architecture of our application:
...
+Attaching to exocom0.26.1, web-service
+exocom0.26.1    | ExoCom online at port 80
+exocom0.26.1    | 'web-service' registered
+web-service     | ExoRelay for 'web-service' online at port 80
+web-service     | HTML server online at port 3000
+web-service     | HTML server is running
 ```
 
-You should see output like:
+The Exosphere framework itself is written as a bunch of loosely coupled
```
kevgo: This isn't correct anymore. Exosphere is now a single monolithic Go binary. Exocom is a different product, and the web service is part of the application.

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -110,138 +115,159 @@ Here is the current architecture of our application:
...
-We got a running microservice-based web site!
+Finally, exosphere tells us that the application is now fully started and ready
+to be used. Open a browser and navigate to
+[http://localhost:3000](http://localhost:3000). We got a running
+microservice-based web site!
 
 Takeaway:
-> The web server in a microservice application is much simpler than in a monolith,
-> because it only focuses on interacting with the user via HTML.
+
```
kevgo: Why do you add an empty line here?

website/tutorial/part_2/part_2a/08_todo_service.md
```diff
@@ -9,6 +9,12 @@
 
 # Building the Todo service
 
+> Note: This is the second service of the tutorial.
```
kevgo: Do we have to say that here? I think we can assume that people follow a tutorial correctly. This could be said on every page.

kevgo (commented):
Very nice! Please revert the changes to the line breaks, then this is pretty much good to go from my perspective.

website/tutorial/part_2/part_2a/08_todo_service.md
brian-dlee: The reason I stated this here is because this page relies on concrete changes from previous pages. This page instructs you to select template number 2. If template number 1 has not been added then it makes this page function incorrectly. Basically, I added it to mitigate confusion in case anyone finds their way to this page without following the appropriate steps.

If you still feel it's unnecessary considering this then I have no problem removing it.

website/tutorial/part_2/part_2a/04_html_server.md
brian-dlee: I agree. 

I think my editor took this page by storm. I caught it in it's antics in the the following page and disabled that behavior. I'll revert all of it's formatting changes in this page.

website/tutorial/part_2/part_2a/04_html_server.md
brian-dlee: I think this was the formatter. I don't remember consciously changing this nor do I have a good reason if I did. Either way, I will blame the formatter and correct it.

website/tutorial/part_2/part_2a/04_html_server.md
brian-dlee: Were you implying that I should remove the reference to exocom from the list? I redid this but kept the text referring to exocom. If that isn't desired I will adjust.

website/tutorial/part_2/part_2a/08_todo_service.md
kevgo: I think we can assume that people aren't idiots and can follow a tutorial correctly as long as all changes we want them to do are easily visible and clearly spelled out. If there are particular concerns that people will get tripped up here, maybe we can provide a working copy of the code base for this stage somewhere in a folder? That way, if our readers don't get it working, they can look at our code and see what is different.

website/tutorial/part_1/03_installation.md
```diff
@@ -11,67 +11,34 @@
...
-* verify that you can run Exosphere commands: `exo version`
+* Download and install the binary executable for your platform from the 
+[GitHub release page](https://github.com/Originate/exosphere/releases/latest)
+* Move the binary to a location in your path and name it `exo`
 
+### Mac installation
 
-## Windows
+_Note: Depending on the privileges of `/usr/local/bin` on your system 
+some of these steps could require root privileges_
```
kevgo: I would also add: 
> If you don't want to use root privileges here, please move Exosphere to a different folder in your home directory that is in the $PATH

kevgo: Please move this hint down to the line where it applies (line 40).

website/tutorial/part_1/03_installation.md
```diff
@@ -11,67 +11,34 @@
...
   You don't want to get into the business of having to install several programming languages,
   frameworks, and other dependencies of your polyglot applications manually.
 
-The installation instructions below include
-making these infrastructure components available on your system.
+## Dependencies
 
-### macOS
+Before installing Exosphere, ensure you have the following applications installed:
+ * [Git](https://git-scm.com/)
```
kevgo: Please remove the trailing slash from the URL here.

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -6,44 +6,47 @@
...
+building our Todo app as a traditional monolith, 
+this would be the only code base 
+and perform all of the application's functionality: 
 receiving requests,
-storing todo items, user accounts, and API tokens in the database,
-configuring and using the search engine,
-tracking user sessions and permissions,
-rendering HTML for the browser, JSON for the REST API,
-sending out emails,
+storing todo items, user accounts, and API tokens in the database,
```
kevgo: Sorry to bother you about this again. Your text editor is adding trailing spaces to all lines here. Please configure it to trim trailing spaces on save. There are ways to do that for all common editor types. Google is your friend here. :)

kevgo (commented):
This is coming along nicely, almost there!  :)

website/tutorial/part_2/part_2a/04_html_server.md
kevgo: Let's focus this tutorial on Exosphere and leave Exocom and micro-services out as much as possible. Exosphere works equally well for monoliths. We can use Exocom in the example application, but don't have to go into details on how that application works.

website/tutorial/part_2/part_2a/04_html_server.md
brian-dlee: No sweat. It looks like turning the formatter off altogether caused some issues.

website/tutorial/part_2/part_2a/02_app_config.md
```diff
@@ -25,14 +25,20 @@ name: todo-app
...
+
 services:
 ```
 
 </a>
 
 It defines the name, a description, and a version of the application.
-It also defines a section for listing all the services of the application.
-It is empty right now, since our application doesn't contain any services yet.
+It also defines a section for listing all the services of the application and
```
kevgo, line 39: > It also defines a section for listing all the services and dependencies of the application. In our case this includes `exocom`.

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -6,12 +6,12 @@
   </tr>
 </table>
 
-
```
kevgo: Please don't delete this line. I think having two empty lines in front of headers helps separate them visually.

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -6,12 +6,12 @@
...
 # The HTML Server Service
 
 Let's build the first service for our application:
 the HTML server!
-If we were building our Todo app as a traditional monolith,
+If we were
+building our Todo app as a traditional monolith,
```
kevgo: Why the line break here?

website/tutorial/part_2/part_2a/04_html_server.md
```diff
@@ -35,11 +35,14 @@ than it would be in a traditional monolithic application.
 Since our html server is so simple,
 we'll build it using [ExpressJS](http://expressjs.com).
 Exosphere provides a template for building ExpressJS html servers.
-Let's use it:
+So, before we add the service,
+we will add the HTML server service template and then we will add the new service.
 
 <a class="runMarkdown_consoleWithInputFromTable">
+
```
kevgo, line 42: Why the empty line here?

kevgo (commented):
Once these changes are addressed, this is good to go. Thanks again, Brian!

website/tutorial/part_2/part_2a/04_html_server.md
brian-dlee: The code block doesn't render correctly via Markdown preview without the line break between the anchor tag and the opening of the code block (```).

kevgo: Okay, makes sense. Please let me know when this is ready to be looked at again.

brian-dlee:
Alright @kevgo, I think I've addressed all the whitespace issues. Take a look.

kevgo:
Thanks!! 👍

Issue #780, opened 2017-11-14
service env vars: external origin production support


charlierudolph:
Add `<service>_EXTERNAL_ORIGIN`support in production

PR #792, opened 2017-11-22
inject external origin endpoint in production

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/780

Also makes a change to read env vars from `run_production.yml` instead of generating those variables again. I thought this was more idiomatic, but unfortunately that meant we needed to checkout an example app in the terraform unit test, which caused a lot of changes.

No test cases have been deleted, they've simply been merged or moved around.


charlierudolph:
I'm not reviewing this as I don't think #793 should go in and this is dependent on that. I also expected this to be touching `src/docker/composebuilder/service_endpoints.go` since the value of the endpoint will be different in deployments from what it is locally.

PR #793, opened 2017-11-22
compile production env vars

hugoesthere:
When building `run_production.yml` we weren't reading from the `environment.production` block of `service.yml`


charlierudolph:
I don't think we should be doing this. To me the `environment.production` is for deployments only. Local production is only used for testing the production docker files. @alexdavid thoughts?

hugoesthere:
Closing in favor of https://github.com/Originate/exosphere/issues/801

PR #795, opened 2017-11-27
fix lint

charlierudolph:
Appears some lint updates broke our build. Nothing I know of that we could do here to prevent this type of thing in the future.

### DIFF
--- a/lint-config.json
+++ b/lint-config.json
@@ -1,6 +1,6 @@
 {
   "Exclude": ["src/terraform/bindata.go"],
-  "Disable": ["gotype", "gotypex"],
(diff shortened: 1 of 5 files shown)


lint-config.json
```diff
@@ -1,6 +1,6 @@
 {
   "Exclude": ["src/terraform/bindata.go"],
-  "Disable": ["gotype", "gotypex"],
+  "Disable": ["gas", "gotype", "gotypex"],
```
charlierudolph, line 3: There were a lot of errors in the lint output for this as `gas` was unable to load types (same problem gotype was hitting)

PR #798, opened 2017-11-27
Move buildmode to types

hugoesthere:
~This will allow for changes coming in a later PR.~

I thought I needed these changes in a later PR but it turns out I don't. Let me know what you guys think of this anyways. In the future this will allow us to do things like attach `BuildMode` as a field of `DeployConfig`. We couldn't do this before because we'd get an import cycle.

### DIFF
--- a/src/application/clean.go
+++ b/src/application/clean.go
@@ -17,9 +17,9 @@ func CleanContainers(appContext types.AppContext, writer io.Writer) error {
 	if err != nil {
 		return err
 	}
-	for _, dockerComposeFileName := range composebuilder.GetComposeFileNames() {
+	for _, dockerComposeFileName := range types.GetComposeFileNames() {
 		var composeProjectName string
-		if dockerComposeFileName == composebuilder.LocalTestComposeFileName {
+		if dockerComposeFileName == types.LocalTestComposeFileName {
 			composeProjectName = composebuilder.GetTestDockerComposeProjectName(appContext.Config.Name)
 		} else {
 			composeProjectName = composebuilder.GetDockerComposeProjectName(appContext.Config.Name)
--- a/src/application/deployer/push_application_images.go
+++ b/src/application/deployer/push_application_images.go
@@ -4,7 +4,6 @@ import (
 	"path"
 
 	"github.com/Originate/exosphere/src/aws"
-	"github.com/Originate/exosphere/src/docker/composebuilder"
 	"github.com/Originate/exosphere/src/docker/tools"
 	"github.com/Originate/exosphere/src/types"
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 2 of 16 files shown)


charlierudolph:
I don't like having all of these in the types folder. What do you think of putting the types in subfolders that make use of them (ie these would be in `composebuilder/types`). This would only be for types that would be deleted if the parent folder got deleted and are not a general purpose type.

We could extend this for other folders in our app if you agree. @alexdavid what do you think as you have mentioned not liking this folder either

hugoesthere:
Yeah I didn't like having all the types in one folder either. I think we should keep package specific types with the package as well - whether it's in a `packagename/types` folder or just in the `packagename` dir doesn't matter to me. Then types used across multiple packages should be kept in `src/types`.

This PR is still relevant though because `BuildMode` is used across multiple packages.

charlierudolph:
If we put in `packagename/types` though we can use it across multiple packages without having the import cycle (as that makes it a separate package) and keep it in a more logical place. BuildMode is already used across multiple packages but the reason it exists is because its an argument to functions in composebuilder.

hugoesthere:
If `package_name/types` objects are being used across multiple packages anyways why doesn't it make sense to just keep it in `types`? If we put them in`package_name/types` then we can't delete the parent dir as you suggested above, since these cross-package types are imported elsewhere. We'd also have 3 locations for types: `package_name/some_type.go`, `package_name/types/some_type.go`, and `src/types/some_type.go`. Wouldn't it make sense just to keep it to `package_name/some_type.go` for package-specific types, and `src/types/some_type.go` for general use types (aka any type used by more than one package)?

I would also argue that `BuildMode` has more uses outside of composebuilder (i.e. getting a certain environment's docker-compose file name).

charlierudolph:
I don't think used in multiple-packages is something we should be using to determine if something belongs in types. The `composebuilder.BuildMode` is currently used in:
* application
* application/tester
* cmd
* docker/composebuilder

You want to move it to `types` in order to allow it to be used in `types`. Its an import cycle if you try and use it in types. I don't see any good reason for it to be in there other than to avoid an import cycle.

I think the nested types in each package allows us to avoid the import cycle and keep it close to the code that is the reason for its existence. I think `DeployConfig` should be moved in `application/deployer/types`.

charlierudolph:
> I would also argue that BuildMode has more uses outside of composebuilder (i.e. getting a certain environment's docker-compose file name)

That is tied to composebuilder to me as its information about the compose file that composebuilder makes. It we stop making compose files, I don't think that function would be useful

hugoesthere:
Ok, that makes sense then. I'm okay with making this switch to `packagename/types` and reserving `src/types` for general use types. There probably shouldn't be any types defined directly under a package dir (`packagename/type.go`) right? 

Closing this PR for now. We can discuss this further here: https://github.com/Originate/exosphere/issues/800

PR #804, opened 2017-11-28
cleanup dependencies

charlierudolph:
In working with dependencies service data noticed we had some functions which weren't needed.

### DIFF
--- a/src/application/runner.go
+++ b/src/application/runner.go
@@ -1,58 +1,25 @@
 package application
 
 import (
-	"io"
 	"os"
 	"os/signal"
 	"path"
 
-	"github.com/Originate/exosphere/src/config"
-	"github.com/Originate/exosphere/src/docker/composebuilder"
 	"github.com/Originate/exosphere/src/docker/composerunner"
-	"github.com/Originate/exosphere/src/types"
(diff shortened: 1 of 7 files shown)


src/application/runner.go
```diff
@@ -1,58 +1,25 @@
...
-func NewRunner(appContext types.AppContext, writer io.Writer, dockerComposeProjectName string, buildMode composebuilder.BuildMode) (*Runner, error) {
-	serviceConfigs, err := config.GetServiceConfigs(appContext.Location, appContext.Config)
-	if err != nil {
-		return &Runner{}, err
-	}
-	allBuiltDependencies := config.GetBuiltDevelopmentDependencies(appContext.Config, serviceConfigs, appContext.Location)
-	return &Runner{
-		AppContext:               appContext,
-		ServiceConfigs:           serviceConfigs,
-		BuiltDependencies:        allBuiltDependencies,
```
charlierudolph, line 36: ServiceConfigs and BuiltDependencies were assigned here and then never used. I would guess these were needed before a refactor that simplified this file.

hugoesthere (approved):

PR #807, opened 2017-11-29
Run terraform in docker container

### LINKED ISSUE #788: run terraform in a docker image

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/788


src/application/deployer/index.go
```diff
@@ -97,19 +104,19 @@ func validateConfigs(deployConfig types.DeployConfig) error {
...
 		return err
 	}
 
-	fmt.Fprintln(deployConfig.Writer, "Retrieving secrets...")
-	secrets, err := aws.ReadSecrets(deployConfig.AwsConfig)
+	fmt.Fprintln(deployConfig.Writer, "Applying changes...")
+	err = terraform.RunApply(deployConfig, secrets, imagesMap, deployConfig.DeployServicesOnly)
 	if err != nil {
-		return err
+		if strings.Contains(err.Error(), "exit status") {
```
hugoesthere: Not sure what the best thing to do here is. Trying to catch the error that is thrown and the apply is cancelled, but I'm thinking there could be other errors in which "exit status" is a string.

src/terraform/templates/aws.tf
```diff
@@ -26,7 +26,7 @@ variable "key_name" {
 }
 
 module "aws" {
-  source = "git@github.com:Originate/exosphere.git//terraform//aws?ref={{terraformCommitHash}}"
+  source = "github.com/Originate/exosphere.git//terraform//aws?ref={{terraformCommitHash}}"
```
hugoesthere, line 29: This switches to downloading via https so we don't need to mount a user's github credentials (and so they don't need a github account at all).

src/application/deployer/index.go
charlierudolph: Since we already output everything, I think this is good

charlierudolph: I believe these two if's could be joined into one.

src/docker/tools/index.go
```diff
@@ -103,11 +103,23 @@ func PullImage(c *client.Client, image string) error {
 	return err
 }
 
-// RunInDockerImage runs the given command in a new writeable container layer
+// RunInDockerContainer runs the given command in a new writeable container layer
 // over the given image, removes the container when the command exits, and returns
 // the output string and an error if any
-func RunInDockerImage(imageName, command string) (string, error) {
-	return util.Run("", fmt.Sprintf("docker run --rm %s %s", imageName, command))
+func RunInDockerContainer(runConfig RunConfig) error {
```
charlierudolph: Thoughts on creating `RunInDockerContainer` and `RunAndPipeInDockerContainer` so that the transform from runConfig to a command could be shared? I don't like the other uses of `util.Run` with `docker run`

src/terraform/commands.go
```diff
@@ -3,14 +3,30 @@ package terraform
...
 // RunInit runs the 'terraform init' command and force copies the remote state
 func RunInit(deployConfig types.DeployConfig) error {
+	homeDir, err := util.GetHomeDirectory()
+	if err != nil {
+		return err
+	}
+	volumes := []string{
+		fmt.Sprintf("%s:/app", deployConfig.TerraformDir),
+		fmt.Sprintf("%s/.aws:/root/.aws", homeDir),
+	}
```
charlierudolph: I think this should be inline since its small enough. Gets rid of an extra variable. Same for below

src/terraform/commands.go
```diff
@@ -3,14 +3,30 @@ package terraform
...
 	backendConfig := fmt.Sprintf("-backend-config=profile=%s", deployConfig.AwsConfig.Profile)
-	return util.RunAndPipe(deployConfig.TerraformDir, []string{}, deployConfig.Writer, "terraform", "init", "-force-copy", backendConfig)
+	dockerRunConfig := tools.RunConfig{
```
charlierudolph: This could also be defined inline as the parameter instead of creating a variable. Same for below

src/terraform/commands.go
```diff
@@ -3,14 +3,30 @@ package terraform
...
+		Volumes:    volumes,
+		WorkingDir: "/app",
+		ImageName:  fmt.Sprintf("hashicorp/terraform:%s", deployConfig.TerraformVersion),
```
charlierudolph: Thoughts on making the terraform version and terraform image name constants in the terraform package? The module hash could moved in there as well.

src/docker/tools/index.go
hugoesthere: Hmm, I'm less convinced that's necessary because the two cases we would use `RunInDockerContainer` is 1) internally in `CatFile` and 2) for a test helper. Nothing in `runConfig` is set except imageName and command. It seems to me to be creating more complexity than needed. Plus `RunInDockerContainer` would be using `util.Run` as well anyways.

charlierudolph: K

src/application/deployer/index.go
```diff
@@ -97,19 +105,17 @@ func validateConfigs(deployConfig deploy.Config) error {
...
-	if err != nil {
-		return err
-	}
-
 	fmt.Fprintln(deployConfig.Writer, "Applying changes...")
-	return terraform.RunApply(deployConfig, secrets, imagesMap, deployConfig.DeployServicesOnly)
+	err = terraform.RunApply(deployConfig, secrets, imagesMap, deployConfig.DeployServicesOnly)
+	if err != nil && strings.Contains(err.Error(), "exit status") {
+		return nil
+	}
```
charlierudolph: Thoughts on checking on the err containing the "exit status" in the `terraform.RunApply`. Since its specific to that I think it belongs there and we can comment that it is expected to exit with non-zero status when the user choose to abandon the deployment

charlierudolph (approved):

PR #809, opened 2017-11-29
[refactor 2] context: use pointers

charlierudolph:


kevgo:
Nice!

hugoesthere (approved):
What benefit does this provide us?

charlierudolph:
Talked about in person but for anymore else: nicer to return nil instead of an empty version of the object. Also using pointers avoids copying large structs. This article appears to give some nice background: http://goinbigdata.com/golang-pass-by-pointer-vs-pass-by-value/

PR #810, opened 2017-11-29
[refactor 3] move deploy config into own package

charlierudolph:
I will be moving context into its own package and deployConfig contains an AppContext


src/application/deployer/push_application_images.go
```diff
@@ -6,13 +6,13 @@ import (
 	"github.com/Originate/exosphere/src/aws"
 	"github.com/Originate/exosphere/src/docker/composebuilder"
 	"github.com/Originate/exosphere/src/docker/tools"
-	"github.com/Originate/exosphere/src/types"
+	"github.com/Originate/exosphere/src/types/deploy"
```
hugoesthere, line 9: Weren't we putting these in `application/deployer/types`? Or is that not what you meant?

charlierudolph: That was what I originally thought but having packages with the same name prevents goimports from working and thus would have to deal with imports in a more manual way.

hugoesthere (approved):
Cool, I think this is the next best solution then.

PR #812, opened 2017-11-29
context refactor

charlierudolph:
I need to break this up into a couple PRs but just so you are aware. 

* Move context into its own package
* Load all service contexts when loading the user context (instead of having code that gets the service config anytime we need it
* Remove some unnecessary fields in service ServiceContext
* Update `docker-compos/test.yml` to use the same service names as the others. 
* Remove parameters `serviceConfig` and  `serviceData` from `composebuilder.GetServiceDockerCompose` as they can be retrieved from the other parameters


test/helpers/index.go
```diff
@@ -97,7 +97,10 @@ func cleanApp(appDir string) error {
 	if doesExist {
 		cmdPlus := execplus.NewCmdPlus("exo", "clean")
 		cmdPlus.SetDir(appDir)
-		return cmdPlus.Run()
+		err := cmdPlus.Run()
+		if err != nil {
+			return errors.Wrapf(err, "'exo clean' errored with output:\n%s", cmdPlus.GetOutput())
+		}
```
charlierudolph, line 103: This change was needed while debugging an error with exo clean. It had given a useless error message when it failed.

charlierudolph:
@hugobho this is now ready

hugoesthere (approved):

charlierudolph:
@hugobho nevermind. Ended up removing appContext as an argument to the serviceEndpoint since its not needed.

src/docker/composebuilder/index.go
```diff
@@ -84,10 +78,12 @@ func getServicesDockerCompose(options ApplicationOptions, portReservation *types
 	return result, nil
 }
 
-func getServiceEnvVarEndpoints(options ApplicationOptions, serviceConfigs map[string]types.ServiceConfig, portReservation *types.PortReservation) map[string]*types.ServiceEndpoints {
+func getServiceEnvVarEndpoints(options ApplicationOptions) map[string]*types.ServiceEndpoints {
+	portReservation := types.NewPortReservation()
```
hugoesthere, line 82: This will not reserve ports properly. Each time a service calls `getServiceEnvVarEndpoints`, the starting port number will be reset. The initialization of portReservation needs to be kept in `GetApplicationDockerCompose` and passed to each service that uses `getServiceEnvVarEndpoints`

src/docker/composebuilder/service_endpoints_test.go
```diff
@@ -1,56 +0,0 @@
...
-			Mount:       true,
-			Environment: types.BuildModeEnvironmentProduction,
-		}
-		s := types.NewServiceEndpoint(appContext, serviceRole, serviceConfigs[serviceRole], portReservation, buildMode)
-		endpoints := s.GetEndpointMappings()
-		Expect(endpoints["WEB_EXTERNAL_ORIGIN"]).To(Equal("http://localhost:3000"))
-		mapping := s.GetPortMappings()
-		Expect(mapping[0]).To(Equal("3000:80"))
-	})
-})
```
hugoesthere, line 56: Did I forget to delete this file?

src/terraform/command_helpers.go
```diff
@@ -86,16 +86,16 @@ func compileDependencyVars(deployConfig deploy.Config) ([]string, error) {
...
 func compileServiceEnvVars(deployConfig deploy.Config, secrets types.Secrets) ([]string, error) {
 	envVars := []string{}
-	for serviceRole, serviceConfig := range deployConfig.ServiceConfigs {
+	for serviceRole, serviceContext := range deployConfig.AppContext.ServiceContexts {
 		serviceEnvVars := map[string]string{"ROLE": serviceRole}
-		dependencyEnvVars := getDependencyServiceEnvVars(deployConfig, serviceConfig, secrets)
+		dependencyEnvVars := getDependencyServiceEnvVars(deployConfig, serviceContext.Config, secrets)
 		util.Merge(serviceEnvVars, dependencyEnvVars)
-		productionEnvVar, serviceSecrets := serviceConfig.GetEnvVars("remote")
+		productionEnvVar, serviceSecrets := serviceContext.Config.GetEnvVars("production")
```
hugoesthere: This should be "remote", no?

src/docker/composebuilder/index.go
charlierudolph: Right below this is where we loop through the services. This function is also only called in one place

src/docker/composebuilder/service_endpoints_test.go
charlierudolph: yep its still in master

src/terraform/command_helpers.go
charlierudolph: correct. Bad merge on my part

hugoesthere (approved):

PR #815, opened 2017-11-29
remove test role

charlierudolph:
extracted from #812. Updates the `test.yml` to look more similar to the other compose files. When testing we find one service role that uses the service. This is a new restriction that you can't test a service that is not used by your application. I don't think that will be an issue though. 

For now separated the clean tests. Will work on removing the container names from the docker compose files in another PR since we currently can't run the app and the tests at the same time


src/types/app_context.go
```diff
@@ -11,12 +9,12 @@ type AppContext struct {
 }
 
 // GetServiceContext returns a ServiceContext for the service found at the given directory base
-func (a *AppContext) GetServiceContext(serviceLocation string) (*ServiceContext, error) {
-	serviceConfig, err := NewServiceConfig(serviceLocation)
+func (a *AppContext) GetServiceContext(serviceRole string, serviceData ServiceData) (*ServiceContext, error) {
```
hugoesthere: Can't `serviceData` be determined from `a.Config.Services[serviceRole]`?

src/types/app_context.go
```diff
@@ -11,12 +9,12 @@ type AppContext struct {
...
+	serviceConfig, err := NewServiceConfig(path.Join(a.Location, serviceData.Location))
 	return &ServiceContext{
-		Dir:        path.Base(serviceLocation),
-		Location:   serviceLocation,
+		Role:       serviceRole,
 		Config:     serviceConfig,
 		AppContext: a,
+		AppData:    &serviceData,
```
hugoesthere: Would it make more sense to call this ServiceData? After all that's what it represents.

hugoesthere (commented):
This still doesn't address the use case of being able to run `exo run` and `exo test` simultaneously right? Were we going to try to remove container names to do that?

PR #831, opened 2017-12-01
Switch service dependencies to local/remote

### LINKED ISSUE #802: service dependencies: local/remote
Currently `service.yml` has
```yml
development:
  dependencies:
  scripts:
  port:

production:
  dependencies:
  url:
  cpu:
  memory:
  port:
  health-check:
```

Update this to
```yml
development:
  scripts:
  port:

production:
  port:

local:
  dependencies:

remote:
  dependencies:
  url:
  cpu:
  memory:
  health-check:
```

@alexdavid @hugobho please review

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/802


src/types/service_config.go
```diff
@@ -11,6 +11,12 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
+// PublicService is the value for the type field of a public service
+const PublicService = "public"
+
+// WorkerService is the value for the type field of a worker service
+const WorkerService = "worker"
```
charlierudolph: 👍 very nice

src/types/local_config.go
```diff
@@ -1,6 +1,6 @@
 package types
 
-// LocalConfig represents development specific configuration for an application
-type LocalConfig struct {
```
charlierudolph: I think this should still be called the `LocalConfig`. Why was it renamed?

charlierudolph (approved):

src/types/local_config.go
hugoesthere: Because it didn't contain anything other than dependencies, but I can change it back

src/types/service_config.go
charlierudolph: Actually thoughts on these being `ServiceTypePublic` and `ServiceTypeWorker`?

PR #833, opened 2017-12-01
fail deploy if terraform is out of date

### LINKED ISSUE #830: exo deploy: always fail if terraform is out of date
Instead of only when passing `--update-services`

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/830


charlierudolph (approved):

charlierudolph:
Actually we should update the error message to something like 
```
"terraform/main.tf" is out of date. Please run "exo generate terraform" and review the changes
```

PR #839, opened 2017-12-05
Inject internal origin env vars

### LINKED ISSUE #836: service endpoint internal origin env var
from #685

For both local and remote

@hugobho please start on this if you need something to work on

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/836


charlierudolph:
Can we also create a feature test for this? And example could be a web service which makes an internal http call to another service in determining what to display.

src/types/endpoints/service_endpoint.go
```diff
@@ -51,7 +52,10 @@ func (s *ServiceEndpoint) GetPortMappings() []string {
 func (s *ServiceEndpoint) GetEndpointMappings() map[string]string {
 	switch s.ServiceConfig.Type {
 	case "public":
-		return s.getExternalOrigin()
+		externalOrigin := s.getExternalOrigin()
+		internalOrigin := s.getInternalOrigin()
```
charlierudolph: Thoughts on joining these two functions? Maybe into `getPublicEndpointMappings`

features/run/service_endpoint_environment_variables.feature
```diff
@@ -1,12 +1,14 @@
...
   As an Exosphere developer
-  I want the services to know where each other are available from the outside world
-  So clients can talk directly to other services
+  I want the services to know where each other are available
+  So that communication between services are possible
 
   Rules:
   - "<NAME>_EXTERNAL_ORIGIN" is passed as an environment variable to
     each service for any public service
+  - "<NAME>_ORIGIN" is passed as an environment variable to
```
charlierudolph: thoughts on making this env var `<NAME>_INTERNAL_ORIGIN`. I think that is more clear.

features/run/service_endpoint_environment_variables.feature
```diff
@@ -18,3 +20,12 @@ Feature: Service endpoint enviornment variable
...
       Backend located at http://localhost:3000
       """
+
+  Scenario: internal origin env var available at run time
+    Given I am in the root directory of the "service-internal-origin" example application
+    And starting "exo run" in my application directory
+    And it prints "frontend service online" in the terminal
+    Then http://localhost:3010 displays:
+      """
+      Backend service reached
```
charlierudolph: Thoughts on updating this to `Backend service content`?

src/types/endpoints/service_endpoint.go
```diff
@@ -51,22 +51,26 @@ func (s *ServiceEndpoint) GetPortMappings() []string {
...
 	}
 }
 
-func (s *ServiceEndpoint) getExternalOrigin() map[string]string {
+func (s *ServiceEndpoint) getPublicEndpoints() map[string]string {
 	externalKey := fmt.Sprintf("%s_EXTERNAL_ORIGIN", toConstantCase(s.ServiceRole))
+	internalKey := fmt.Sprintf("%s_INTERNAL_ORIGIN", toConstantCase(s.ServiceRole))
+	endpoints := map[string]string{}
 	if s.BuildMode.Type == types.BuildModeTypeLocal {
 		if s.HostPort != "" {
```
charlierudolph: Unrelated to this PR but why do we check the hostPort is defined here? I don't think that is necessary to check at this point. Maybe in another PR we can set the hostPort in test environment and thus be able to remove this check

test/fixtures/applications/service-internal-origin/frontend-service/service.yml
```diff
@@ -0,0 +1,8 @@
+type: public
+description:
+author:
+
+development:
+  port: 3000
```
charlierudolph: Can we use port 8080 for both services. That way the backend service doesn't have the same host port and container port and we can validate things a bit better

charlierudolph (commented):
I think you need a few more updates to complete the rename to `<SERVICE>_INTERNAL_ORIGIN`

src/types/endpoints/service_endpoint.go
hugoesthere: Host port is only populated once a container port for the service has been attained. This was just to handle cases where the user hasn't defined a container port for a service yet to prevent passing env vars that looked like `https://localhost:`

charlierudolph: Then I think this should also prevent us from returning the env vars when building for remote. Could this become
```go
if s.ContainterPort != "" {
  if s.BuildMode.Type == types.BuildModeTypeLocal {
    // set local endpoint vars
  } else {
    // set remote endpoint vars
  }
}
```

hugoesthere: Changed it to that. Updated the constructor as a result

charlierudolph (approved):

PR #845, opened 2017-12-05
docker-compose: remove container name

charlierudolph:
From docker-compose docs:

> by default, any service can reach any other service at that service’s name

Thus we don't need to specify these container names in order to allow them to access each other by the service name which is all we were doing. Not specifying the container names allows docker-compose to generate names for each which uses the form: `<COMPOSE_PROJECT_NAME>_<SERVICE_NAME>_<NUMBER>`


charlierudolph:
@hugobho okay this is ready for review

hugoesthere (commented):
Now that I look at this, it seems like it will be more difficult to reach these containers since their names become more complex. What was the reason for this change again?

I also think we'd have to update `INTERNAL_ORIGIN_ENDPOINT` to point to these new container names.

charlierudolph:
The services are reachable via the service name (the key in the compose file). https://docs.docker.com/compose/networking/

alexdavid (approved):
Awesome, +1 for utilizing docker compose to remove code from our repo

hugoesthere:
That's true, but isn't the whole point of `INTERNAL_ORIGIN` to set it to that service name so other services can just hit `#{SERVICE_NAME}_INTERNAL_ORIGIN` instead of hard-coding the service name?

charlierudolph:
Correct. We do still need `<X>_INTERNAL_ORIGIN` but we are now setting it to the service name instead of the container name. We were setting the container name to be the same as the service name so that code doesn't need any updates to keep working.

hugoesthere (approved):
Right, forgot we had it implemented as service name already. Looks good 👍

Issue #841, opened 2017-12-05
Exo generate terraform doesn't error if there are config schema errors


alexdavid:
I would expect any config schema parsing errors to occur on `exo generate terraform` just as they do for `exo deploy`

charlierudolph:
Fixed in eea2873588ef24abe720466fa56280252cc6477d

PR #848, opened 2017-12-06
exo-generate: check for config schema errors

hugoesthere:
Resolves: https://github.com/Originate/exosphere/issues/841


src/cmd/generate.go
```diff
@@ -55,7 +55,9 @@ var generateTerraformCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 		deployConfig := getBaseDeployConfig(userContext.AppContext)
-		err = terraform.GenerateFile(deployConfig)
+		writer := os.Stdout
```
charlierudolph: This write variable seems unneeded. Could we write `deployConfig.Writer = os.Stdout`?

src/application/deployer/index.go
```diff
@@ -12,7 +12,7 @@ import (
 // StartDeploy starts the deployment process
 // nolint gocyclo
 func StartDeploy(deployConfig deploy.Config) error {
-	err := validateConfigs(deployConfig)
+	err := ValidateConfigs(deployConfig)
```
charlierudolph: Can be dealt with separately, but I think this function should be updated to check the terraform file contents are up to date here (instead of after pushing the docker images). I also think we should not write any terraform updates to file as we will always role them back. Thus we could diff it in the same way we are with docker-compose.

charlierudolph: As part of that I think the ValidateConfigs function should move into the generate file instead of in here as I don't think we need to validate configs as part of deploy and only as part of the generate process.

hugoesthere: We aren't rolling back file changes anymore - we got rid of that when we switched to Terraform 0.11.0: https://github.com/Originate/exosphere/commit/3218aee19bf720b1202b6ac82d92f9b187c78558

I also didn't like that feature much anyways. I think it was much more helpful to see those changes written to file that way you could diff them with source control.

hugoesthere: Nevermind, with checking for terraform updates before running Terraform commands we should be able to preserve the previous file. And we can always diff with source control by running `exo generate terraform`

hugoesthere:
I moved checking the terraform file to before we start any of the deployment process, but I kept `ValidateConfigs` as the first step because in order to check if the terraform file is up-to-date, you have to generate one to compare to the existing one, and in order to generate one you must have valid configs. Let me know what you think: https://github.com/Originate/exosphere/pull/851

charlierudolph (approved):

hugoesthere:
Shipped in https://github.com/Originate/exosphere/commit/eea2873588ef24abe720466fa56280252cc6477d

Issue #843, opened 2017-12-05
Health-check should be under development/produciton not remote


alexdavid:
A healthcheck describes where in the application I can get a 200 response, which depends on the dockerfile provided not the environment in which it's deployed

charlierudolph:
Fixed in 5f1a0219be6793d5f174e7cd63a79382fe6fba8e

PR #850, opened 2017-12-06
move health-check into development/production blocks

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/843


src/types/service_development_config.go
```diff
@@ -2,6 +2,7 @@ package types
 
 // ServiceDevelopmentConfig represents development specific configuration for a service
 type ServiceDevelopmentConfig struct {
-	Scripts map[string]string `yaml:",omitempty"`
-	Port    string            `yaml:",omitempty"`
+	HealthCheck string            `yaml:"health-check,omitempty"`
```
charlierudolph: Is it worth it to have this here just for consistency? We aren't using it for anything at the moment. @alexdavid thoughts?

alexdavid: Probably YAGNI, at least in the near term. We can always add it if needed

charlierudolph: Alright @hugobho lets remove this.

charlierudolph (approved):
Good to merge after deleting it from the development config

PR #851, opened 2017-12-07
Check terraform files before pushing images

hugoesthere:
I'm not sure why github is rending my changes so weirdly, but basically I moved a lot of the diff-ing logic into a new function `checkTerraformFile` and got rid of the old way of writing to file to diff


src/application/deployer/index.go
```diff
@@ -32,32 +37,46 @@ func StartDeploy(deployConfig deploy.Config) error {
...
 		return err
 	}
-	err = terraform.GenerateFile(deployConfig)
+
+	fmt.Fprintln(deployConfig.Writer, "Applying changes...")
+	return terraform.RunApply(deployConfig, secrets, imagesMap, deployConfig.AutoApprove)
+}
+
+func checkTerraformFile(deployConfig deploy.Config) error {
+	_, err := fmt.Fprintln(deployConfig.Writer, "Generating Terraform files...")
```
charlierudolph: Thoughts on updating this to "Validating terraform..."?

charlierudolph (approved):

PR #853, opened 2017-12-07
generalize dependency data

### LINKED ISSUE #756: generalize service data env var for dependencies
Want to do this in order to remove exocom's code for loading the service routes. Maybe something like:

```yml
# service1/service.yml
dependency-data:
  exocom:
    a: b

# service2/service.yml
dependency-data:
  exocom:
    c: d

# application.yml
services:
  serviceA:
    location: './service1'
    dependeny-data:
      exocom:
        e: f
  serviceB:
    location: './service2'
```
Then exocom would receive the env var `SERVICE_DATA` with the content
```json
{
  "serviceA": {
    "a": "b",
    "e": "f"
  },
  "serviceB": {
    "c": "d"
  }
}
```
This should allow us to delete `compileServiceRoutes` from `config.exocomDevelopmentDependency`

The following should also be removed since they are exocom specific:
* `MessageTranslations` from `types.ServiceData` 
* `ServiceMessages` from `types.ServiceConfig`

Exocom will need to be updated to support this new version of env variables

@alexdavid @hugobho thoughts?

charlierudolph:
resolves #756

Ended up going with this for now in order to support exocom. We can extract further down the line if needed

I'll update exocom before merging this in


src/types/context/app_context.go
```diff
@@ -33,6 +33,15 @@ func (a *AppContext) GetServiceContextByLocation(location string) *ServiceContex
 	return nil
 }
 
+// GetDependencyServiceData returns the service data for a particular dependency
```
hugoesthere: Can we make this a little more clear, something like: returns a map from service role to a service's dependency data (or a dependency's service data, whichever makes more sense)

src/types/context/service_context.go
```diff
@@ -18,3 +18,22 @@ type ServiceContext struct {
 func (s *ServiceContext) ID() string {
 	return path.Base(s.Source.Location)
 }
+
+// GetDependencyData returns the data to pass to a dependency for theis service
```
hugoesthere: Maybe: "returns the dependency data for this service"

hugoesthere (approved):
Looks good, ship when green

Issue #755, opened 2017-11-09
development generic dependencies


charlierudolph:
update so the user specifies an image (allowing for using name-spaced images) and a name.

```yml
# current 
dependencies: 
  # (mongo3.0.4 becomes the service name) 
  - name: mongo
    version: 3.0.4

# new
dependencies: 
  # (my-custom will be the service name)
  my-custom:
    image: 'mongo:0.26.4'
```

Also note that switching to using an object means we will need to sort by the dependency name in order to ensure out docker-compose file stays consistent.

@alexdavid please review to see if this captures the issue you ran into yesterday

alexdavid:
👍 I also like that we join the imagename/version like the docker convention.

What are your thoughts on not making this an array and instead making the dependencies a map from name -> configuration
```yaml
my-custom:
  image: 'mongo:0.26.4'
  config: {}
```

Wouldn't this make it easier to merge the service and application dependencies?

charlierudolph:
I'm fine with that I don't think there is a reason for deps being an array.

hugoesthere:
Should we also get rid of referring to a dependency as `#{dependency_name}#{version}`? We use this in several places such as for docker-compose.yml keys and container names. We had this initially to support the same dependency of different versions.

charlierudolph:
I believe we can get rid of referring to it by `#{dependency_name}#{version}` since we will now use the custom name. With this new version they can still use the same dependency with different versions of even now multiple copies of the same dependency.

charlierudolph:
Tied to this I think we can do the following
```yml
local:
  - name:
    version:
    config:

remote:
  - name:
    version:
    config:

# becomes

local:
  <id>:
     type: (exocom|nats|generic) # generic by default, field goes away once everything is generic
     image: # always required
     config: 

remote:
  <id>:
     type: (exocom|rds|generic) # for rds name & version for rds moves into config as engine & engineVersion, for exocom version moves into config
     config:
```

Would be done in two parts. Local first and then remote second. 
@hugobho @alexdavid thoughts

PR #854, opened 2017-12-08
Generic dependencies: refactor local dependencies

hugoesthere:
Resolves local version of: https://github.com/Originate/exosphere/issues/755

### DIFF
--- a/src/docker/composebuilder/service_builder_test.go
+++ b/src/docker/composebuilder/service_builder_test.go
@@ -15,28 +15,29 @@ var _ = Describe("ComposeBuilder", func() {
 	var _ = Describe("building external dependencies", func() {
 		var dockerCompose *types.DockerCompose
 		var serviceEndpoints map[string]*endpoints.ServiceEndpoint
+		var serviceRole string
 
 		var _ = BeforeEach(func() {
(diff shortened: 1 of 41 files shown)


test/fixtures/applications/rds/my-sql-service/service.yml
```diff
@@ -21,8 +21,9 @@ remote:
 
 local:
   dependencies:
-    - name: mysql
-      version: 5.6.17
+    mysql:
+      type: mysql
```
charlierudolph: Can we remove this type since its not used? Same for any type that is not exocom / nats.

src/docker/composebuilder/service_builder_test.go
```diff
@@ -15,28 +15,29 @@ var _ = Describe("ComposeBuilder", func() {
 	var _ = Describe("building external dependencies", func() {
 		var dockerCompose *types.DockerCompose
 		var serviceEndpoints map[string]*endpoints.ServiceEndpoint
+		var serviceRole string
```
charlierudolph: I'd prefer not having this variable and instead harding the service role everywhere. I think that makes the test easier to understand. Duplication in tests doesn't matter as much to me and I would rather have it be easy to understand

charlierudolph (approved):

PR #855, opened 2017-12-08
Generic dependencies: refactor remote dependencies

### LINKED ISSUE #755: development generic dependencies

hugoesthere:
Partially resolves: https://github.com/Originate/exosphere/issues/755

### DIFF
--- a/src/application/deployer/index.go
+++ b/src/application/deployer/index.go
@@ -8,7 +8,6 @@ import (
 	"github.com/Originate/exosphere/src/terraform"
 	"github.com/Originate/exosphere/src/types"
 	"github.com/Originate/exosphere/src/types/deploy"
-	"github.com/Originate/exosphere/src/util"
 )
 
 // StartDeploy starts the deployment process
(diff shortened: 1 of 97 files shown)


src/application/deployer/index.go
```diff
@@ -92,23 +92,19 @@ func ValidateConfigs(deployConfig deploy.Config) error {
...
 		if err != nil {
 			return err
 		}
-		validatedDependencies[dependency.Name] = dependency.Version
 	}
 
 	fmt.Fprintln(deployConfig.Writer, "Validating service dependencies...")
 	for _, serviceContext := range deployConfig.AppContext.ServiceContexts {
 		for _, dependency := range serviceContext.Config.Remote.Dependencies {
-			if validatedDependencies[dependency.Name] == "" {
```
hugoesthere, line 107: I don't think this check is needed anymore

src/config/remote_generic_dependency.go
```diff
@@ -12,24 +10,18 @@ type remoteGenericDependency struct {
...
 // GetDockerConfig returns docker configuration and an error if any
 func (g *remoteGenericDependency) GetDockerConfig() (types.DockerConfig, error) {
-	return types.DockerConfig{
-		Image: fmt.Sprintf("%s:%s", g.config.Name, g.config.Version),
-	}, nil
-}
-
-func (g *remoteGenericDependency) GetServiceName() string {
-	return g.config.Name + g.config.Version
+	return types.DockerConfig{}, nil
```
hugoesthere, line 18: We aren't using generic remote dependencies. When we do support them, we have to define a `docker-image` field somewhere under `remote.dependencies`, and specify that remote generic dependencies are for docker type dependencies only

charlierudolph: Lets just get rid of remote generic dependencies then and if the type isn't exocom/rds we throw an error

charlierudolph (approved):

src/config/remote_generic_dependency.go
hugoesthere: Hmm, we'd be breaking a lot of apis if we started returning errors with `NewRemoteAppDependency`. Do you think it makes sense to keep remote generic dependency as a placeholder until we define what it does exactly?

charlierudolph: Okay. Good to leave as is for now since this will be changing soon

PR #857, opened 2017-12-11
release 0.33.0

hugoesthere:


CHANGELOG.md
```diff
@@ -1,3 +1,98 @@
+## 0.33.0 (2017-12-11)
+
+### BREAKING CHANGES
+* Local and remote dependencies have been refactored to be more generic:
+```
```
charlierudolph: Please add the syntax here. Thus this line would be "```yml". Same for the other blocks

CHANGELOG.md
```diff
@@ -1,3 +1,98 @@
...
+# becomes:
+
+local:
+  environment:
+
+remote:
+  environment:
+  secrets:
+
+# default is removed in favor of using yaml defaults: https://github.com/Originate/exosphere/issues/429#issuecomment-332694981
```
charlierudolph: I don't think we need to reference the issue here

CHANGELOG.md
```diff
@@ -1,3 +1,98 @@
...
+            internal: mongo ping
+```
+
+#### New Features
+* For each public service, inject `<PUBLIC_SERVICE>_INTERNAL_ORIGIN` environment variable to every other service. This points to exoposed origins on an internal network
+* Switch local secrets to using variables in generated docker-compose files
+* `exo generate`: check for application and service config schema errors
+* Remote container names from docker-compose files
+* Fix app config validation error messages
+* Move `health-check` field from remote into production block
```
charlierudolph: This is also a breaking change

charlierudolph (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,98 @@
...
+# becomes
+services:
+  users-service:
+    dependency-data:
+      exocom:
+        translations:
+          - public: users ping
+            internal: mongo ping
+```
+* Move `health-check` field from remote into production block
```
charlierudolph, line 85: Can we add an example of this as well?

PR #864, opened 2017-12-12
release 0.33.1

charlierudolph:
Moving forward what do you think about having each PR with a user-facing change to update the changelog under the Unreleased section I added?


hugoesthere (approved):
👍 to updating changelog with each PR

PR #867, opened 2017-12-12
local dependency: remove ports

charlierudolph:
This was a hold over and I don't think makes with our updated port reservation


hugoesthere:
Can services still hit dependencies now? How does this work if a docker container doesn't expose any ports?

charlierudolph:
The `ports` is used for linking container ports to host ports: https://docs.docker.com/compose/compose-file/#ports. The ports are always available on the container within the docker network

hugoesthere (approved):

PR #871, opened 2017-12-12
website: cleanup volumes

charlierudolph:
update to the latest options


charlierudolph:
Closing as I think we should update the website once we hit a more stable version

Issue #751, opened 2017-11-09
extract production dependencies


charlierudolph:
In order to make these more pluggable:
* remove the go code associated with exocom/nats/rds production dependencies
~* `DEPENDENCY_<name>_HOST` should be injected into services automatically~
* A production dependency template should be a terraform template and readme with a suggestion of what docker image to use in development and what should be specified in the config
  * could possibly include a yml file listing the keys it expects to be present in the config that exosphere could validate.
  * the template will be passed a config object (from app/service.yml)
  * for now can live in new `production-dependency-templates` folder in this repo (can extract to own repo later / add in ability to load other templates)
* the dependency will have the following structure when used in app/service.yml
    ```yml
    template: <name>
    template-config:  
       engine: postgres
       password_secret_name: some_key
       ...
    ``` 
   ~the merged config key/values will also be passed as service env vars as `DEPENDENCY_<NAME>_CONFIG_<KEY>`~

@hugobho @alexdavid please review

charlierudolph:
@alexdavid did you get a chance to circle back around to this?

hugoesthere:
```
#config.yml
service-env-vars: #array of keys or map?
  - DEPENDENCY_<name>_HOST
  - <env var>
  - <env var>

docker-image:

```

PR #872, opened 2017-12-13
[extract production dependencies 2] Get deployment data

hugoesthere:
Part of: #751


src/terraform/command_helpers.go
```diff
@@ -148,3 +132,12 @@ func getDependencyServiceEnvVars(deployConfig deploy.Config, serviceConfig types
 	}
 	return result
 }
+func extractDependencyData(dependencyName, dependencyType string, deployConfig deploy.Config) (string, error) {
```
charlierudolph: Thoughts on naming this `GetDependencyEnvVariables`?

src/terraform/command_helpers.go
```diff
@@ -148,3 +132,12 @@ func getDependencyServiceEnvVars(deployConfig deploy.Config, serviceConfig types
...
+	serviceData := deployConfig.AppContext.GetDependencyServiceData(dependencyType)
+	serviceDataBytes, err := json.Marshal(serviceData)
+	if err != nil {
+		return "", err
+	}
+	keyName := strings.ToUpper(fmt.Sprintf("%s_DEPENDENCY_DATA", dependencyName))
```
charlierudolph: Shouldn't this be `SERVICE_DATA` instead of `%s_DEPENDENCY_DATA`?

src/terraform/command_helpers.go
```diff
@@ -148,3 +132,12 @@ func getDependencyServiceEnvVars(deployConfig deploy.Config, serviceConfig types
 	}
 	return result
 }
+func extractDependencyData(dependencyName, dependencyType string, deployConfig deploy.Config) (string, error) {
+	serviceData := deployConfig.AppContext.GetDependencyServiceData(dependencyType)
```
charlierudolph: I think the input to `GetDependencyServiceData` should be `dependencyName` instead of `dependencyType`

src/config/local_app_dependency_test.go
```diff
@@ -108,39 +108,6 @@ var _ = Describe("LocalAppDependency", func() {
...
-							},
-						},
-					},
-				})
-				Expect(err).NotTo(HaveOccurred())
-				Expect(variables).To(Equal(map[string]string{
-					"SERVICE_DATA": string(serviceData),
-				}))
-			})
-		})
```
charlierudolph, line 142: Can these unit tests be replaced with some in the terraform package?

hugoesthere: They are. You've covered them by testing the dependency data stuff.

charlierudolph: k

src/terraform/command_helpers.go
hugoesthere: We already have a similarly named `compileDependencyVars`, to which `extractDependencyData` is a helper. Don't you think the latter is a more clear name of what the function is doing?

charlierudolph: `compileDependencyVars` is getting the env vars for all dependencies. Thoughts on renaming that to  `compileDependenciesVars` (dependency becomes plural) and renaming `extractDependencyData` to `compileDependencyVar`? I think `extractDependencyData` isn't a good name because the function is doing more than that.

src/terraform/command_helpers.go
```diff
@@ -148,3 +132,12 @@ func getDependencyServiceEnvVars(deployConfig deploy.Config, serviceConfig types
 	}
 	return result
 }
+func extractDependencyData(dependencyName string, deployConfig deploy.Config) (string, error) {
+	serviceData := deployConfig.AppContext.GetDependencyServiceData(dependencyName)
+	serviceDataBytes, err := json.Marshal(serviceData)
+	if err != nil {
+		return "", err
+	}
+	keyName := strings.ToUpper("SERVICE_DATA")
```
charlierudolph: I think ToUpper / the variable keyName is not unneeded. Can we update the unit tests to validate that the key is `SERVICE_DATA`. I would have liked a test failure when we had used the wrong key name

src/terraform/command_helpers.go
hugoesthere: I modified the `extractDependencyData` so that that's solely what it's doing - since dependency data can be used for more than just env vars I think it's apt to keep that name. I also moved all terraform flag/env var related things back up to `compileDependencyVars` since that logic is more specific these command helpers.

src/terraform/command_helpers.go
```diff
@@ -148,3 +135,8 @@ func getDependencyServiceEnvVars(deployConfig deploy.Config, serviceConfig types
 	}
 	return result
 }
+func extractDependencyData(dependencyName string, deployConfig deploy.Config) (string, error) {
```
charlierudolph: Can we name this `getDependencyServiceData`? (and add a newline above it so we have en empty line between functions). Feels weird that the return value of `extractDependencyData` is assigned to `serviceData`

charlierudolph (approved):
Good to merge once previous outstanding comment is addressed

PR #873, opened 2017-12-13
[extract production dependencies 3] Generate dependency terraform

hugoesthere:
Part of: https://github.com/Originate/exosphere/issues/751


bin/build_bin_data
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -e
 
-go-bindata -o src/terraform/bindata.go -pkg "terraform" src/terraform/templates/...
```
charlierudolph: I think the terraform templates (for public service, etc) should stay in terraform for now. We could move those up into their own folder if we want to make it easier to create new types of dependencies. I think we would want to do that separately though

charlierudolph (approved):

Issue #883, opened 2017-12-14
cleanup `exo configure`


charlierudolph:
Currently `exo configure` creates a secrets store when not given a subcommand. I think we should remove this and have `exo configure create` create the secrets store if necessary. I think the current setup is a bit awkward and anything with subcommands should not do anything on its own.

@hugobho thoughts?

PR #884, opened 2017-12-14
[multiple remotes 8] deploy to multiple environments

### LINKED ISSUE #498: exo deploy: ability to deploy to multiple environments
I want to be able to use exo-deploy to deploy multiple environments such as dev / qa / staging / production. We already have the idea of a `dev` section which may need to be renamed to `local`.

@hugobho update this issue as we discover more about the requirements / how we would do this

charlierudolph:
resolves #498

Sorry for the size of this PR, uncertain how to break it up further. Let me know if you have any ideas for things I could extract

* the remote section in `application.yml` / `service.yml` is now a map from string to remote config 
* `exo configure` / `exo deploy` now have the required argument of the remote-id.
* the `terraform/...` directory becomes `terraform/<remote-id>/...` so we will have a different directory for each remote.
* I did not update the error message for remote dependencies to include the remote id as those are changing soon anyways. I'd like to clean up the validation process a bit more in a follow up PR.


features/configure.feature
```diff
@@ -0,0 +1,148 @@
+Feature: configure command
+
+  Scenario: 'exo configure' with not enough arguments
+    When starting "exo configure" in the terminal
+    Then I see:
+      """
+      Error: Wrong number of arguments
+      Usage:
+        exo configure [remote-id] [flags]
+        exo configure [command]
```
hugoesthere: What's the command that's referred to here? Is it just `exo configure help`?

hugoesthere (approved):

features/configure.feature
charlierudolph: create/delete/read/update

charlierudolph: See #883 for a related thought

hugoesthere: I'm confused. I thought `exo configure read qa` is a valid call. None of the usage examples match that?

charlierudolph:
Closing in favor of #893

PR #885, opened 2017-12-14
Refactor terraform var builders

hugoesthere:
Refactoring our command helpers to make things easier to test, and to add to/modify these var flags in the future

### DIFF
--- a/src/terraform/command_helpers.go
+++ b/src/terraform/command_helpers.go
@@ -14,20 +14,20 @@ import (
 
 // CompileVarFlags compiles the variable flags passed into a Terraform command
 func CompileVarFlags(deployConfig deploy.Config, secrets types.Secrets, imagesMap map[string]string) ([]string, error) {
-	varFlagsMap := CompileDockerImageVars(deployConfig, imagesMap)
-	envVars, err := CompileServiceEnvVars(deployConfig, secrets)
+	varMap := GetDockerImageVarMap(deployConfig, imagesMap)
+	servicesVarMap, err := GetServicesVarMap(deployConfig, secrets)
 	if err != nil {
 		return []string{}, errors.Wrap(err, "cannot compile service environment variables")
 	}
-	util.Merge(varFlagsMap, envVars)
-	dependencyVars, err := CompileDependencyVars(deployConfig)
+	util.Merge(varMap, servicesVarMap)
+	dependenciesVarMap, err := GetDependenciesVarMap(deployConfig)
 	if err != nil {
 		return []string{}, errors.Wrap(err, "cannot compile dependency variables")
 	}
-	util.Merge(varFlagsMap, dependencyVars)
-	util.Merge(varFlagsMap, map[string]string{"aws_profile": deployConfig.AwsConfig.Profile})
-	util.Merge(varFlagsMap, secrets)
-	return BuildFlags(varFlagsMap), nil
+	util.Merge(varMap, dependenciesVarMap)
+	util.Merge(varMap, secrets)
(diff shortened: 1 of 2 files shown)


src/terraform/command_helpers.go
```diff
@@ -14,79 +14,79 @@ import (
...
+	dependencyVars, err := CompileDependencyVars(deployConfig)
 	if err != nil {
 		return []string{}, errors.Wrap(err, "cannot compile dependency variables")
 	}
-	vars = append(vars, dependencyVars...)
-	return append(vars, "-var", fmt.Sprintf("aws_profile=%s", deployConfig.AwsConfig.Profile)), nil
+	util.Merge(varFlagsMap, dependencyVars)
+	util.Merge(varFlagsMap, map[string]string{"aws_profile": deployConfig.AwsConfig.Profile})
+	util.Merge(varFlagsMap, secrets)
+	return BuildFlags(varFlagsMap), nil
```
charlierudolph: Thoughts on making the variable names:
* varFlagsMap -> varMap
* envVars -> servicesVarMap
* dependencyVars -> dependenciesVarMap

And the function names:

* CompileDockerImageVars -> GetDockerImageVarMap
* CompileServiceEnvVars -> GetServicesVarMap
* CompileDependencyVars -> GetDependenciesVarMap

Just want to make these a bit more consistent. Also within the functions use `varMap` as the return value

src/terraform/command_helpers.go
```diff
@@ -14,79 +14,79 @@ import (
...
-	dependencyVars, err := compileDependencyVars(deployConfig)
+	util.Merge(varFlagsMap, envVars)
+	dependencyVars, err := CompileDependencyVars(deployConfig)
 	if err != nil {
 		return []string{}, errors.Wrap(err, "cannot compile dependency variables")
 	}
-	vars = append(vars, dependencyVars...)
-	return append(vars, "-var", fmt.Sprintf("aws_profile=%s", deployConfig.AwsConfig.Profile)), nil
+	util.Merge(varFlagsMap, dependencyVars)
+	util.Merge(varFlagsMap, map[string]string{"aws_profile": deployConfig.AwsConfig.Profile})
```
charlierudolph: I don't think its worth using util.Merge here and instead we can use:
`varFlagsMap["aws_profile"] = deployConfig.AwsConfig.Profile`

src/terraform/command_helpers.go
```diff
@@ -14,79 +14,79 @@ import (
 
 // CompileVarFlags compiles the variable flags passed into a Terraform command
 func CompileVarFlags(deployConfig deploy.Config, secrets types.Secrets, imagesMap map[string]string) ([]string, error) {
```
charlierudolph, line 16: For testing purposes, thoughts on having a `GetVarMap()` function which builds the entire var map and this function calls that and converts it to flags. That way we can test that function instead of the individual functions its calls.

src/terraform/command_helpers.go
```diff
@@ -14,79 +14,88 @@ import (
...
+	util.Merge(varMap, secrets)
+	varMap["aws_profile"] = deployConfig.AwsConfig.Profile
+	return varMap, nil
 }
 
-// compile docker image var flags for each service
-func compileDockerImageVars(deployConfig deploy.Config, imagesMap map[string]string) []string {
-	vars := []string{}
+// GetDockerImageVarMap compiles the docker image variables for each service
+func GetDockerImageVarMap(deployConfig deploy.Config, imagesMap map[string]string) map[string]string {
```
charlierudolph: These helpers can be made private again

charlierudolph (approved):

PR #893, opened 2017-12-15
support deploying to multiple environments

### LINKED ISSUE #498: exo deploy: ability to deploy to multiple environments

charlierudolph:
rework of #884

resolves #498

* the remote section in application.yml / service.yml now contains a `environments` key holds all environment specific data. In app its region/account-id/url/ssl-certificate. In service it is url/environment.
* `exo configure` / `exo deploy` have a required argument of the remote-environment-id.
* The terraform state and secrets s3 buckets moved from `<accountId>-<appName>-*` to `<accountId>-<appName>-<remoteEnvironmentId>-*`


src/cmd/shared.go
```diff
@@ -2,23 +2,26 @@ package cmd
...
-		Region:               appConfig.Remote.Region,
-		AccountID:            appConfig.Remote.AccountID,
-		SslCertificateArn:    appConfig.Remote.SslCertificateArn,
+		Region:               appRemoteEnv.Region,
+		AccountID:            appRemoteEnv.AccountID,
+		SslCertificateArn:    appRemoteEnv.SslCertificateArn,
 		Profile:              profile,
-		SecretsBucket:        fmt.Sprintf("%s-%s-terraform-secrets", appConfig.Remote.AccountID, appConfig.Name),
-		TerraformStateBucket: fmt.Sprintf("%s-%s-terraform", appConfig.Remote.AccountID, appConfig.Name),
+		SecretsBucket:        fmt.Sprintf("%s-%s-%s-terraform-secrets", appRemoteEnv.AccountID, appConfig.Name, remoteEnvironmentID),
```
hugoesthere, line 23: I think it might be _slightly_ easier for the user to distinguish state buckets if they were prefixed with the remote environment ID? Since that's the only distinguishing factor between these names. Same for the `TerraformStateBucket`

hugoesthere: We could even consider creating a folder with the environment, i.e. `qa/id-appName-terraform-secrets`. That might provide even further separation, and I think that's what Terraform does when you use their built in environment manager

src/types/app_remote_config_environment.go
```diff
@@ -0,0 +1,28 @@
+package types
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// AppRemoteEnvironment represents represents configuration for a particular environment
+type AppRemoteEnvironment struct {
+	Environment       map[string]string
```
hugoesthere, line 10: I think we should open an issue renaming this to EnvironmentVariables, here and everywhere else we use env vars

src/types/service_remote_config.go
```diff
@@ -8,30 +8,22 @@ import (
...
 type ServiceRemoteConfig struct {
 	Dependencies map[string]RemoteDependency
-	Environment  map[string]string `yaml:",omitempty"`
-	Secrets      []string          `yaml:",omitempty"`
-	URL          string            `yaml:"url,omitempty"`
-	CPU          string            `yaml:"cpu,omitempty"`
-	Memory       string            `yaml:"memory,omitempty"`
+	Environments map[string]ServiceRemoteEnvironment
+	CPU          string `yaml:"cpu,omitempty"`
+	Memory       string `yaml:"memory,omitempty"`
```
hugoesthere, line 13: Wouldn't CPU and Memory be tweaked per environment ideally as well?

src/types/app_remote_config_environment.go
charlierudolph: Sounds good

src/types/service_remote_config.go
charlierudolph: I don't think so since these are relative units.

src/cmd/shared.go
charlierudolph: If we need the account id in the bucket name, I think that should be first. This is going from highest to lowest level. 

An account id could deploy many applications and for each app we could deploy multiple environments.

If we do a folder structure, I think it should follow that as well. Why exactly did we need the account-id in the bucket name again?

src/types/service_remote_config.go
hugoesthere: What do you mean by relative units? You don't need the same amount of cpu/memory in dev as you might production

src/cmd/shared.go
hugoesthere: Because bucket names are global, so account id was an easy way to distinguish.

Ok, I think going with the hierarchical structure is good, but there doesn't seem to be a straightforward way to translate that to folder structure

charlierudolph: Global in what scope?

charlierudolph: Per discussion on standup going with the current setup

src/types/app_remote_config_environment.go
charlierudolph: done

src/types/service_remote_config.go
charlierudolph: One reason I thought this made more sense is that the dependency like rds has the same allocated storage on each env. I'd say lets leave as is and we can make this configurable per environment if needed.

charlierudolph:
@hugobho comments address

hugoesthere (approved):

PR #895, opened 2017-12-15
cleanup `exo configure` functionality

charlierudolph:
read / create / delete / update was already creating the secrets store if it doesn't exist. Renamed to make this easier to see.

### DIFF
--- a/src/aws/secrets_helpers.go
+++ b/src/aws/secrets_helpers.go
@@ -15,9 +15,9 @@ import (
 
 const secretsFile string = "secrets.json"
 
-// GetOrCreateSecrets reads secret key value pair from remote store
+// ReadSecrets reads secret key value pair from remote store
 // It creates an empty secrets store if necessary
-func GetOrCreateSecrets(awsConfig types.AwsConfig) (types.Secrets, error) {
+func ReadSecrets(awsConfig types.AwsConfig) (types.Secrets, error) {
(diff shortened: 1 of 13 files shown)


src/aws/secrets_helpers.go
```diff
@@ -15,14 +15,9 @@ import (
...
-func CreateSecretsStore(awsConfig types.AwsConfig) error {
-	s3client := createS3client(awsConfig)
-	return createS3Object(s3client, strings.NewReader("{}"), awsConfig.SecretsBucket, secretsFile)
-}
-
-// ReadSecrets reads secret key value pair from remote store
-func ReadSecrets(awsConfig types.AwsConfig) (types.Secrets, error) {
+// GetOrCreateSecrets reads secret key value pair from remote store
+// It creates an empty secrets store if necessary
+func GetOrCreateSecrets(awsConfig types.AwsConfig) (types.Secrets, error) {
```
hugoesthere: Can we just keep this as `ReadSecrets`, and either extract a `CreateSecretsStore` function or just make it very clear in the comments that `ReadSecrets` creates the store if it doesn't exist? I think it's actually more confusing to call this `GetOrCreateSecrets`. It kind of sounds like this function could be used for POST-ing a secret as well as GETting one

charlierudolph:
@hugobho comments addressed

hugoesthere (approved):

Issue #892, opened 2017-12-15
supporting multiple remotes: reduce terraform noise


charlierudolph:
In order to support multiple remotes we some more things to be variables in the terraform file if we want to have only one terraform file:

* region needs to become a variable
* account-id needs to become a variable
* s3-bucket needs to become a variable 
  * also needs to be updated to include the remote environment id
* external dns for the application needs to become a variable
* ssl-certificate-arn needs to become a variable

Probably also want to make  "env"  a variable. It is currently hardcoded as production and could be equal to the remote environment id.

@hugobho @alexdavid thoughts?

PR #898, opened 2017-12-15
[terraform variables 3] aws ssl certificate arn

charlierudolph:
part of #892


hugoesthere (approved):
Have you tried a dry run of `exo deploy` will all these new changes? I know when I tried to pass the aws profile into the backend_config block, terraform threw an error because it couldn't start the command if it didn't know which aws profile to use (?). Instead I had to use a special flag `-backend-config` instead of the regular `-var` flag. But that may not be the case here

charlierudolph:
Ran into the error 
```
Error loading backend config: 1 error(s) occurred:

* terraform.backend: configuration cannot contain interpolations

The backend configuration is loaded by Terraform extremely early, before
the core of Terraform can be initialized. This is necessary because the backend
dictates the behavior of that core. The core is what handles interpolation
processing. Because of this, interpolations cannot be used in backend
configuration.

If you'd like to parameterize backend configuration, we recommend using
partial configuration with the "-backend-config" flag to "terraform init".
```
So will try moving the config to the backend-config

hugoesthere (approved):

Issue #901, opened 2017-12-15
terraform noise: dependencies


charlierudolph:
The order of the dependencies in the terraform file is not deterministic. Ensure we sort the dependency keys first and iterate through them that way.

PR #904, opened 2017-12-15
sort terraform dependencies

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/901


src/config/config_helpers.go
```diff
@@ -55,3 +56,18 @@ func GetAllRemoteDependencies(appContext *context.AppContext) map[string]types.R
...
+		result = append(result, k)
+	}
+	for _, serviceContext := range appContext.ServiceContexts {
+		for k := range serviceContext.Config.Remote.Dependencies {
+			result = append(result, k)
+		}
+	}
+	sort.Strings(result)
+	return result
+}
```
charlierudolph: I think this should be a method on `appContext` since that is the only argument.

charlierudolph (approved):
Good to merge after moving the method

src/config/config_helpers.go
hugoesthere: There are also a number of other methods that only rely on `appContext`. Should we move those as well? I think we need to also redefine what this `config` package is used for/what we should name it. I don't think it serves the purpose of config helpers anymore and `config` always felt like a vague package name to me anyways

charlierudolph: For now, can we move just this method and then refactor the config package in another PR. I'll create an issue where we can discuss where to move the methods

PR #906, opened 2017-12-15
cleanup tests

charlierudolph:
Sorry missed this in review. Checkout http://onsi.github.io/gomega/#consistofelement-interface. We should be using gomega matchers as much as we can and do more custom matchers to make sure we are getting good errors messages. When checking a complex thing, we don't want the error message to say "expected true to be false".


hugoesthere (approved):
Didn't you could do this with nested objects 👍

PR #913, opened 2017-12-16
remote dependency templates: update docs

charlierudolph:
Added some formatting, tried to make things a bit easier to understand and document the why.

### DIFF
--- a/remote-dependency-templates/rds/README.md
+++ b/remote-dependency-templates/rds/README.md
@@ -34,7 +34,7 @@ remote:
 ```
 
 ## Environment variables
-The database name and username should be remote environment variables so that the RDS instance can be accessed. Name them as you choose but ensure they match the values in the template config.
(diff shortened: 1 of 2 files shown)


remote-dependency-templates/rds/README.md
```diff
@@ -35,26 +33,25 @@ remote:
         password-secret-name: POSTGRES_PASSWORD
 ```
 
-Environment variables:
-- `DB_NAME`: environment variable for db name, must be the same as the one listed in `<dependency-id>.db-name`
-- `DB_USERNAME` environment variable for db username, must be the same as the one listed in `<dependency-id>.username`
+## Environment variables
+The database name and username should be remote environment variables so that the RDS instance can be accessed. Name them as you choose but ensure they match the values in the template config.
```
hugoesthere: Maybe: "Name them as you choose but ensure their values match those defined in the template config."

hugoesthere (approved):

Issue #902, opened 2017-12-15
cleanup: exo deploy output


charlierudolph:
When pushing an image the command and the first line are joined together:
```
docker push 855366641436.dkr.ecr.us-west-2.amazonaws.com/edx_security:016bdde5fb1842489edf9aa02051c3ea7475ae27The push refers to a repository [855366641436.dkr.ecr.us-west-2.amazonaws.com/edx_security]: 
Preparing: 11804e94519e
Preparing: 678428ce4269
Preparing: 915ecf7beb8e
Preparing: b6d3b4c8edcc
Preparing: 7f9dddb59a03
Preparing: c38858d70938
Preparing: 7df250eb9e5c
Preparing: e9591a63b7e4
Preparing: d752a0310ee4
Preparing: db64edce4b5b
Preparing: d5d60fc34309
Preparing: c01c63c6823d
```

PR #917, opened 2017-12-18
cleanup exo-deploy output

hugoesthere:
Resolving https://github.com/Originate/exosphere/issues/902

### DIFF
--- a/src/util/print.go
+++ b/src/util/print.go
@@ -25,6 +25,7 @@ func PrintSectionHeader(writer io.Writer, text string) {
 	if err != nil {
 		panic(err)
 	}
+	fmt.Println("")
 }
 
 // PrintCommandFooter prints the time elapsed since startTime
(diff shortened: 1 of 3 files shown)


charlierudolph:
I think `PrintSectionHeader` should be updated to print a trailing newline. Then any calls can remove the added newline. Also the call you updated can be updated to use `PrintSectionHeaderf`

src/util/print.go
```diff
@@ -25,6 +25,7 @@ func PrintSectionHeader(writer io.Writer, text string) {
 	if err != nil {
 		panic(err)
 	}
+	fmt.Println("")
```
charlierudolph: I think we could change line 24 to use `Fprintln`

charlierudolph:
If you search for `PrintSectionHeader` there are a few other places where its used an a trailing newline is passed in. We should remove those now. Appears to only be in `src/application/tester/tester.go`

charlierudolph (approved):

PR #920, opened 2017-12-19
update rds dependency

hugoesthere:


charlierudolph (approved):
Please search for the other uses of this as it appears we have some tests that involve it. Merge on green

hugoesthere:
I don't think the tests ever got to testing these cases (they fail when earlier fields aren't listed).

PR #924, opened 2017-12-19
cleanup unneeded build modes

charlierudolph:
We had some build modes left over that we were only using for getting the docker compose filenames. Since the building is all done by `exo generate`, we didn't need these anymore and could use the filenames directly instead.

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -26,7 +26,7 @@ var runCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
-		dockerComposeFileName := types.LocalProductionComposeFileName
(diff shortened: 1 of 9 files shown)


src/cmd/run.go
```diff
@@ -26,17 +26,13 @@ var runCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
-		buildMode := types.BuildMode{
-			Type:        types.BuildModeTypeLocal,
-			Mount:       true,
```
charlierudolph, line 31: This confused me as I was trying to figure out if we mounted in production mode and this code made it look like we did.

hugoesthere (approved):

Issue #908, opened 2017-12-16
terraform apply: command is too long


charlierudolph:
Currently have a command that is way to long as the var flags are just massive. Was there a reason we decided against the var file? I think we had that originally and think that would be nicer for someone to look at. We could write it in the terraform dir and have that gitignored automatically.

hugoesthere:
I can't remember exactly why we decided against a var file. I'm not opposed to using one again

charlierudolph:
K. It appears we were initially using the secrets as a tf vars file and then switched it to json. I think this would be a really nice switch. Lets try it out.

PR #925, opened 2017-12-19
exo-deploy: use terraform var file

hugoesthere:
resolving https://github.com/Originate/exosphere/issues/908


src/terraform/command_helpers.go
```diff
@@ -12,13 +12,17 @@ import (
...
-func CompileVarFlags(deployConfig deploy.Config, secrets types.Secrets, imagesMap map[string]string) ([]string, error) {
+// GenerateVarFile compiles the variable flags passed into a Terraform command
+func GenerateVarFile(deployConfig deploy.Config, secrets types.Secrets, imagesMap map[string]string) error {
 	varMap, err := GetVarMap(deployConfig, secrets, imagesMap)
 	if err != nil {
-		return []string{}, err
+		return err
 	}
-	return buildFlags(varMap), nil
+	jsonVarMap, err := json.Marshal(varMap)
```
charlierudolph: Please use `MashalIndent` in order to make the file easier for a human to read.

charlierudolph (approved):

src/terraform/command_helpers.go
```diff
@@ -12,13 +12,17 @@ import (
...
-func CompileVarFlags(deployConfig deploy.Config, secrets types.Secrets, imagesMap map[string]string) ([]string, error) {
+// GenerateVarFile compiles the variable flags passed into a Terraform command
+func GenerateVarFile(deployConfig deploy.Config, secrets types.Secrets, imagesMap map[string]string) error {
 	varMap, err := GetVarMap(deployConfig, secrets, imagesMap)
 	if err != nil {
-		return []string{}, err
+		return err
 	}
-	return buildFlags(varMap), nil
+	jsonVarMap, err := json.MarshalIndent(varMap, "", "    ")
```
charlierudolph: Thoughts on 2 spaces instead of 4?

src/terraform/template_builder.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 const terraformFile = "main.tf"
+const terraformVarFile = "terraform.tfvars"
```
charlierudolph: Thoughts on this being `main.tfvars`? I think `terraform.tfvars` feels weirdly redundant. In supporting multiple remotes this could become `<remote-environment-id>.tfvars` so you can see the different var files per environment.

hugoesthere: The name comes from the default file name that terraform uses: https://www.terraform.io/intro/getting-started/variables.html#from-a-file. In fact we could probably do without the `-var-file` flag but I added it just to be specific. I'm also not sure about naming a var file by remote environment id, being that these files are auto-generated and not meant to be checked in. I had thought about attaching the var file generation to `exo generate terraform`, but that would mean checking if the var file was up-to-date, and again that seemed weird seeing as these files were gitignores. But maybe that isn't. What do you think?

charlierudolph: I am expecting this to always be git-ignored. It was solely if a user deploys a couple environments locally and can thus look at two the env vars file to see how they are different. I do not think we need generate it as part of `exo generate terraform`.

hugoesthere: One problem with naming it `<remote-environment-id>.tfvars` is that we'll have to add it to the `.gitignore` each time the user adds/renames an environment, which means we'd have to check the `.gitignore` each time the user runs `exo deploy`, etc. If a user renames an environment, we could be left with "dangling" files in the `.gitignore`.

My question is what is in the `terraform.tfvars` file that isn't already defined in `application.yml` and `service.yml`? Environment variables are already explicitly listed per environment in the yaml files. Secrets can be read in `exo configure read`. It shouldn't be a mystery what is being passed to Terraform, and if it is then we should make it more transparent what we're doing. If it comes down to it, a user can always run `exo deploy`, and check the `.tfvars` file before confirming "yes" with the prompt

charlierudolph: In the gitignore we would add `*.tfvars` to ensure it would always be gitignored no matter the environment.

I believe the current process is transparent and the user should not be surprised by anything thats in there but I think the environment id is a better name for the file then the generic terraform. I also think it shows better that we are reusing the terraform file but using different variables in the different environments by storing them in different files.

charlierudolph: I also believe using the environment in the name makes it more consistent with existing workflows (and thus easier to transition) which has a set of terraform modules and then one var file per environment.

hugoesthere: Ok, I guess this wasn't as cumbersome as i'd imagined. Comments addressed

charlierudolph (approved):
Merge on green

PR #927, opened 2017-12-20
Refactor config package

### LINKED ISSUE #915: refactor: remove config package
* `GetBuiltLocalAppDependencies`, `GetBuiltLocalServiceDependencies`, `LocalAppDependency` could all live in `docker/composebuilder` since that is the only place they are used
* `GetAllRemoteDependencies`,  `UpdateAppConfig` could become methods on `AppContext`

@hugobho thoughts?

hugoesthere:
resolves https://github.com/Originate/exosphere/issues/915


src/types/context/app_context.go
```diff
@@ -24,8 +24,8 @@ func NewAppContext(location string, config types.AppConfig) (*AppContext, error)
 	return appContext, appContext.initializeServiceContexts()
 }
 
-// GetServiceContextByLocation returns the service context for the given location
-func (a *AppContext) GetServiceContextByLocation(location string) *ServiceContext {
+// GetServiceContext returns the service context for the given location
```
charlierudolph: I don't agree with this rename as we often get the service context by the role and thus by location helps differentiate it.

src/types/app_config.go
```diff
@@ -39,6 +39,20 @@ func NewAppConfig(appDir string) (result AppConfig, err error) {
 	return result, result.validateAppConfig()
 }
 
+// UpdateAppConfig adds serviceRole to the appConfig object and updates
+// application.yml
+func (a AppConfig) UpdateAppConfig(appDir string, serviceRole string) error {
```
charlierudolph: How about renaming this to `AddServiceAndSave`?

src/docker/composebuilder/localdependencies/helpers.go
```diff
@@ -0,0 +1,26 @@
+package localdependencies
```
charlierudolph: Does this need to be in its own package or can it live in the `composebuilder` package? It it doesn't need to be in its own package I'd prefer it lives in `composebuilder`

hugoesthere: It doesn't need to be, but it seemed like we're conflating too many things in one package. Just because `composebuilder` is the only package that uses this local dependency stuff doesn't mean they should be lumped into the same package. I would even be open to extracting a `dependency` package and putting `localdependencies` and `remotedependencies` in there.

src/types/app_config.go
hugoesthere: Or even just `AddService`? I'm not a big fan of these method names being verbose/revealing implementation details, it makes it harder to keep track of. That's what I was trying to simplify in this refactor.

src/types/context/app_context.go
hugoesthere: I see that we're getting service context by role in the private method `AppContext.getServiceContext()` (which perhaps we can rename to `getServiceContextByRole`, but where else are we doing that? It seems like `GetAppContext` uses location, so it would make sense that `GetServiceContext` does as well. I'm just trying to make the public functions as simple and sensible as possible.

src/types/app_config.go
charlierudolph: Thoughts on breaking this into two functions `AddService` which takes in a serviceRole and updates the config. And `Write` which takes in an appDir and write the appConfig to disk?

src/types/context/app_context.go
charlierudolph: `AppContext` has a `ServiceContexts` property which is a map from role to `ServiceContext`

src/docker/composebuilder/localdependencies/helpers.go
charlierudolph: I think this `LocalDependency` strut should actually be removed in favor a `DependencyBuilder` struct. We don't need a local/remote prefix on dependencies as `composebuilder` is only concerned with local

Within the `types` package, we could move LocalDependency and RemoteDependency into a dependencies subpage

src/types/app_config.go
hugoesthere: Hm..now that I'm actually implementing that, two functions might be more complexity than is needed. First, neither of these functions are used anywhere else. Second, you wouldn't add a service to the AppConfig struct without writing that to app.yml, and conversely you wouldn't write anything to app.yml without having updated the AppConfig struct.

I extracted a `write` function, but I kept it private and I just had `AddService` call that function. I'll push my changes and you can let me know how that feels

src/types/context/app_context.go
```diff
@@ -48,6 +80,23 @@ func (a *AppContext) GetDockerComposeDir() string {
 	return path.Join(a.Location, "docker-compose")
 }
 
+// AddService adds serviceRole to the appConfig object
+func (a *AppContext) AddService(serviceRole string) {
+	if a.Config.Services == nil {
+		a.Config.Services = map[string]types.ServiceSource{}
+	}
+	a.Config.Services[serviceRole] = types.ServiceSource{Location: fmt.Sprintf("./%s", serviceRole)}
+}
```
charlierudolph: I think this method still belongs on the `AppConfig`. I was thinking we would have the following:
```golang
func (a *AppContext) AddService(serviceRole string) error {
  a.Config.AddService(serviceRole)
  return a.Config.Write(a.Location)
}
```
I liked `Write` being on `AppConfig` as that keeps the read and write of the `AppConfig` in one file.

charlierudolph (approved):

hugoesthere:
@charlierudolph I switched to the method that you suggested, but I'm having trouble getting tests to pass for some reason. The feature test `add/add-service.feature` seems to behave differently than when I cd into the `tmp` dir and run `exo add` with the same exact arguments. In the first case, the `services` block seems to get deleted comepletely.

src/types/app_config.go
```diff
@@ -39,6 +39,14 @@ func NewAppConfig(appDir string) (result AppConfig, err error) {
 	return result, result.validateAppConfig()
 }
 
+// AddService adds serviceRole to the appConfig object
+func (a AppConfig) AddService(serviceRole string) {
```
charlierudolph: This doesn't update the config because this function is passed a copy instead of a reference. We should update all of these to be pointer receivers and have NewAppConfig return a pointer. Please do that in a separate PR that this one depends on.

https://tour.golang.org/methods/4

PR #931, opened 2017-12-21
WIP:  remote dependency plugin documentation

hugoesthere:
A rough draft resolving https://github.com/Originate/exosphere/issues/914, could use some refining. Let me know if this is what you had in mind.

PR #935, opened 2017-12-23
move rds template

charlierudolph:
to conform with documentation in #931


hugoesthere (commented):
What's wrong with keeping them in `remote-dependencies/<dependency-type>/modules`?

charlierudolph:
You mean leave the two directly in `modules` and not put them in a subfolder? That seems weird to me to have a single module in a `modules` folder. This also follows the pattern in exocom as well since it has two sub-folders

hugoesthere:
Ah ok, I took a closer look at the file structure and I see what you're doing. Looks good.

hugoesthere (approved):

PR #941, opened 2017-12-28
update exocom dependency

### LINKED ISSUE #937: exocom dependency: cluster name
The cluster name is currently `exocom`. Update it to be `<var.env>-exocom` in order to support multiple deployments on the same aws account.

charlierudolph:
Add support for multiple deploys in a single account

resolves #937

### DIFF
--- a/remote-dependency-templates/exocom/modules/exocom-cluster/cluster.tf
+++ b/remote-dependency-templates/exocom/modules/exocom-cluster/cluster.tf
@@ -1,5 +1,5 @@
 resource "aws_ecs_cluster" "exocom" {
-  name = "exocom"
+  name = "${var.env}-exocom"
 
 }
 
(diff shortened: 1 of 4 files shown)


hugoesthere (approved):
This might be useful for labeling exocom resources by environment, but it doesn't allow us to deploy multiple environments on a single account. We'd have the same issue with limited elastic IPs as we did when we tried to use Terraform's built-in environment manager. The same goes for all the other prefixed aws resourses - useful for identification but doesn't actually allow for multiple deployments on a single account.

charlierudolph:
On titan, we are deploying 4 environments in a single aws account (each cluster is prefixed with a different env). Each is available as a subdomain of "internal.xxx.com"

hugoesthere:
Just something to bring up, Terraform has something similar built-in: https://www.terraform.io/docs/state/workspaces.html
 It requires switching workspaces each time your deploy (which might actually be a better way of switching environments than doing `exo deploy qa`). But it might also be more trouble than it's worth. All it does is provide the ability to prefix your resources (which we do manually), and creates different state files, which we do by creating separate buckets. Just something to think about.

charlierudolph:
Looking at the documentation on workspaces, I don't see that it does the prefixing of resources for you. Instead it appears we can use the workspace in the names of resources.

hugoesthere:
Yeah that's what I meant. Not too different from what we're doing, just that it's provided by terraform

PR #942, opened 2017-12-28
internal network name includes environment

### LINKED ISSUE #938: route 53: internal network
Update the name of the internal network to include the environment in order to support multiple environments in the same AWS account.

charlierudolph:
In support of deploying multiple environments on the same account

resolves #938

### DIFF
--- a/remote-dependency-templates/exocom/README.md
+++ b/remote-dependency-templates/exocom/README.md
@@ -20,14 +20,15 @@ remote:
 ```
 
 ## Environment variables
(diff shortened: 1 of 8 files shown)


hugoesthere (commented):
This will require some more changes/testing. For instance a lot of the production endpoints will have to change. I'm also not sure this change is necessary/worth it since this only affects internal routing, and again since we can't deploy onto the same account, then having the same internal routing dns name on different accounts doesn't make a practical difference.

charlierudolph:
@hugobho service endpoints updated

charlierudolph:
@hugobho so is this good to merge or is there more you wanted to discuss?

hugoesthere (approved):

Issue #869, opened 2017-12-12
worker service can expose a port


charlierudolph:
Extracted from #685

The env var passed to other services would be

---
### `<SERVICE>_HOST`
<table>
<tr><th>Local development value</th><td>ftp-api:2121</td></tr>
<tr><th>Local production value</th><td>ftp-api:21</td></tr>
<tr><th>AWS production value</th><td>ftp-api.local:65000</td></tr>
</table>
Useful for services that need to communicate directly with non-web services. note that for worker services we need to have them query the ec2metadata api and update their route53 record to match its scheduled EC2 box.

charlierudolph:
Requires putting a script on each docker image

charlierudolph:
Sample terraform: we would probably update the name and role to what we prefer
```tf
resource "aws_iam_role_policy" "allow_dns_changes" {
  name = "ecs-allow-dns-changes-dbp-${module.stack.environment}"
  role = "${element(split("/", module.stack.iam_role), 1)}"

  policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "route53:ChangeResourceRecordSets"
      ],
      "Resource": "*"
    }
  ]
}
EOF
}
```
Sample script: we would probably make the whole name an environment variable and thus would need a couple env variables, AWS_INTERNAL_ZONE_ID and SELF_HOSTNAME.
```sh
#!/usr/bin/env bash

set -euo pipefail

INSTANCE_IP="$(curl -fsSL http://169.254.169.254/latest/meta-data/local-ipv4)"

aws route53 change-resource-record-sets --hosted-zone-id "$INTERNAL_ZONE_ID" --change-batch '{
  "Comment": "Update nats record to point to this container",
  "Changes": [
      {
          "Action": "UPSERT",
          "ResourceRecordSet": {
              "Name": "nats.'$INTERNAL_ZONE_NAME'.",
              "Type": "A",
              "TTL": 30,
              "ResourceRecords": [
                  {
                      "Value": "'$INSTANCE_IP'"
                  }
              ]
          }
      }
  ]
}'
```
This script requires curl / awscli to be installed on the machine.

So when a worker service exposes a port we automatically add the necessary terraform. Adding the script may be a manual thing if we need to add it to the production docker image, but would need some way to disable it locally. Could have a command that generates the script for them and then they update their dockerfile with all the needed stuff.

[...]

PR #945, opened 2018-01-04
Route53 updater

hugoesthere:
Resolves: https://github.com/Originate/exosphere/issues/869
See: https://github.com/Originate/space-tweet/pull/43 for an example
 
Tested and works:
screen shot 2018-01-04 at 2 15 34 pm


resources/update-route53/main.go
```diff
@@ -0,0 +1,67 @@
...
+	"os/exec"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/route53"
+)
+
+func main() {
+	if len(os.Args) < 3 {
+		panic(errors.New("Not enough arguments. Arguments must be non-empty strings passed into 'route53-updater <service-role> <internal-hosted-zone-name>'"))
```
charlierudolph: Lets use `log.Fataln` instead of `panic` as we don't need to have the stack trace printed

charlierudolph (approved):
Looks great to me. I added in #946 so we can add some documentation for how to use this. Please wait for a review from @alexdavid

resources/update-route53/main.go
```diff
@@ -0,0 +1,67 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/route53"
+)
+
+func main() {
+	if len(os.Args) < 3 {
```
charlierudolph: How about checking `if len(os.Args) != 3`?

terraform/aws/ecs-cluster/iam.tf
```diff
@@ -43,7 +43,9 @@ resource "aws_iam_role_policy" "ecs_instance" {
         "ecr:GetDownloadUrlForLayer",
         "ecr:BatchGetImage",
         "ecs:StartTask",
-        "autoscaling:*"
+        "autoscaling:*",
+        "route53:ChangeResourceRecordSets",
```
alexdavid: I think we should restrict this to only allow changes on the internal hosted zone

resources/update-route53/main.go
```diff
@@ -0,0 +1,68 @@
...
+func main() {
+	if len(os.Args) != 3 {
+		log.Fatalln(errors.New("Wrong number of arguments. Arguments must be non-empty strings passed into 'route53-updater <service-role> <internal-hosted-zone-name>'"))
+	}
+	serviceRole := os.Args[1]
+	internalHostedZoneName := os.Args[2]
+	if serviceRole == "" || internalHostedZoneName == "" {
+		log.Fatalln(errors.New("Service role or internal hosted zone name missing. Both arguments must be non-empty strings passed into 'route53-updater <service-role> <internal-hosted-zone-name>'"))
+	}
+	internalIP, err := exec.Command("curl", "-fsSL", "http://169.254.169.254/latest/meta-data/local-ipv4").Output()
```
alexdavid: does it make sense to run this as an external command or just make an http request here?

alexdavid (approved):

terraform/aws/ecs-cluster/iam.tf
hugoesthere: How would we restrict it to only the internal hosted zone in aim roles?

PR #949, opened 2018-01-05
cleanup configuration files

charlierudolph:


src/docker/composebuilder/dependency_builder.go
```diff
@@ -50,14 +50,13 @@ func (g *LocalDependency) GetDockerConfig() (types.DockerConfig, error) {
 func (g *LocalDependency) GetServiceEnvVariables() map[string]string {
 	result := map[string]string{}
 	result[fmt.Sprintf("%s_HOST", strings.ToUpper(g.name))] = g.name
-	util.Merge(result, g.config.Config.ServiceEnvironment)
```
charlierudolph: Remove `ServiceEnvironment` as we can use app wide environment variables to achieve the same thing

src/types/local_dependency_config.go
```diff
@@ -1,8 +0,0 @@
-package types
-
-// LocalDependencyConfig represents the configuration of a local dependency
-type LocalDependencyConfig struct {
-	Persist               []string          `yaml:",omitempty"`
-	DependencyEnvironment map[string]string `yaml:"dependency-environment,omitempty"`
-	ServiceEnvironment    map[string]string `yaml:"service-environment,omitempty"`
```
charlierudolph, line 7: De-nested these from under Config with the given changes

* Renamed `DependencyEnvironment` to `EnvironmentVariables` for passing only to the dependency
* Added `Secrets` to be consistent with other passing of environment variables
* Removed `ServiceEnvironment` as we can use app wide environment variables to achieve the same thing

src/types/remote_dependency.go
```diff
@@ -9,7 +9,6 @@ import (
 
 // RemoteDependency represents a production dependency
 type RemoteDependency struct {
-	Config         RemoteDependencyConfig `yaml:",omitempty"`
```
charlierudolph, line 12: Removed this as it is no longer used as we are using TemplateConfig now

src/types/service_config.go
```diff
@@ -21,10 +21,7 @@ const ServiceTypeWorker = "worker"
 // service.yml
 type ServiceConfig struct {
 	Type           string                   `yaml:",omitempty"`
-	Description    string                   `yaml:",omitempty"`
-	Author         string                   `yaml:",omitempty"`
 	DependencyData ServiceDependencyData    `yaml:"dependency-data,omitempty"`
-	Docker         DockerConfig             `yaml:",omitempty"`
```
charlierudolph, line 27: Removed these fields that are not used

hugoesthere (approved):

PR #950, opened 2018-01-05
document `application.yml`

charlierudolph:


documentation/configuration_files/application.md
```diff
@@ -0,0 +1,38 @@
...
+# configuration for locally running the application
+local:
+  dependencies:
+    <id>: # each dependency needs an id
+      image: # docker image
+      environment-variables: # map of environment variables to pass to this dependency
+      secrets: # array of shell environment variables to pass through to this dependency
+      persist: # array of paths within the docker images you would like saved in a docker volume
+  environment-variables: # map of environment variables to pass to every service  
+  secrets: # array of shell environment variables to pass to all services
```
hugoesthere: Can we move `environment-variables` and `secrets` up before `dependencies`? Even though there's a `dependency.<id>.environment-variables` section it's a little unclear that the env vars are for the application and not the dependencies.

documentation/configuration_files/application.md
```diff
@@ -0,0 +1,38 @@
...
+      url: # url for the application
+      ssl-certificate-arn: # aws ssl cetificate arn
+      environment-variables: # map of environment variables to pass to every service
+      secrets: # array of secret keys variables to pass to every service (see `exo configure`)
+
+# map of services, used locally and in deployments
+services:
+  <role>: # each service has its own role
+    location: # relative path to the service source (use this or `docker-image`)
+    docker-image: # docker image of the service source (use this or `location`)
```
hugoesthere: We should specify that only one of these fields can be populated

documentation/configuration_files/application.md
charlierudolph: I don't think its unclear as its not nested under dependencies. The comment should make it clear too as it says "to pass to every service".

documentation/configuration_files/application.md
```diff
@@ -0,0 +1,39 @@
...
+      persist: # array of paths within the docker image you would like saved to docker volumes
+  environment-variables: # map of environment variables to pass to every service  
+  secrets: # array of shell environment variables to pass to every services
+
+# configuration for deployments of the application
+remote:
+  dependencies:
+    <id>: # each dependency needs an id
+      type: # dependency type (exocom or rds)
+      template-config: # map of data to pass to the dependency template
```
hugoesthere: Could we add some links to the remote dependency readme like you did in the service.yml PR?

hugoesthere (approved):

Issue #946, opened 2018-01-05
document service types


charlierudolph:
Maybe in `documentation/services.md` add some documentation about the different service types and how the resulting endpoint variables that will be available. For worker services, document how to use the update-route53 binary.

PR #967, opened 2018-01-09
document service types

hugoesthere:
https://github.com/Originate/exosphere/issues/946

### DIFF
--- a/documentation/configuration_files/service.md
+++ b/documentation/configuration_files/service.md
@@ -44,7 +44,7 @@ dependency-data:
 
 # Service types
 
-## Public services
+## Public
 Public services expose an external and internal http endpoint. Container ports must be listed in `development.port` and `production.port`.
 
 Endpoints available in local development:
@@ -56,21 +56,21 @@ Endpoints availabe in deployment:
   - `#{SERVICE_ROLE}_INTERNAL_ORIGIN`: The internal http endpoint at which a service can be reached. Terraform manages creation of these records in route53. Used for internal communication with other services.
 
 
-## Worker services
+## Worker
 Worker services have the option of exposing an internal non-http, non-load-balanced endpoint. If a worker service exposes a port, the container port must be listed in `development.port` and `production.port`.
 Additionally, the proper `update-route53` binary must be downloaded from Exosphere's [release](https://github.com/Originate/exosphere/releases), copied into the worker service's Docker container and ran as part of that service's start-up script.
-For example, a start script could look like the following:
+It only needs to be copied over in the `Dockerfile.prod` file. An example start script could look like the following:
 ```
 #!/usr/bin/env bash
 set -e
 
 update-route53 $ROLE $INTERNAL_HOSTED_ZONE_NAME
-node src/server.ls
+node src/server.js
 ```
 `$ROLE` is passed into every service automatically, but `$INTERNAL_HOSTED_ZONE_NAME` must be set by the user as an environment variable in each remote environment block, and they must be of the form `#{service-role}.#{remote-environment-id}-#{app-name}.local`.
 
 Endpoints available in local development:
-  - `#{SERVICE_ROLE}_HOST: #{service-role}.#{container-port}`: The internal endpoint at which a service can be reached
+  - `#{SERVICE_ROLE}_HOST`: The internal endpoint at which a service can be reached
 
 Endpoints available in deployment:
-  - `#{SERVICE_ROLE}_HOST: #{service-role}.#{remote-environment-id}-#{app-name}.local`: The internal endpoint at which a service can be reached
+  - `#{SERVICE_ROLE}_HOST`: The internal endpoint at which a service can be reached
--- a/documentation/configuration_files/service.md
+++ b/documentation/configuration_files/service.md
@@ -41,3 +41,36 @@ dependency-data:
 
 * `remote.cpu`: Number of CPU units to reserve for service container (see "cpu" under [Container Definitions/Environment](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_environment))
 * `remote.memeory`: The hard limit (in MiB) of memory to allocate to the service container (see "memory" under [Container Definitions](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definitions))
+
+# Service types
+
+## Public services
+Public services expose an external and internal http endpoint. Container ports must be listed in `development.port` and `production.port`.
+
+Endpoints available in local development:
+  - `#{SERVICE_ROLE}_EXTERNAL_ORIGIN: http://localhost:#{host-port}`: The external endpoint at which a public service can be reached. Exosphere automatically picks an available host port and binds it to the specified service container port.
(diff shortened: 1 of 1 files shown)


documentation/configuration_files/service.md
```diff
@@ -41,3 +41,36 @@ dependency-data:
...
 * `remote.cpu`: Number of CPU units to reserve for service container (see "cpu" under [Container Definitions/Environment](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_environment))
 * `remote.memeory`: The hard limit (in MiB) of memory to allocate to the service container (see "memory" under [Container Definitions](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definitions))
+
+# Service types
+
+## Public services
+Public services expose an external and internal http endpoint. Container ports must be listed in `development.port` and `production.port`.
+
+Endpoints available in local development:
+  - `#{SERVICE_ROLE}_EXTERNAL_ORIGIN: http://localhost:#{host-port}`: The external endpoint at which a public service can be reached. Exosphere automatically picks an available host port and binds it to the specified service container port.
```
charlierudolph: Thoughts on updating these to something like 

> `#{SERVICE_ROLE}_EXTERNAL_ORIGIN`: The external endpoint ... 

I don't think we need to give the user the implementation details here. We should probably mention all of there *_ORIGIN include the http or https protocol but I don't think we need anything beyond that. They can look at the docker-compose file or the terraform var files to see the values

documentation/configuration_files/service.md
```diff
@@ -41,3 +41,36 @@ dependency-data:
...
+  - `#{SERVICE_ROLE}_INTERNAL_ORIGIN: http://#{service-role}:#{container-port}`: The internal endpoint at which a service can be reached. Used for internal communication with other services.
+
+Endpoints availabe in deployment:
+  - `#{SERVICE_ROLE}_EXTERNAL_ORIGIN: http://#{service-url}`: The load-balanced external endpoint, as defined by the URL listed in the `remote.envorinments.#{remote-environment-id}` block of `service.yml`. Terraform manages creation of these records in route53.
+  - `#{SERVICE_ROLE}_INTERNAL_ORIGIN: http://#{service-role}.#{remote-environment-id}-#{app-name}.local`: The internal endpoint at which a service can be reached. Terraform manages creation of these records in route53. Used for internal communication with other services.
+
+
+## Worker services
+Worker services have the option of exposing an internal non-http, non-load-balanced endpoint. If a worker port is exposed, they must be listed in `development.port` and `production.port`.
```
charlierudolph: Thoughts on updating 

> If a worker port is exposed, they must be listed in `development.port` and `production.port`.

to 

> If a worker service exposes a port, the container port must be listed in `development.port` and `production.port`.

documentation/configuration_files/service.md
```diff
@@ -41,3 +41,36 @@ dependency-data:
...
+Additionally, the proper `update-route53` binary must be downloaded from Exosphere's [release](https://github.com/Originate/exosphere/releases), copied into the worker service's Docker container and ran as part of that service's start-up script.
```
charlierudolph: I think we should build something into the binary so `update-route53` does nothing when the application is being running locally otherwise I expect this to break things if running the production image locally

hugoesthere: What I can think of is passing an environment variable `$ENV` into the binary that determines whether it actually executes. Do you have any ideas?

charlierudolph: How about something like `EXOSPHERE_ENV` = `local` as an environment variable passed to all services in the production docker-compose file. The binary will inherit all the env variables or the running process and when `EXOSPHERE_ENV ` is local it does nothing.

documentation/configuration_files/service.md
```diff
@@ -41,3 +41,36 @@ dependency-data:
...
+#!/usr/bin/env bash
+set -e
+
+update-route53 $ROLE $INTERNAL_HOSTED_ZONE_NAME
+node src/server.ls
+```
+`$ROLE` is passed into every service automatically, but `$INTERNAL_HOSTED_ZONE_NAME` must be set by the user as an environment variable in each remote environment block, and they must be of the form `#{service-role}.#{remote-environment-id}-#{app-name}.local`.
+
+Endpoints available in local development:
+  - `#{SERVICE_ROLE}_HOST: #{service-role}.#{container-port}`: The internal endpoint at which a service can be reached
```
charlierudolph: Here we can also remove the sample values.

documentation/configuration_files/service.md
```diff
@@ -41,3 +41,36 @@ dependency-data:
...
+  - `#{SERVICE_ROLE}_INTERNAL_ORIGIN`: The internal http endpoint at which a service can be reached. Used for internal communication with other services.
+
+Endpoints availabe in deployment:
+  - `#{SERVICE_ROLE}_EXTERNAL_ORIGIN`: The load-balanced external https endpoint, as defined by the URL listed in the `remote.envorinments.#{remote-environment-id}` block of `service.yml`. Terraform manages creation of these records in route53.
+  - `#{SERVICE_ROLE}_INTERNAL_ORIGIN`: The internal http endpoint at which a service can be reached. Terraform manages creation of these records in route53. Used for internal communication with other services.
+
+
+## Worker services
+Worker services have the option of exposing an internal non-http, non-load-balanced endpoint. If a worker service exposes a port, the container port must be listed in `development.port` and `production.port`.
+Additionally, the proper `update-route53` binary must be downloaded from Exosphere's [release](https://github.com/Originate/exosphere/releases), copied into the worker service's Docker container and ran as part of that service's start-up script.
```
charlierudolph: We should mention that this is only needed for the production dockerfile

documentation/configuration_files/service.md
```diff
@@ -41,3 +41,36 @@ dependency-data:
...
+For example, a start script could look like the following:
+```
+#!/usr/bin/env bash
+set -e
+
+update-route53 $ROLE $INTERNAL_HOSTED_ZONE_NAME
+node src/server.ls
```
charlierudolph: Lets change the extension from `ls` to `js` as we don't really use livescript

documentation/configuration_files/service.md
```diff
@@ -41,3 +41,36 @@ dependency-data:
...
+Endpoints available in local development:
+  - `#{SERVICE_ROLE}_EXTERNAL_ORIGIN`: The external http endpoint at which a public service can be reached. Exosphere automatically picks an available host port and binds it to the specified service container port.
+  - `#{SERVICE_ROLE}_INTERNAL_ORIGIN`: The internal http endpoint at which a service can be reached. Used for internal communication with other services.
+
+Endpoints availabe in deployment:
+  - `#{SERVICE_ROLE}_EXTERNAL_ORIGIN`: The load-balanced external https endpoint, as defined by the URL listed in the `remote.envorinments.#{remote-environment-id}` block of `service.yml`. Terraform manages creation of these records in route53.
+  - `#{SERVICE_ROLE}_INTERNAL_ORIGIN`: The internal http endpoint at which a service can be reached. Terraform manages creation of these records in route53. Used for internal communication with other services.
+
+
+## Worker services
```
charlierudolph: I think the word services can be removed. Same for public.

charlierudolph (approved):

PR #969, opened 2018-01-10
exit route53-updater on local run

hugoesthere:


charlierudolph (approved):

charlierudolph:
Can we also update the documentation so the user knows this was designed to not run locally with `exo run --production`?

hugoesthere:
Where would that documentation belong? In the same place we should also list that `ROLE` is passed as well

charlierudolph:
I was thinking next to the `update-route53` documentation in `service.md`

charlierudolph:
closing in favor of #977

Issue #961, opened 2018-01-09
exo generate terraform updates


charlierudolph:
Generate the `tfvars` files as well so terraform could be run manually

PR #970, opened 2018-01-11
exo generate tfvars

hugoesthere:
https://github.com/Originate/exosphere/issues/961


documentation/commands/generate.md
```diff
@@ -4,9 +4,13 @@ _Generates docker compose and terraform files_
 
 Usage: `exo generate [command]`
 
+Flags:
+- `-p, --profile string`  AWS profile to use (defaults to "default"). Only available for `exo generate terraform-var-files`
```
charlierudolph: Thoughts on actually breaking up this file into three? One for each command? I don't think we need to have them all in one file. They all are subcommands but each is distinct

charlierudolph:
Putting this on hold until #975 is in

Issue #966, opened 2018-01-09
application yml: default aws profile


charlierudolph:
Thoughts on updating the default aws profile used by `exo configure` / `exo deploy` to be the name of the project instead deferring to the aws default which is `default`. I think that will be the name of profile 99% of the time and makes it one less flag to supply as I have pretty much always  the `--profile` flag.

@alexdavid @hugobho thoughts?

alexdavid:
I think this is too magical. I think we should mirror the aws tools here and be explicit about profile names

charlierudolph:
Okay we can go without it for now. We'll see if we get any feedback about it down the road

PR #973, opened 2018-01-12
Use app name as default aws profile

hugoesthere:
https://github.com/Originate/exosphere/issues/966

### DIFF
--- a/src/cmd/shared.go
+++ b/src/cmd/shared.go
@@ -16,6 +16,9 @@ import (
 var awsProfileFlag string
 
 func getAwsConfig(appConfig *types.AppConfig, remoteEnvironmentID string, profile string) types.AwsConfig {
+	if profile == "" {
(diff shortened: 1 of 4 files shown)


src/cmd/shared.go
```diff
@@ -38,6 +41,9 @@ func getSecrets(awsConfig types.AwsConfig) types.Secrets {
 }
 
 func getBaseDeployConfig(appContext *context.AppContext, remoteEnvironmentID string) deploy.Config {
+	if awsProfileFlag == "" {
+		awsProfileFlag = appContext.Config.Name
+	}
```
charlierudolph, line 46: I don't think this is needed since it calls `getAwsConfig` which defaults it to the app name

charlierudolph (approved):

charlierudolph:
Closing for now. May revisit this later

PR #975, opened 2018-01-15
two stage deploy

charlierudolph:
TODO
* update services terraform with `module.aws` to `data.terraform_remote_state.main_infrastructre`
* use different buckets for holding the state


charlierudolph:
closing in favor of #978 that doesn't have a dependency on #970

PR #903, opened 2017-12-15
catch exo run errors

hugoesthere:
Resolves https://github.com/Originate/exosphere/issues/866

PR #982, opened 2018-03-14
exo-run: Catch manual shutdown error message

hugoesthere:
I think we had this check in before, did we take it out for a reason? It may have been hiding actual important exo run errors

### DIFF
--- a/src/application/runner/index.go
+++ b/src/application/runner/index.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"os/signal"
 	"path"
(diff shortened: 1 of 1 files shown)


charlierudolph:
https://github.com/Originate/exosphere/pull/903

hugoesthere:
Hm okay. This isn't really an issue but we pretty much always see an error message that `docker-compose up` failed even when that's expected when a user uses ctrl + c to stop the process

